PR #18, opened 2024-03-24
organizeImports -> sortImports

zach-planet-nine:
As discussed, this changes the .vscode setting from `organizeImports` to `sortImports` to avoid disappearing imports while developing.


dahlia (approved):
Thanks for your contribution!

PR #20, opened 2024-03-24
switched 'only can' to 'can only'

wakest:
it sounds weird to me the other way around


dahlia (approved):
Thank you for adjustment!

PR #101, opened 2024-07-15
Fixed use of `setInboxDispatcher` after `setInboxListeners`

allouis:
There is a bug where setting the inbox dispatcher after the listeners causes an error to be thrown due to path mismatch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,12 @@ Version 0.11.3
 
 To be released.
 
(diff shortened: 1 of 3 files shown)


dahlia (commented):
Thanks for your contribution! As this feature was introduced at Fedify 0.11.0, I'm going to change the target branch of this pull request from *main* to *0.11-maintenance*.

dahlia (approved):
I'm going to merge this immediately after the CI passes.

PR #140, opened 2024-09-19
Update Actor documentation to prefer Actor IDs over handles

ThisIsMissEm:
Using a user-supplied handle as the Actor ID is somewhat of an activitypub anti-pattern, as it prevents the user from changing their username in the future (since the user would effectively need to be Move'd in that situation)

Instead, it is better to use the Actor ID from the database, e.g., a uuid primary key for the Actor.

These changes improve the documentation to emphasize this best practice.

## Summary by CodeRabbit

- **New Features**
	- Updated the user interface to include a `username` property for actor identification.
	- Enhanced actor identification by transitioning from `handle` to `actorId` in the dispatcher paths.

- **Documentation**
	- Revised documentation to reflect changes in actor identification and the updated `User` interface.

### DIFF
--- a/docs/manual/actor.md
+++ b/docs/manual/actor.md
@@ -30,7 +30,7 @@ The below example shows how to register an actor dispatcher:
 // @noErrors: 2451 2345
 import type { Federation } from "@fedify/fedify";
 const federation = null as unknown as Federation<void>;
-interface User { }
+interface User { username: string }
 const user = null as User | null;
 // ---cut-before---
 import { createFederation, Person } from "@fedify/fedify";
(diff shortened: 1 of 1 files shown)


docs/manual/actor.md
```diff
@@ -39,19 +39,19 @@ const federation = createFederation({
   // Omitted for brevity; see the related section for details.
 });
 
-federation.setActorDispatcher("/users/{handle}", async (ctx, handle) => {
-  // Work with the database to find the actor by the handle.
+federation.setActorDispatcher("/actors/{actorId}", async (ctx, actorId) => {
```
dahlia, line 42: There is one problem, the current URL template only allows `{handle}` as a placeholder. It would be better to rename the placeholder to `{identifier}` before the v1.0 release, I will work on it.

ThisIsMissEm:
Sure, that works! EmeliaOn 20. Sep 2024, at 06:50, Hong Minhee ***@***.***> wrote:﻿
@dahlia commented on this pull request.



In docs/manual/actor.md:
> @@ -39,19 +39,19 @@ const federation = createFederation({
   // Omitted for brevity; see the related section for details.
 });
 
-federation.setActorDispatcher("/users/{handle}", async (ctx, handle) => {
-  // Work with the database to find the actor by the handle.
+federation.setActorDispatcher("/actors/{actorId}", async (ctx, actorId) => {

There is one problem, the current URL template only allows {handle} as a placeholder. It would be better to rename the placeholder to {identifier} before the v1.0 release, I will work on it.

dahlia:
Close as 276c412a6f70b7e98b6d4bfe7fe3e2ce4e3566cf addressed it.

PR #215, opened 2025-03-08
Vendor multibase

### LINKED ISSUE #127: Internalize multibase
Fedify currently depends on the [multibase](https://github.com/multiformats/js-multibase) package, which is obsolete. Although it is succeeded by [multiformats](https://github.com/multiformats/js-multiformats), it does not expose the API Fedify needs. Therefore, we need to internalize the multibase code so that Fedify no longer depends on a deprecated library.

frodi-karlsson:
Closes #127 

### Description

Vendors multibase into the fedify repository. Care was taken to migrate types faithfully into Typescript, and relevant tests 1-to-1 into @std/testing, functionality-wise.

Unused functions were removed, as well as their tests, as there's no guarantee they'd be used in the future. If the need arises, they're easy to vendor.

Given that multibase uses MIT license, and fedify as well, there should be no issue there (not legal advice 😅)

### Reason for it being needed

From original ticket:
> Fedify currently depends on the [multibase](https://github.com/multiformats/js-multibase) package, which is obsolete. Although it is succeeded by [multiformats](https://github.com/multiformats/js-multiformats), it does not expose the API Fedify needs. Therefore, we need to internalize the multibase code so that Fedify no longer depends on a deprecated library.

### How to test

Aside from being passively tested through usage in common functionality, some relevant tests were vendored as well.
I am not very experienced in @std/testing so I do expect some pushback in that file, but the functionality tested in that file was mapped from the original multibase test files.

### DIFF
--- a/src/codegen/class.ts
+++ b/src/codegen/class.ts
@@ -112,7 +112,7 @@ export async function* generateClasses(
   yield `import { LanguageTag, parseLanguageTag }
     from "@phensley/language-tag";\n`;
   yield `import { decode as decodeMultibase, encode as encodeMultibase }
-    from "multibase";`;
+    from "../multibase/index.ts";`;
   yield `import { type DocumentLoader, getDocumentLoader, type RemoteDocument }
     from "${runtimePath}/docloader.ts";\n`;
   yield `import {
--- a/src/multibase/multibase.test.ts
+++ b/src/multibase/multibase.test.ts
@@ -1,0 +1,128 @@
+import { decode, encode } from "./index.ts";
+import * as constants from "./constants.ts";
+import { decodeText, encodeText } from "./util.ts";
+import { assertEquals, assertThrows } from "@std/assert";
+import { test } from "../testing/mod.ts";
+import type { BaseName } from "./types.d.ts";
+
+test("multibase.encode and decode", () => {
+  const testCases: Array<[BaseName, string, string]> = [
+    ["base16", decodeText(Uint8Array.from([0x01])), "f01"],
+    ["base16", decodeText(Uint8Array.from([15])), "f0f"],
+    ["base16", "f", "f66"],
+    ["base16", "fo", "f666f"],
+    ["base16", "foo", "f666f6f"],
+    ["base16", "foob", "f666f6f62"],
+    ["base16", "fooba", "f666f6f6261"],
+    ["base16", "foobar", "f666f6f626172"],
+    ["base32", "yes mani !", "bpfsxgidnmfxgsibb"],
+    ["base32", "f", "bmy"],
+    ["base32", "fo", "bmzxq"],
+    ["base32", "foo", "bmzxw6"],
+    ["base32", "foob", "bmzxw6yq"],
+    ["base32", "fooba", "bmzxw6ytb"],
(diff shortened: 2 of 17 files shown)


src/multibase/multibase.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      `Decoding buffer of ${name} failed`,
+    );
+  }
+
+  test("should allow base32pad full alphabet", () => {
+    const encodedStr = "ctimaq4ygg2iegci7";
+    const decoded = decode(encodedStr);
+    const encoded = encode("c", decoded);
+    assertEquals(decode(encoded), decoded);
+  });
```
dahlia: To break down tests into smaller parts, you should use `t.step()` method:

```suggestion
  await t.step("should allow base32pad full alphabet", () => {
    const encodedStr = "ctimaq4ygg2iegci7";
    const decoded = decode(encodedStr);
    const encoded = encode("c", decoded);
    assertEquals(decode(encoded), decoded);
  });
```

To get a `t` object, you should change the first line of this test:

```ts
test("multibase.encode and decode", async (t) => {
```

See also <https://docs.deno.com/runtime/fundamentals/testing/#test-steps>.

src/codegen/class.ts
```diff
@@ -112,7 +112,7 @@ export async function* generateClasses(
   yield `import { LanguageTag, parseLanguageTag }
     from "@phensley/language-tag";\n`;
   yield `import { decode as decodeMultibase, encode as encodeMultibase }
-    from "multibase";`;
+    from "../multibase/index.ts";`;
```
dahlia: Even though it already works, could you move the *src/multibase* directory to *src/runtime/multibase* for consistency? All local code that the generated code depends on should be inside *src/runtime/*:

```suggestion
    from "${runtimePath}/multibase/index.ts";`;
```

dahlia (changes requested):
Thanks for your contribution! Although it looks fine generally, I left a few comments. Could you take them a look?

src/multibase/multibase.test.ts
frodi-karlsson: Thank you for the patience and the link 🙏

frodi-karlsson: Is this too much granularity do you think?

```ts
./runtime/multibase/multibase.test.ts => multibase.encode and decode ... Encoding/Decoding base32 with yes mani ! ... ok (0ms)
```

dahlia: I think it's fine?

src/codegen/class.ts
```diff
@@ -112,7 +112,7 @@ export async function* generateClasses(
   yield `import { LanguageTag, parseLanguageTag }
     from "@phensley/language-tag";\n`;
   yield `import { decode as decodeMultibase, encode as encodeMultibase }
-    from "multibase";`;
+    from "../runtime/multibase/index.ts";`;
```
dahlia:
```suggestion
    from "${runtimePath}/multibase/index.ts";`;
```

To avoid hard-coding paths.

dahlia (changes requested):
Everything looks fine, except for one small thing!

dahlia (approved):
Thank you!

Issue #218, opened 2025-03-19
.well-known/webfinger endpoint thows 404 on localhost


revathskumar:
### Description

Following the [basic tutorial](https://fedify.dev/tutorial/basics#actor-dispatcher) when trying to 

```sh
curl http://localhost:8000/.well-known/webfinger?resource=acct:me@localhost:8000
```

### Expected Result

webfinger response.

### Actual Result 

Node.js: 404
Bun: 404
Deno: works as expected 

| Header | Header |
|--------|--------|
| Environment | Node.js 23.3.0 |
| Fedify version | 1.4.6 & 1.5.0-dev.717 |
| OS | Debian GNU/Linux 12 (bookworm) x86_64 | 

Additional details: 

This happens only when hostname is localhost with port. If we use `fedify tunnel` **it works as expected.** 
From my basic debugging, it seems like the issue is originating from 

https://github.com/fedify-dev/fedify/blob/37c51836c5940793e8ee81346081481cd076c570/src/webfinger/handler.ts#L174-L177

when the hostname is `localhost:8000` (hostname with port), `url.domainToASCII` will return a empty string. 

EDIT: 

Added info: On deno, this works as expected. 

EDIT 2: 

Added info: On Bun, throws 404 (same as Node.js)

dahlia:
Fixed in [Fedify 1.4.7](https://github.com/fedify-dev/fedify/releases/tag/1.4.7).

PR #219, opened 2025-03-19
fix(webfinger): handle domainToASCII in Nodejs & bun

revathskumar:
In Node.js & Bun, `domainToASCII` return empty string when hostname with port is given (Eg. `localhost:8000`).

Fixes: #218

### DIFF
--- a/src/webfinger/handler.test.ts
+++ b/src/webfinger/handler.test.ts
@@ -19,7 +19,7 @@ test("handleWebFinger()", async (t) => {
       url,
       data: undefined,
       getActorUri(identifier) {
-        return new URL(`https://example.com/users/${identifier}`);
+        return new URL(`${url.protocol}//${url.host}/users/${identifier}`);
       },
       async getActor(handle): Promise<Actor | null> {
         return await actorDispatcher(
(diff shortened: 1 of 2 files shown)


ThisIsMissEm:
Could you also add a test case for a valid hostname like example.com, but with a port? I'm wondering if that is handled incorrectly at the moment too. The webfinger spec doesn't mention anything about handling ports / domains: https://datatracker.ietf.org/doc/html/rfc7033

This test also doesn't test for what it says it does: https://github.com/fedify-dev/fedify/blame/94721fe927df0cbf105f053783552481b9979e39/src/webfinger/handler.test.ts#L278 (both resources are `https:` instead of `http:`, though it does test not found behaviour

dahlia:
Thank you for reporting #218 and debugging this! As it should be applied to previous versions (1.4.x), I'm going to make a separate patch to fix it.

revathskumar:
@ThisIsMissEm Added a testcase with `http://example.com:8000` as the hostname.

src/webfinger/handler.test.ts
```diff
@@ -115,6 +116,90 @@ test("handleWebFinger()", async (t) => {
...
+        href: "http://example.com:8000/users/someone",
+        rel: "self",
+        type: "application/activity+json",
+      },
+      {
+        href: "http://example.com:8000/@someone",
+        rel: "http://webfinger.net/rel/profile-page",
+      },
+      {
+        href: "https://example.org/@someone",
```
ThisIsMissEm, line 175: I think you missed one?

```suggestion
        href: "https://example.org:8000/@someone",
```

revathskumar: @ThisIsMissEm I left it intentionally assuming that it's alternate url and not based on `actorUri`
for the other testcase "localhost with port", I left it as "example.org". 

https://github.com/fedify-dev/fedify/blob/d3ac8f8734fb6fe522a95e8830ebf40526ad9be2/src/webfinger/handler.test.ts#L128-L136

Do I need to make change in both places?

dahlia:
Fixed in [Fedify 1.4.7](https://github.com/fedify-dev/fedify/releases/tag/1.4.7).

PR #223, opened 2025-03-27
fix(docLoader): handle exception from kvstore and log error message

revathskumar:
In case of an exception when writing or reading from kvStore, fedify logs the below error, which is confusing and not helpful for the developers.

```
jsonld.InvalidUrl: Dereferencing a URL did not result in a valid JSON-LD
object. Possible causes are a
n inaccessible URL perhaps due to a same-origin policy (ensure the
server uses CORS if you are using client-side JavaScript), too many
redirects, a non-JSON response, or more than one HTTP Link Header was
provided for a remote context.
URL: "https://w3id.org/security/multikey/v1".
```

So, this patch tries to handle the exception from kvStore and log the error message.

### Steps to reproduce the problem

* Use `@fedify/postgres` as kvStore.
* Configure invalid database url.
* Try `curl  -H"Accept: application/activity+json" http://localhost:3000/users/demo`

Sample logs after the handling the exception.

image

### DIFF
--- a/src/runtime/docloader.test.ts
+++ b/src/runtime/docloader.test.ts
@@ -3,6 +3,7 @@ import { assertEquals, assertRejects, assertThrows } from "@std/assert";
 import * as mf from "mock_fetch";
 import process from "node:process";
 import metadata from "../deno.json" with { type: "json" };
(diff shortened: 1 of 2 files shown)


dahlia (approved):
Thanks for your contribution!

Issue #230, opened 2025-05-09
Error With Fedify-CLI Init


LithiumSource:
Hello, could you please help me with this error.

```
# npm resolution error report

While resolving: undefined@undefined
Found: @fedify/fedify@1.5.1
node_modules/@fedify/fedify
  @fedify/fedify@"^1.4.8" from the root project

Could not resolve dependency:
peer @fedify/fedify@">=0.12.0, <1" from @fedify/express@0.1.4
node_modules/@fedify/express
  @fedify/express@"^0.1.3" from the root project

Fix the upstream dependency conflict, or retry
this command with --force or --legacy-peer-deps
to accept an incorrect (and potentially broken) dependency resolution.

```

ThisIsMissEm:
@dahlia I can reproduce this on npm version 10.9.2, and specifically, when installing individually each dependency, I cannot reproduced.

However, if I run the command that `fedify init` would use, then it would be these:
```sh
$ echo "{}" > package.json
$ npm add @fedify/fedify@1.5.1 @logtape/logtape@^0.8.2 express@^4.19.2 @fedify/express@^0.1.3 @dotenvx/dotenvx@^1.14.1 tsx@^4.17.0
```
and this **does** fail with the error above.

It does work if I change it to include `--legacy-peer-deps`:
```sh
$ npm add @fedify/fedify@1.5.1 @logtape/logtape@^0.8.2 express@^4.19.2 @fedify/express@^0.1.3 @dotenvx/dotenvx@^1.14.1 tsx@^4.17.0 --legacy-peer-deps
```

I suspect part of the issue is this semver range, which doesn't look quite right: 
```
"peerDependencies": {
    "@fedify/fedify": ">=0.12.0, <2"
  },
```
Since it doesn't seem like a valid range: https://jubianchi.github.io/semver-check/#/%3E%3D0.12.0%2C%20%3C2/1.5.1

However, without the comma it is: https://jubianchi.github.io/semver-check/#/%3E%3D0.12.0%20%3C2/1.5.1

The other option that would work is doing two installs:
1. for fedify & logtape
2. for initializer packages

This does still complain:
```sh
$ npm add express@^4.19.2 @fedify/express@^0.1.3 @dotenvx/dotenvx@^1.14.1 tsx@^4.17.0 --force                
npm warn using --force Recommended protections disabled.
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: @fedify/express@0.1.4
[...]

PR #231, opened 2025-05-10
Update fedify CLI init to use @fedify/express ^0.2.1

ThisIsMissEm:
This is necessary for fixing #230 as the new version exceeds the semver range of ^0.1.3

https://jubianchi.github.io/semver-check/#/^0.1.3/0.2.1

So 0.2.1 isn't being automatically installed.


dahlia (approved):
Thanks!

PR #282, opened 2025-07-10
fix(bug): `fedify node` breaks favicon on terminal emulators without truecolor

### LINKED ISSUE #168: `fedify node` breaks favicon on terminal emulators without truecolor
Currently the `fedify node` command breaks favicon on terminal emulators without 24-bit truecolor.[^1]  Here's an example in macOS's built-in Terminal app, which doesn't support truecolor:

fedify node is broken on macOS's built-in Terminal app

For your information, it should be shown like below, which is taken in GNOME Terminal:

fedify node normally works on GNOME Terminal

We have two options to address it:

1. Properly fix the favicon.

    - [Detect if the running terminal emulator supports truecolor.](https://github.com/termstandard/colors?tab=readme-ov-file#truecolor-detection)
    - If the terminal emulator does not support it, display favicon in 256 color.

2. Make it hide favicon at all.

    - Detect if the running terminal emulator supports truecolor.
    - If the terminal emulator does not support it, do not display favicon at all.

[^1]: Note that quite many terminal emulators do not support truecolor. See also [Terminal Colors](https://github.com/termstandard/colors).

dodok8:
Fix #168

By detecting the current terminal emultor supports 24bit colors, if it doesn't support it shows 256-colors instead.

스크린샷 2025-07-10 18 21 46

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -56,6 +56,10 @@ To be released.
      -  The `--allow-private-address` or `-p` option allows looking up
         WebFinger information for private addresses (e.g., `localhost`).
 
+ - Fixed favicon corruption bug that occurred in terminals that don't support 24-bit color when using `fedify node`.
+    [[#168] by Kim, Hyeonseo]
+
+[#168]: https://github.com/fedify-dev/fedify/issues/168
 [#248]: https://github.com/fedify-dev/fedify/issues/248
 [#260]: https://github.com/fedify-dev/fedify/issues/260
 [#278]: https://github.com/fedify-dev/fedify/pull/278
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -56,6 +56,10 @@ To be released.
      -  The `--allow-private-address` or `-p` option allows looking up
         WebFinger information for private addresses (e.g., `localhost`).
 
+ - Fixed favicon corruption bug that occurred in terminals that don't support 24-bit color when using `fedify node`.
+    [[#168] by Kim, Hyeonseo]
```
dahlia: Is the comma a part of your name? It looks like two different names.

dahlia (commented):
Thanks for your first contribution! Looks good in general, but you need to run the formatter: `deno fmt`.

dahlia (approved):
Thanks!

Issue #270, opened 2025-07-02
[Proposal] @fedify/testing


malkoG:
### Summary
We need a testing library for Fedify that makes it easy to test ActivityPub interactions using familiar Jest-style expectations. The library should be framework-agnostic (work with Express, Hono, AdonisJS, NestJS, etc.) and provide a clean API for verifying both incoming and outgoing activities.

### Problem
Testing federated applications is challenging because:
- You need to mock remote actors and their inboxes/outboxes
- You need to verify that activities are sent correctly
- You need to simulate incoming activities and verify handlers are called
- Current testing requires a lot of boilerplate

### Proposed Solution

A testing library with this API:

```typescript
import { expect, mockInbox, mockOutbox } from '@fedify/test';

describe('Follow flow', () => {
  it('handles follow request', async () => {
    const bobInbox = mockInbox('https://local.example/users/bob/inbox');
    const bobOutbox = mockOutbox('https://local.example/users/bob/outbox');
    
    // Expect to receive activity
    expect(bobInbox).to.receiveActivity('Follow')
      .from('https://remote.example/users/alice');
    
    // Expect to send response
    expect(bobOutbox).to.sendActivity('Accept')
      .to('https://remote.example/users/alice/inbox')
      .accepting('Follow');
    
    // Simulate the incoming activity
    await simulateActivity(bobInbox, {
      type: 'Follow',
      actor: 'https://remote.example/users/alice',
      object: 'https://local.example/users/bob'
    });
  });
});
```

### Key Features

1. **Jest-style expectations**
   - `expect(inbox).to.receiveActivity('Follow')`
   - `expect(outbox).to.sendActivity('Accept')`
   - Familiar API for developers

2. **Framework agnostic**
   - Works with any web framework
   - No dependencies on specific implementations

3. **Activity simulation**
   - Easy to simulate incoming activities
   - Verify handlers are called correctly

4. **Clean API**
   - Minimal boilerplate
   - Intuitive method chaining
   - Type-safe with TypeScript

### Use Cases

- **Follow/Unfollow**: Test the complete follow handshake
- **Accept/Reject**: Verify proper responses to activities  
- **Block/Undo**: Test blocking behavior
- **Activity Delivery**: Verify activities are sent to correct inboxes
- **Handler Verification**: Ensure inbox listeners are called

### Benefits

- Makes testing federated apps as easy as testing regular apps
- Reduces test complexity and boilerplate
- Provides confidence that federation works correctly
[...]

PR #283, opened 2025-07-10
feat(test): `MockFederation` and `MockContext`

### LINKED ISSUE #197: `MockFederation` & `MockContext` for easier unit tests
To make Fedify apps testable, Fedify should provide `MockFederation` and `MockContext`.  They need to satisfy `Federation` and `Context` interfaces, and have few more properties and methods such as:

- `receiveActivity(activity: Activity): Promise<void>`
- `getSentActivities(): Activity[]`
- `clearSentActivities(): void`

notJoon:
# Description

This PR implements mock classes for the `Federation` and `Context` interface to improve testability, addressing issue #197 

## `MockFederation` Class

The new `MockFederation` class provides a complete mock implementation of the `Federation` interface that allows developers to test their fedify applications without setting up a real federation server. It racks all sent activities with some metadata including:

- Whether they were queued or sent immediately
- Which queue was used
- The order in which they were sent

The class also remembers all registered paths (actor, inbox, outbox, etc.) and uses them when generaing URIs, ensuring that tests accurately reflect the application's routing configuration.

## `MockContext` Class

The `MockContext` class implements the `Context` interface and interacts with `MockFederation` to provide a complete testing environment. It generates URIs based on the paths registered in the associated `MockFederation` instance and tracks all activities sent through it with full sender and recipient information.

## Testing Methods

Both classes provides testing-specific functions that aren't part of the interfaces:

- `receiveActivity()`: simulating incoming activities
- `getSentActivities()`: inspecting what was sent
- `clearSentActivities()`: resetting state

### Queue Awarenss

When `startQueue` is called on a `MockFederation` instance, all subsequent activities are marked as queued with appropriate queue type metadata. This allows tests to verify whether activities would be sent immediately or through the queue system in production.[^1]

### Supported Patterns

The mock classes support all standard Fedify patterns including:

- Custom path registration
- Multiple activity type listeners
- Propoer context data passing

For exaple, you can register custom path like `/actors/{identifier}` instead of the default `/users/{identifier}`, and all URL generation methods will respect those custom paths. We can also set up multiple inbox listners for different activity types and simulate receiving activities to verify they're processed correctly.

## Related Issues

closing #197
related with #270

- [^1]: https://github.com/fedify-dev/fedify/pull/283#discussion_r2197589478


fedify/testing/mock.test.ts
```diff
@@ -0,0 +1,197 @@
...
+  assertEquals(typeof mockFederation.setNodeInfoDispatcher, "function");
+  assertEquals(typeof mockFederation.setActorDispatcher, "function");
+  assertEquals(typeof mockFederation.setObjectDispatcher, "function");
+  assertEquals(typeof mockFederation.setInboxDispatcher, "function");
+  assertEquals(typeof mockFederation.setOutboxDispatcher, "function");
+  assertEquals(typeof mockFederation.setInboxListeners, "function");
+  assertEquals(typeof mockFederation.startQueue, "function");
+  assertEquals(typeof mockFederation.createContext, "function");
+  assertEquals(typeof mockFederation.fetch, "function");
+});
```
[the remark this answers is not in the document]
notJoon: https://github.com/fedify-dev/fedify/pull/283/commits/ef2416c18f2cd509e598fdd893165ea7645b24e1

fedify/testing/mock.test.ts
```diff
@@ -0,0 +1,197 @@
...
+  });
+
+  assertEquals(typeof mockContext.getActorUri, "function");
+  assertEquals(typeof mockContext.getInboxUri, "function");
+  assertEquals(typeof mockContext.getOutboxUri, "function");
+  assertEquals(typeof mockContext.getFollowingUri, "function");
+  assertEquals(typeof mockContext.getFollowersUri, "function");
+  assertEquals(typeof mockContext.parseUri, "function");
+  assertEquals(typeof mockContext.getSentActivities, "function");
+});
```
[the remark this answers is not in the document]
notJoon: https://github.com/fedify-dev/fedify/pull/283/commits/ef2416c18f2cd509e598fdd893165ea7645b24e1

fedify/testing/mock.ts
```diff
@@ -0,0 +1,691 @@
...
+      },
+    };
+  }
+
+  async startQueue(
+    contextData: TContextData,
+    _options?: FederationStartQueueOptions,
+  ): Promise<void> {
+    this.contextData = contextData;
+    // not actually starting a queue
```
ThisIsMissEm: I'd probably expect this to add to a `queuedActivities` state, such that there's a difference detectable between sending an activity immediately and sending it via queueing

fedify/testing/mock.ts
```diff
@@ -0,0 +1,691 @@
...
+
+  /**
+   * Gets all activities that have been sent through this mock federation.
+   * This method is specific to the mock implementation and is used for
+   * testing purposes.
+   *
+   * @returns An array of sent activities.
+   * @since 1.8.0
+   */
+  getSentActivities(): Activity[] {
```
ThisIsMissEm: Would it be better to make this a getter? Then you could write: `mock.sentActivities.at(0)....` without the extra function call?

fedify/testing/mock.ts
```diff
@@ -0,0 +1,691 @@
...
+  }
+
+  /**
+   * Clears all sent activities from the mock federation.
+   * This method is specific to the mock implementation and is used for
+   * testing purposes.
+   *
+   * @since 1.8.0
+   */
+  clearSentActivities(): void {
```
ThisIsMissEm: Maybe `reset` would be a better term here?

fedify/testing/mock.ts
```diff
@@ -0,0 +1,691 @@
...
+      tracerProvider: this.tracerProvider,
+    });
+  }
+
+  getNodeInfoUri(): URL {
+    return new URL("/nodeinfo/2.0", this.origin);
+  }
+
+  getActorUri(identifier: string): URL {
+    return new URL(`/users/${identifier}`, this.origin);
```
ThisIsMissEm: Isn't this based on the path registered by `setActorDispatcher`?

ThisIsMissEm: (Same comment applies for the below methods as well)

fedify/testing/mock.ts
```diff
@@ -0,0 +1,692 @@
...
+  async receiveActivity(activity: Activity): Promise<void> {
+    this.receivedActivities.push(activity);
+
+    // Find and execute appropriate inbox listeners
+    const typeName = activity.constructor.name;
+    const listeners = this.inboxListeners.get(typeName) || [];
+
+    for (const listener of listeners) {
+      const context = createInboxContext({
+        data: this.contextData!,
```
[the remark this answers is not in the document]
notJoon: https://github.com/fedify-dev/fedify/pull/283/commits/2b64f7cc36c0a5f05c21eda532ae376d6dbf5f27

fedify/testing/mock.ts
notJoon: I agree. I've implementing the queue tracking functionality as you suggested. As far as I aware, in fedify, the presence of `MessageQueue` determines whether activities are processed _asynchronously_ (queued) or _synchronously_ (immediate). It serves some batch processing and distribute work across multiple workers and so on.

For the mock implementation, I focused on making the observable behavior rather than fully simulating the async queue mechanics. When `startQueue` is called, all subsequent activities are marked as `queued` to indicate they would go through the queue system in a real federation setup. how do you think?

https://github.com/fedify-dev/fedify/pull/283/commits/2d8f3b9225edb6a799fa96ec74ca7186c4dca47c

fedify/testing/mock.ts
notJoon: In the issue, the name `clearSentActivities` was suggested, so I used it as is for now. `reset` also seems like a good name. 👍

fedify/testing/mock.ts
ThisIsMissEm: Would probably combine the metadata and the activity into one, so each item in sent might look like:
```
type SentActivity {
  queued: boolean
  queue: string
  activity: Activity
}
```

That way you can grab just the first sent activity and have direct access to all the context about that activity.

notJoon: That seems much more intuitive. additionaly, I put a counter (`sentOrder`) instead of a timestamp within `SentActivity` so that can track the activity order. https://github.com/fedify-dev/fedify/pull/283/commits/f23f127d75621d886a4bedeeddc37e07e6e9d385

fedify/testing/mock.ts
notJoon: Recently I got too used to Golang, so I got a habit of making everything into getters 😅
removed https://github.com/fedify-dev/fedify/pull/283/commits/97cb4c558d7b9cd09a30335909019c0a2ed591fd

fedify/testing/mock.ts
notJoon: That's the very tricky part to me. Within the codebase, I found that fedify uses:
- A `Router` class with RFC 6570 URI template
https://github.com/fedify-dev/fedify/blob/62db879971116fa4fbf71d15dadcdd2335a4f33b/fedify/federation/router.ts#L49-L53
- Path patterns are stored when dispatchers are registered, as you mentioned
- URI templates are expanded by replacing placeholders like `{identifier}` or `{handle}`

To the mock implementation, I've added same patterns:

1. Added an `expandUriTemplate` function that replaces placeholders with actual values
2. All dispatcher setter methods now store the registered path patterns

By doing so, each `getXXXUri` method now checks if a custom path was registered via the dispatcher and decide to use template expansion to generate the URI or just fallback to use default hardcoded paths.

The usage is as follows. For example:

~~~typescript
// When you register:
mockFederation.setActorDispatcher("/actors/{identifier}", dispatcher);

// getActorUri will use the registered path:
mockContext.getActorUri("john") // Returns: https://example.com/actors/john
~~~

This is the approach I had in mind for now. I'm sure there are better ways to do this.
https://github.com/fedify-dev/fedify/pull/283/commits/18e0af7568430e30d128fb9f385509dfc5ff6078

ThisIsMissEm: mmm, I'd probably just use the Router class itself inside the mock code.

ThisIsMissEm: So base it on: https://github.com/fedify-dev/fedify/blob/1.7.3/fedify/federation/builder.ts#L156

Something like this in the constructor:
```
constructor() {
    this.router = new Router();
}
```

And then in the dispatchers just use the router, and make `sendActivity` and `receiveActivity` commit the router automatically if it's not been committed already.

ThisIsMissEm: Actually, it looks like the code "builds" the routes each time when methods are invoked that need it, instead of having a single commit phase (admittedly I don't know much about the router implementation)

dahlia:
When I first wrote issue #197, what I had in mind was making this API usable not just by Fedify developers, but also by Fedify users. Was that what you had in mind too? Just asking to confirm.

notJoon:
> When I first wrote issue #197, what I had in mind was making this API usable not just by Fedify developers, but also by Fedify users. Was that what you had in mind too? Just asking to confirm.

@dahlia
Ultimately, I think other Fedify apps should also use this mock, but for now, I implemented it for internal testing purposes only. I think that by gradually improving it while accumulating real usage experience, we could create a better API based on actual needs. What do you think?

notJoon:
@dahlia @ malkoG Could you checks a this PR's direction? thank you!

fedify/testing/mock.ts
dahlia: Ditto, `reset()` would be better!

PR #300, opened 2025-07-16
Fix doubleKnock request.clone() errors

allouis:
This refactors the doubleKnock function and related signing functions to get the request body once at the top level instead of cloning it multiple times during iterations and redirects.

Changes:
- Add optional `body` parameter to SignRequestOptions and DoubleKnockOptions
- Update signRequestDraft and signRequestRfc9421 to accept pre-read body
- Modify doubleKnock to read body once and pass through all operations
- Update redirect handling to reuse the same body buffer

### DIFF
--- a/fedify/sig/http.ts
+++ b/fedify/sig/http.ts
@@ -1228,7 +1228,7 @@ export interface DoubleKnockOptions {
 
   /**
    * The request body as ArrayBuffer. If provided, avoids cloning the request body.
-   * @since 1.7.6
+   * @since 1.8.0
    */
   body?: ArrayBuffer | null;
 
--- a/fedify/sig/http.ts
+++ b/fedify/sig/http.ts
@@ -59,6 +59,12 @@ export interface SignRequestOptions {
   currentTime?: Temporal.Instant;
 
   /**
+   * The request body as ArrayBuffer. If provided, avoids cloning the request body.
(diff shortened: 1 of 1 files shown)


fedify/sig/http.ts
```diff
@@ -58,6 +58,12 @@ export interface SignRequestOptions {
    */
   currentTime?: Temporal.Instant;
 
+  /**
+   * The request body as ArrayBuffer. If provided, avoids cloning the request body.
+   * @since 1.7.6
```
dahlia: I think it should be included in v1.8.0 instead of v1.7.6.

fedify/sig/http.ts
```diff
@@ -1214,6 +1226,12 @@ export interface DoubleKnockOptions {
    */
   log?: (request: Request) => void;
 
+  /**
+   * The request body as ArrayBuffer. If provided, avoids cloning the request body.
+   * @since 1.7.6
```
dahlia: I think it should be included in v1.8.0 instead of v1.7.6.

dahlia (commented):
Looks fine to me, but I think it should be included in v1.8.0 instead of v1.7.6, because it introduces new option fields.

dahlia (approved):
Looks good to me. Thanks!

PR #308, opened 2025-07-17
fix(bug, test): ASCII art tests fail in PR #299

### LINKED ISSUE #304: ASCII art tests fail in PR #299: `getAsciiArt()` returns colored output instead of plain characters
The tests added in PR #299 for the `getAsciiArt()` function are failing both [in CI](https://github.com/fedify-dev/fedify/actions/runs/16333928670/job/46142623304#step:6:24427) and locally. The tests expect plain ASCII characters but are receiving ANSI color-coded output instead.

Failing tests:

- `getAsciiArt - Darkest Letter`: expects `█` but gets `[38;2;0;0;0m█[39m`
- `getAsciiArt - Brightest Letter`: expects ` ` (space) but gets `[38;2;255;255;255m [39m`

The issue appears to be that the `getAsciiArt()` function is applying ANSI color codes even when the tests expect plain text output. The tests were written with the assumption that the function would return uncolored characters, but the current implementation includes terminal color formatting by default.

This suggests either the test expectations need to be updated to account for colored output, or the `getAsciiArt()` function needs a parameter to control whether color codes should be applied, allowing the tests to request plain text output.

dodok8:
# Summary
Fix the issue that `getAsciiArt` tests failed according to the terminal emulator, add support for the non-color environment, and add test cases for color support. 

# Related Issue
- close #304 

# Problem / Motivation
Test added by #299 has failed according to the color support of the terminal emulator.

# Changes
- Refactor `trueColorSupport(): boolean` to `colorSupport(): 'truecolor'| '256color' | 'none'`.
- Add support for non-color environments.
- Add test case for color support.

# Benefits
- Add support for non-color terminals.
- Add more suitable test cases.

# Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -78,11 +78,7 @@ To be released.
 
   -  Fixed a bug where the `fedify node` command had failed to correctly
      render the favicon in terminal emulators that do not support 24-bit
-     colors.  [[#168], [#282] by Hyeonseo Kim]
-
-  - Fixed a bug where the getAsciiArt tests failed on terminal emulators with 
-    different color support capabilities. Enhanced color detection to support true 
-    color, 256-color, and non-color environments with appropriate test coverage. [[#304] by Hyeonseo Kim]
+     colors.  [[#168], [#282], [#304] by Hyeonseo Kim]
 
 [#168]: https://github.com/fedify-dev/fedify/issues/168
 [#248]: https://github.com/fedify-dev/fedify/issues/248
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -80,11 +80,16 @@ To be released.
      render the favicon in terminal emulators that do not support 24-bit
      colors.  [[#168], [#282] by Hyeonseo Kim]
 
+  - Fixed a bug where the getAsciiArt tests failed on terminal emulators with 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -80,11 +80,16 @@ To be released.
      render the favicon in terminal emulators that do not support 24-bit
      colors.  [[#168], [#282] by Hyeonseo Kim]
 
+  - Fixed a bug where the getAsciiArt tests failed on terminal emulators with 
+    different color support capabilities. Enhanced color detection to support true 
+    color, 256-color, and non-color environments with appropriate test coverage. [[#304] by Hyeonseo Kim]
```
dahlia: Since we haven't released a new version yet after adding 256-color output functionality to `fedify node`, you can just include it in the previously added changeset, like:

```markdown
  -  Fixed a bug where the `fedify node` command had failed to correctly
     render the favicon in terminal emulators that do not support 24-bit
     colors.  [[#168], [#282], [#304] by Hyeonseo Kim]
```

dodok8: changed 3b0f51b6e458e3a2901574fc9a5b65f91fea1f2f

dahlia (approved):
Thanks!

PR #312, opened 2025-07-18
Create GitHub Actions workflow remove-npm-pr-tags.yaml

### LINKED ISSUE #296: Automatically clean up PR tags from npm packages when PR is closed/merged
## Problem

Currently, when PRs are created or updated, our GitHub Actions workflow publishes packages to npm with tags like `pr-123` (where 123 is the PR number). Over time, these PR tags accumulate and clutter the npm package, making important tags like `latest` and `dev` less visible.

## Proposed Solution

Automatically remove PR-tagged npm dist-tags when the corresponding PR is closed or merged. This would involve:

1. Adding a GitHub Actions workflow that triggers on PR close/merge events
2. Detecting the PR number from the closed/merged PR
3. Using `npm dist-tag rm` to remove the corresponding `pr-{number}` tag
4. Ensuring proper error handling for cases where the tag might not exist

## Benefits

- Keeps npm package tags clean and organized
- Improves visibility of important tags (`latest`, `dev`, etc.)
- Better developer experience when browsing available package versions
- Preserves the actual package versions while removing temporary tags

## Implementation Considerations

- Need to handle both merged and closed PRs
- Should gracefully handle cases where PR tag doesn't exist
- Consider rate limiting and npm API constraints
- May need npm authentication tokens in GitHub Actions secrets
- Use `npm dist-tag rm @fedify/fedify pr-{number}` command

w8385:
## Summary

Automatically clean up temporary npm dist-tags (e.g., pr-123) when a pull request is closed.

## Related Issue

- closes #296 

## Changes

- Added a new GitHub Actions workflow: `remove-pr-npm-tags.yaml`
  - Triggers on `pull_request` events of type `closed`
    - When automatically closed by pull request merge.
  - Gracefully removes npm pr tags
    - Uses `npm dist-tag ls` to check if the tag exists
    - If it exists, removes it with `npm dist-tag rm`
  - Targets multiple packages in the monorepo
    - `@fedify/fedify`
    - `@fedify/amqp`
    - `@fedify/express`
    - `@fedify/h3`
    - `@fedify/postgres`
    - `@fedify/redis`

## Benefits

- Keeps npm package tags clean and organized
- Improves visibility of important tags (latest, dev, etc.)
- Better developer experience when browsing available package versions
- Preserves the actual package versions while removing temporary tags

## Checklist

### DIFF
--- a/.github/workflows/remove-npm-pr-tags.yaml
+++ b/.github/workflows/remove-npm-pr-tags.yaml
@@ -13,8 +13,8 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - uses: pnpm/action-setup@v4
-        name: Install pnpm
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
         with:
           version: 10
 


.github/workflows/remove-npm-pr-tags.yaml
```diff
@@ -0,0 +1,35 @@
+name: Remove npm PR tags
+
+on:
+  pull_request:
+    types:
+      - closed
```
dahlia, line 6: Is this workflow invoked also when a pull request is merged?

.github/workflows/remove-npm-pr-tags.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+
+      - name: Remove PR tag from npm
+        # Remove tag if exists
+        run: |
+          PKG="@fedify/fedify"
```
dahlia: We have multiple packages in the monorepo. They also should be dealt with.

dahlia (changes requested):

.github/workflows/remove-npm-pr-tags.yaml
w8385: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges

I just found the relevant section in the documentation. I’ll reflect that in the workflow accordingly.

.github/workflows/remove-npm-pr-tags.yaml
w8385: I should’ve asked this earlier—are the six packages the correct ones we should remove tags? (except for @fedify/cli)
[@fedify/fedify, @fedify/amqp, @fedify/express, @fedify/h3, @fedify/postgres, @fedify/redis]

.github/workflows/remove-npm-pr-tags.yaml
```diff
@@ -0,0 +1,48 @@
+name: Remove npm PR tags
+
+on:
+  pull_request_target:
+    types:
+      - closed
+
+jobs:
+  remove-pr-tags:
+    if: github.event.pull_request.merged == true
```
dahlia: We don't have to limit this workflow to run only when a pull request is merged. It also should be run when a pull request is closed without merge.

.github/workflows/remove-npm-pr-tags.yaml
```diff
@@ -0,0 +1,48 @@
...
+          TAG="pr-${PR_NUMBER}"
+          
+          PKGS=(
+            "@fedify/fedify"
+            "@fedify/amqp"
+            "@fedify/express"
+            "@fedify/h3"
+            "@fedify/postgres"
+            "@fedify/redis"
+          )
```
dahlia: You shouldn't hard-code this list. Instead, you could utilize `pnpm list` command, e.g.:

```sh
pnpm list --depth -1 -r --json | jq -r '.[].name | select(.)' | while IFS= read -r PKG; do
  if npm dist-tag ls "$PKG" | grep -q "^$TAG:" ; then
    npm dist-tag rm "$PKG" "$TAG"
    echo "Removed $TAG from $PKG"
  else
    echo "Tag $TAG does not exist on $PKG"
  fi
done
```

Of course, you should use [`pnpm/action-setup`](https://github.com/pnpm/action-setup) action to set up the pnpm before run this.

.github/workflows/remove-npm-pr-tags.yaml
w8385: The closed type includes both manually closed and merged pull requests, so this line didn’t need to be explicitly added!

.github/workflows/remove-npm-pr-tags.yaml
w8385: Looks like it's being utilized. 👍

PR #322, opened 2025-07-23
Fix Flooding alerts from "sponsors.yml" action

### LINKED ISSUE #317: Flooding alerts from "sponsors.yml" action
# Problem

Even in a forked repository, the workflow defined in .github/workflows/sponsors.yml keeps running repeatedly.

Each time the action runs, it fails and sends an email notification — which ends up being spammy to contributors.

# Proposed Behavior

The workflow should only run in the upstream repository, not in forks.

Image

dodok8:
## Summary

Fix Flooding alerts from "sponsors.yml" action by splitting the trigger action repository.

## Related Issue

- fixes #317 

## Changes

- Remove scheduled sponsors.yml action
- Add <https://github.com/fedify-dev/sponsor-automation>.

## Benefits

After this, the owner of the forked repository doesn't get notifications hourly.

## Checklist

## Additional Notes

### DIFF
--- a/.github/workflows/sponsors.yaml
+++ b/.github/workflows/sponsors.yaml
@@ -14,6 +14,8 @@ on:
 
 jobs:
   update-sponsors:
+    ## Only run on the upstream repository.
+    if: github.repository == 'fedify-dev/fedify'
     runs-on: ubuntu-latest
     permissions:
       contents: write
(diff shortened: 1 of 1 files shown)


ThisIsMissEm:
@dahlia an alternative to updating the sponsors every time they change would be to update at a less-frequent frequency, e.g., once a week or once a month, or even before each release.

PR #324, opened 2025-07-23
Fixed sending activities to path based redirects

allouis:
## Summary

There is a bug when handling redirects in `doubleKnock` when the `Location` header includes a path, rather than a full URL

## Changes

- Redirect URL is now constructed using the original URL as a base

## Benefits

This allows full URLs to continue to work, whilst paths will be resolved relative to the original

## Checklist


dahlia (commented):
Good catch. Thanks for your contribution! Could you make this pull request target at the *1.6-maintenance* branch instead of the *main* branch?

dahlia (approved):

PR #328, opened 2025-07-26
Add `--max-redirection` option to `fedify webfinger` command

### LINKED ISSUE #311: Add `--max-redirection` option to `fedify webfinger` command
### Summary

The `fedify webfinger` command is missing the `--max-redirection` option that should be available to control the maximum number of redirections allowed during WebFinger lookups, similar to the configurable `maxRedirection` option that was recently added to the WebFinger lookup API.

### Problem

The `fedify webfinger` command was added in issue #260 and implemented in PR #278, but it was created before the `maxRedirection` option was added to the WebFinger lookup API (issue #248, PR #281). As a result, users of the CLI command cannot configure the maximum number of redirections allowed during WebFinger lookups, while users of the programmatic API can.

This inconsistency creates a gap in functionality where:
- The underlying `lookupWebFinger()` function accepts a `maxRedirection` option (defaults to 5)
- The CLI command has no way to pass this option through to the underlying function
- Users who need to control redirection behavior must use the programmatic API instead of the convenient CLI command

### Proposed Solution

Add a `--max-redirection` option to the `fedify webfinger` command that:

1. **Accepts a number parameter** specifying the maximum number of redirections to follow
2. **Defaults to the same value as the API** (5 redirections) when not specified
3. **Passes the value through** to the `lookupWebFinger()` function's `maxRedirection` option
4. **Validates the input** to ensure it's a valid positive integer
5. **Provides helpful error messages** for invalid values

#### Implementation Details

Update `cli/webfinger.ts`:
- Add the option definition: `.option("--max-redirection <maxRedirection:number>", "Maximum number of redirections to follow (default: 5)")`
- Pass the option to the `lookupWebFinger()` call: `await lookupWebFinger(url, { ...options, maxRedirection: options.maxRedirection })`
- Add appropriate validation if needed

#### Example Usage

```bash
# Use default redirection limit (5)
fedify webfinger @user@example.com

# Set custom redirection limit
fedify webfinger @user@example.com --max-redirection 3

# Disable redirections entirely
fedify webfinger @user@example.com --max-redirection 0
```

### Alternatives Considered

1. **Leave as-is**: This would maintain the current inconsistency between CLI and API functionality
2. **Add a configuration file option**: More complex to implement and use for a single parameter
3. **Use a different option name**: `--max-redirection` follows the same naming convention as the API parameter

### Scope / Dependencies

**Files to modify:**
- `cli/webfinger.ts` — Add the new option and pass it through to the lookup function

**Dependencies:**
- This enhancement depends on the `maxRedirection` option being available in the `lookupWebFinger()` function (already implemented in PR #281)
- No breaking changes to existing functionality
- Backward compatible — existing commands will continue to work with the default behavior

**Components affected:**
- CLI webfinger command
- No changes required to the underlying WebFinger lookup functionality

w8385:
## Summary

Add `--max-redirection` option to `fedify webfinger` command.

## Related Issue

- closes #311

## Changes

- `cli/webfinger.ts`
  - Added new option `--max-redirection`
  - Validates max redirection value
  - Added new custom error class for invalid max redirection value
- `docs/cli.md`
  - Reflects changes above

## Benefits

- Users can configure the maximum number of redirections by CLI command.

## Checklist

## Additional Comments

For these expected invalid values, provide error messages below.

- negative integer
```
fedify webfinger @user@example.com --max-redirection -1
✖ Invalid max redirection value: -1
```

- real number
```
fedify webfinger @user@example.com --max-redirection 0.1
✖ Invalid max redirection value: 0.1
```


cli/webfinger.ts
```diff
@@ -25,6 +30,7 @@ export const command = new Command()
       }).start();
       try {
         const url = convertUrlIfHandle(resource); // Convert resource to URL
+        validateMaxRedirection(options.maxRedirection); // Validate max redirection
```
notJoon:
```suggestion
        validateMaxRedirection(options.maxRedirection);
```

notJoon (changes requested):

cli/webfinger.ts
```diff
@@ -17,6 +17,11 @@ export const command = new Command()
     "-p, --allow-private-address",
     "Allow private IP addresses in the URL.",
   )
+  .option(
+    "--max-redirection <maxRedirection:number>",
```
dahlia: [Cliffy provides an `integer` type.](https://cliffy.io/docs@v1.0.0-rc.7/command/types#built-in-types)

cli/webfinger.ts
```diff
@@ -74,6 +95,21 @@ class InvalidHandleError extends Error {
   }
 }
 
+/**
+ * Custom error class for invalid max redirection values.
+ * @param {number} maxRedirection The invalid max redirection value.
+ * @extends {Error}
+ */
+class InvalidMaxRedirectionError extends Error {
```
dahlia: Instead of defining `InvalidMaxRedirectionError`, you could utilize [Cliffy's `ValidationError`](https://cliffy.io/docs@v1.0.0-rc.7/command/error-handling#custom-validation-errors).

dahlia (changes requested):

cli/webfinger.ts
w8385: I'll reflect this!

cli/webfinger.ts
w8385: I'll remove this class and handle with Cliffy's `ValidationError`

dahlia (approved):

cli/webfinger.ts
w8385: Reflected in [eca4b3f](https://github.com/fedify-dev/fedify/pull/328/commits/eca4b3f4988a00b6eba0829362cfdbc7ec3f3185)!

Issue #306, opened 2025-07-17
Dynamically Fetch Dependency Versions


notJoon:
# Description

## Current Situation

In the `fedifycli/init.ts` file, dependencies other than fedify are pinned to specific hardcoded versions during code generation.

Related code:

https://github.com/fedify-dev/fedify/blob/0a66a35f166b379476921e60ad08819e1159da01/cli/init.ts#L1071-L1077

## Issues

- Expect `fedify`, dependency versions are hardcoded, preventing the use of latest versions
- Manual version updates are cumbersome or possibly human-error

## Proposed Solutions

1. Add getLatestVersion Function

- Implement a generic function that takes a package name and returns its latest version
- Remove the existing `getLatestFedifyVersion` function and consolidate into the new unified function

2. Leverage Monorepo Environment

- For fedify-related packages, import version information from other packages' `deno.json` files
- This enables automatic synchronization with the latest versions

## Consideration

- `LogTape` still require manual updates
- Use static imports, instead of dynamic file reading with `Deno.readTextFile()` to ensure the data is included when building the executable.

For example:
```typescript
import metadata from "../fedify/deno.json" with { type: "json" };
```

beberiche:
@dahlia I'd like to take this issue.

dahlia:
@beberiche addressed this in #329.

PR #329, opened 2025-07-26
Enhancement/cli dynamic dependency versions

beberiche:
## Summary
add `getLatestVersion()` function in the CLI init command to dynamically fetch the latest package versions from JSR API instead of using local metadata files.

## Related Issue
- #306 

## Changes
- Add `getLatestVersion()` and Replaced all hardcoded @fedify package versions throughout the init command with getLatestVersion() calls
- Added comprehensive test coverage for getLatestVersion() function in init.test.ts

## Benefits

- `@fedify/cli` run support latest version about `@fedify/projects` 

## Checklist

### DIFF
--- a/cli/init.ts
+++ b/cli/init.ts
@@ -5,8 +5,21 @@ import { getLogger } from "@logtape/logtape";
 import { stringify } from "@std/dotenv/stringify";
 import { exists } from "@std/fs";
 import { basename, dirname, join, normalize } from "@std/path";
-import { format, greaterThan, parse } from "@std/semver";
-import metadata from "./deno.json" with { type: "json" };
+import fedifyMetaData from "../fedify/deno.json" with { type: "json" };
+import amqpMetaData from "../amqp/deno.json" with { type: "json" };
+import expressMetaData from "../express/deno.json" with { type: "json" };
+import h3MetaData from "../h3/deno.json" with { type: "json" };
+import postgresMetaData from "../postgres/deno.json" with { type: "json" };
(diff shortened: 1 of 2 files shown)


cli/init.ts
```diff
@@ -5,8 +5,36 @@ import { getLogger } from "@logtape/logtape";
...
+}).then((m) => m.default);
+
+const packagesMetaData: Record<`@fedify/${string}`, Record<string, unknown>> = {
+  "@fedify/fedify": fedifyMetaData,
+  "@fedify/redis": redisMetaData,
+  "@fedify/postgres": postgresMetaData,
+  "@fedify/amqp": amqpMetaData,
+  "@fedify/express": expressMetaData,
+  "@fedify/h3": h3MetaData,
+};
```
dahlia: The current code will probably cause a runtime error when it's actually deployed, because the JSON file doesn't exist at the path you're trying to `import` dynamically.

I'm sorry to keep making you do things multiple times, but on second thought, since all `@fedify/*` packages have the same version number, it might be okay to just read the *deno.json* file of `@fedify/cli`, e.g.:

```ts
import metadata from "./deno.json" with { type: "json" };

const packagesMetaData: Record<`@fedify/${string}`, Record<string, unknown>> = {
  "@fedify/fedify": metadata.version,
  "@fedify/redis": metadata.version,
  "@fedify/postgres": metadata.version,
  "@fedify/amqp": metadata.version,
  "@fedify/express": metadata.version,
  "@fedify/h3": metadata.version,
};
```

Issue #99, opened 2024-07-11
More examples


dahlia:
It would be great to have more example code.

ellemedit:
making an express integration example

malkoG:
How about create section for examples in awesome list? #290 
Until awesome list out, we can reference the example in this issue.

cc: @nyeong

malkoG:
That would be good, too!

dahlia:
Following up on the discussion here, let me try to land something concrete so this can start moving.

## Proposed scope

I'd like to split what's being asked in this issue into two separate things:

 -  *A catalog of canonical example projects*, each a full runnable app covering a distinct scenario:
     1)  Federated *blog*: long-form posts with comments and followers.
     2)  Federated *social network*: multi-user microblog with timelines, follows, and likes. Goes beyond the existing *Creating your own federated microblog* tutorial, which is deliberately single-user.
     3)  Federated *content sharing platform*: media-centric posts (images or short video) with `Announce`/boost semantics.
     4)  A focused *custom collection dispatcher* example covering pagination, filtering, and authorization patterns. There's been a steady trickle of questions about this on Discord and in issues, and none of them have a single place to point at.
 -  *A tutorial on fedify.dev for each example*. The example repo is the finished reference; the tutorial walks through building it. This is the shape I want for the long-form docs, not a wall of isolated snippets.

## Repository layout

As I said earlier in this thread, I'd rather not grow *packages/fedify/examples/* for this. Rather than one shared monorepo, I'd like one repository per scenario under the `fedify-dev` org, following the pattern we already use for `fedify-dev/microblog`. Keeping them separate means each example is independently cloneable, deployable, and discoverable, and each one can pin its own runtime and dependency versions without dragging the others along.

## Stacks

@malkoG proposed building out the full (ORM × DB × framework) matrix. That's the right long-term shape, but trying to ship the whole matrix at once will stall this. I'd rather pick one opinionated stack per scenario for the first pass, and let community contributions fill in variants afterwards.

Across the four examples I want to cover all four of our framework integrations, and I want the first pass to show both patterns our users actually ask about: with and without an ORM. Tentative choices:

 -  Blog: Astro, no ORM (raw `bun:sqlite`), SQLite.
 -  Social network: Next.js, Drizzle ORM, PostgreSQL.
 -  Content sharing: Nuxt, Drizzle ORM, PostgreSQL.
 -  Custom collection dispatcher: Hono, no ORM (raw `node:sqlite`), SQLite.

[...]

PR #333, opened 2025-07-27
Update each examples to use pnpm as package manager

malkoG:
Maybe it is related to #99. 

With this changes, Ensure that all examples in the repository share the same dependency versions.


pnpm-workspace.yaml
```diff
@@ -9,6 +9,7 @@ packages:
 - redis
 - testing
 - docs
+- examples/cloudflare-workers
```
dahlia, line 12: You should add `examples/express`, `examples/next14-app-router`, and `examples/next15-app-router` packages as well.

PR #347, opened 2025-08-02
fix(codegen): randomize output filename to prevent race condition in parallel test runs

HanseoSong:
## Summary

This PR improves test stability by addressing a race condition in the code generation process. It ensures that parallel test runs do not conflict over a shared generated file.

## Related Issue

N/A

## Changes

* Modified `codegen/main.ts` to accept an explicit output path argument.
* Updated `codegen` task in `deno.json` to generate a unique filename using a random UUID.
* Adjusted `.gitignore` to exclude temporary vocab files matching `vocab-*.ts`.

## Benefits

* Prevents parallel test runs from conflicting over a shared generated file, avoiding race conditions.
* Increases robustness and reliability of test and build workflows.
* Enables safe parallelization of tests involving code generation.

## Checklist

* [ ] Did you add a changelog entry to the *CHANGES.md*?
* [ ] Did you write some relevant docs about this change (if it's a new feature)?
* [x] Did you write a regression test to reproduce the bug (if it's a bug fix)? *(N/A — test workaround applied)*
* [x] Did you write some tests for this change (if it's a new feature)? *(Updated test config and codegen flow)*
* [x] Did you run `deno task test-all` on your machine?

### DIFF
--- a/fedify/codegen/main.ts
+++ b/fedify/codegen/main.ts
@@ -1,9 +1,9 @@
-import { loadSchemaFiles } from "./schema.ts";
 import { generateClasses } from "./class.ts";
+import { loadSchemaFiles } from "./schema.ts";
 
 export async function main() {
-  if (Deno.args.length != 2) {
-    if (Deno.args.length < 2) {
+  if (Deno.args.length != 3) {
+    if (Deno.args.length < 3) {
       console.error("error: too few arguments");
     } else {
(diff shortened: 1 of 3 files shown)


malkoG:
Could you rebase?

`git pull upstream main -r`

dahlia (approved):

PR #348, opened 2025-08-02
Display icon/image on fedify lookup

### LINKED ISSUE #169: Display `icon`/`image` on `fedify lookup`
It would be fun if the `fedify lookup` command displays `icon`/`image` when the object has one.  We could utilize [Terminal Graphics Protocol](https://sw.kovidgoyal.net/kitty/graphics-protocol/) or [Inline Images Protocol](https://iterm2.com/documentation-images.html).

sij411:
## Summary

`fedify lookup` command displays`Icon`/`Image` on terminal console depending on the terminal emulator.

## Related Issue

closes #169 

## Changes

- Render images to terminal if the result objects have images when user runs `fedify lookup` 
- It doesn't display images when `-o, --output` option is given.
- It displays image below the fetched object.

## Benefits

Users can see the actual images via terminal instead of the image url if the terminal supports graphic protocol.

## Checklist

## Additional Notes

Image sizes are fixed for now because every terminal emulators support measuring console size and image differently.
Opening issues to support other terminal emulators could be a way to support terminals as many as possible.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -127,8 +127,17 @@ the versioning.
  -  The `fedify` CLI now correctly disables color output when standard output
     isn't a TTY (for example, when redirecting to a file) or when the `NO_COLOR`
     environment variable is set.  [[#257], [#341] by Cho Hasang]
+    
+ -  The 'fedify lookup' command now displays images depending on user's 
+    terminal emulator. [[#169], [#348] by Jiwon Kwon]
+
+     - Supported terminal emulators are `Kitty`, `WezTerm`, `Konsole`, `Warp`,
+      `Wayst`, `st`, and `iterms`.
+
+
 
 [#168]: https://github.com/fedify-dev/fedify/issues/168
+[#169]: https://github.com/fedify-dev/fedify/issues/169
 [#191]: https://github.com/fedify-dev/fedify/issues/191
 [#197]: https://github.com/fedify-dev/fedify/issues/197
 [#248]: https://github.com/fedify-dev/fedify/issues/248
@@ -155,8 +164,9 @@ the versioning.
 [#328]: https://github.com/fedify-dev/fedify/pull/328
 [#331]: https://github.com/fedify-dev/fedify/pull/331
 [#332]: https://github.com/fedify-dev/fedify/pull/332
-[#342]: https://github.com/fedify-dev/fedify/pull/342
 [#341]: https://github.com/fedify-dev/fedify/pull/341
+[#342]: https://github.com/fedify-dev/fedify/pull/342
+[#348]: https://github.com/fedify-dev/fedify/pull/348
 
 
 Version 1.7.7
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -128,11 +128,11 @@ the versioning.
     isn't a TTY (for example, when redirecting to a file) or when the `NO_COLOR`
     environment variable is set.  [[#257], [#341] by Cho Hasang]
     
- -  The 'fedify lookup' command now displays images depending on user's 
+ -  The `fedify lookup` command now displays images depending on user's 
     terminal emulator. [[#169], [#348] by Jiwon Kwon]
 
-     - Supported terminal emulators are `Kitty`, `WezTerm`, `Konsole`, `Warp`,
-      `Wayst`, `st`, and `iterms`.
+     - Supported terminal emulators are Kitty, WezTerm, Konsole, Warp,
+      Wayst, st, and iTerm.
 
 
 
--- a/package.json
+++ b/package.json
@@ -3,5 +3,8 @@
     "patchedDependencies": {
       "vitepress@1.6.3": "patches/vitepress@1.6.3.patch"
     }
+  },
+  "dependencies": {
+    "sharp": "^0.34.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -3,8 +3,5 @@
     "patchedDependencies": {
       "vitepress@1.6.3": "patches/vitepress@1.6.3.patch"
     }
-  },
-  "dependencies": {
-    "sharp": "^0.34.3"
   }
 }
--- a/packages/cli/src/imagerenderer.ts
+++ b/packages/cli/src/imagerenderer.ts
@@ -1,3 +1,4 @@
+import { encodeBase64 } from "@std/encoding/base64";
 import sharp from "sharp";
 
 export type TerminalType = "kitty" | "iterm2" | "none";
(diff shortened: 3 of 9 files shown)


cli/deno.json
```diff
@@ -24,7 +24,8 @@
     "icojs": "npm:icojs@^0.19.4",
     "jimp": "npm:jimp@^1.6.0",
     "ora": "npm:ora@^8.0.1",
-    "shiki": "npm:shiki@^1.6.4"
+    "shiki": "npm:shiki@^1.6.4",
+    "sharp": "npm:sharp"
```
dahlia: Its version should be pinned:


```suggestion
    "sharp": "npm:sharp@^0.34.3"
```

cli/lookup.ts
```diff
@@ -9,7 +9,7 @@ import {
   getAuthenticatedDocumentLoader,
   type Link,
   lookupObject,
-  type Object,
+  Object as ActivityObject,
```
malkoG: How about `APObject` instead of `ActivityObject`?
It could be Actor, Person, Note, ... None of these are not related to Activity

cli/lookup.ts
```diff
@@ -18,6 +18,7 @@ import { getLogger } from "@logtape/logtape";
 import { dirname, isAbsolute, resolve } from "@std/path";
 import ora from "ora";
 import { getContextLoader, getDocumentLoader } from "./docloader.ts";
+import { renderImage } from "./renderimage.ts";
```
malkoG: How about `renderer.ts` instead of `renderimage.ts`

cli/lookup.ts
sij411: Does AP stands for ActivityPub? Then what about the full name it might sound confusing

CHANGES.md
```diff
@@ -127,8 +127,17 @@ the versioning.
  -  The `fedify` CLI now correctly disables color output when standard output
     isn't a TTY (for example, when redirecting to a file) or when the `NO_COLOR`
     environment variable is set.  [[#257], [#341] by Cho Hasang]
+    
+ -  The 'fedify lookup' command now displays images depending on user's
```
```suggestion
 -  The `fedify lookup` command now displays images depending on user's 
```

CHANGES.md
```diff
@@ -127,8 +127,17 @@ the versioning.
  -  The `fedify` CLI now correctly disables color output when standard output
     isn't a TTY (for example, when redirecting to a file) or when the `NO_COLOR`
     environment variable is set.  [[#257], [#341] by Cho Hasang]
+    
+ -  The 'fedify lookup' command now displays images depending on user's 
+    terminal emulator. [[#169], [#348] by Jiwon Kwon]
+
+     - Supported terminal emulators are `Kitty`, `WezTerm`, `Konsole`, `Warp`,
+      `Wayst`, `st`, and `iterms`.
```
```suggestion
     -  Supported terminal emulators are Kitty, WezTerm, Konsole, Warp,
        Wayst, st, and iTerm.
```

packages/cli/src/imagerenderer.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    await Deno.writeFile(tempPath, imageData);
+
+    return tempPath;
+  } catch (_error) {
+    return null;
+  }
+}
+
+export async function renderImage(
```
dahlia: It's trivial, but wouldn't it better to be `renderImages` (plural)?

packages/cli/src/imagerenderer.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+export async function downloadImage(url: string): Promise<string | null> {
+  try {
+    const response = await fetch(url);
+    const imageData = new Uint8Array(await response.arrayBuffer());
+    const tempDir = Deno.env.get("TMPDIR") || Deno.env.get("TMP") || "/tmp";
+    const filename = `terminal_image_${Date.now()}_${crypto.randomUUID()}`;
+    const extension = new URL(url).pathname.split(".").pop() || "jpg";
+    const tempPath = `${tempDir}/${filename}.${extension}`;
```
dahlia: You could utilize [`Deno.makeTempFile()`](https://docs.deno.com/api/deno/~/Deno.makeTempFile) API instead.

package.json
```diff
@@ -3,5 +3,8 @@
     "patchedDependencies": {
       "vitepress@1.6.3": "patches/vitepress@1.6.3.patch"
     }
+  },
+  "dependencies": {
+    "sharp": "^0.34.3"
```
dahlia: I don't think it's a necessary change.

packages/cli/src/imagerenderer.ts
```diff
@@ -0,0 +1,149 @@
...
+type KittyCommand = Record<string, string | number>;
+
+export function detectTerminalCapabilities(): TerminalType {
+  const termProgram = (Deno.env.get("TERM_PROGRAM") || "").toLowerCase();
+
+  for (const id of KITTY_IDENTIFIERS) {
+    if (termProgram === id) {
+      return "kitty";
+    }
+  }
```
dahlia: You could utilize [`Array.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) method.

```suggestion
  if (KITTY_IDENTIFIERS.includes(termProgram)) return "kitty";
```

packages/cli/src/imagerenderer.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  return result;
+}
+
+export async function renderImageKitty(
+  imagePath: string,
+  cmd: KittyCommand,
+): Promise<void> {
+  const data = await Deno.readFile(imagePath);
+  const base64Data = btoa(String.fromCharCode(...data));
```
dahlia: The `btoa()` function is deprecated.  Use the [`encodeBase64()`](https://jsr.io/@std/encoding/doc/~/encodeBase64) function from the [@std/encoding](https://jsr.io/@std/encoding) package.  You can install it using `deno add` command:

```sh
deno add jsr:@std/encoding
```

dahlia (changes requested):

packages/cli/src/imagerenderer.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    isFirst = false;
+  }
+}
+
+export async function renderImageITerm2(
+  imagePath: string,
+): Promise<void> {
+  const imageData = await Deno.readFile(imagePath);
+  const base64Data = btoa(String.fromCharCode(...imageData));
```
dahlia: Use `encodeBase64()` here too!

CHANGES.md
```diff
@@ -127,8 +127,17 @@ the versioning.
  -  The `fedify` CLI now correctly disables color output when standard output
     isn't a TTY (for example, when redirecting to a file) or when the `NO_COLOR`
     environment variable is set.  [[#257], [#341] by Cho Hasang]
+    
+ -  The `fedify lookup` command now displays images depending on user's 
+    terminal emulator. [[#169], [#348] by Jiwon Kwon]
+
+     - Supported terminal emulators are Kitty, WezTerm, Konsole, Warp,
+      Wayst, st, and iterms.
```
dahlia: Alignment!

```suggestion
     -  Supported terminal emulators are Kitty, WezTerm, Konsole, Warp,
        Wayst, st, and iTerm.
```

dahlia (changes requested):
Few more nitpicks!

dahlia (approved):
:+1:

Issue #265, opened 2025-06-26
Add configuration file support to Fedify CLI


dahlia:
## Description

Add support for configuration files to the Fedify CLI tools to allow users to set default options and preferences. This will improve the developer experience by eliminating the need to repeatedly specify common options and allowing for project-specific configurations.

## Implementation Details

- Support `.fedifyrc` (JSON format) and `fedify.config.json` configuration files
- Look for config files in the current directory, then user home directory
- Allow configuration of common options like:
  - Default user-agent string
  - Cache directory location
  - Default output format preferences
  - Timeout settings
  - Debug/verbose mode preferences
- Command-line options should override config file settings
- Use existing JSON parsing patterns from the codebase

## Why This Is a Good First Issue

- **Clear requirements**: Well-defined feature with obvious use cases
- **Existing patterns**: Can follow JSON handling patterns already used in the codebase
- **Incremental implementation**: Can start with basic options and expand gradually
- **High impact**: Significantly improves developer experience for regular CLI users
- **Safe changes**: Additive feature that doesn't break existing functionality

## Acceptance Criteria

- [ ] Support `.fedifyrc` and `fedify.config.json` configuration files
- [ ] Search for config files in current directory, then home directory
- [ ] Allow configuration of common CLI options (user-agent, cache directory, etc.)
- [ ] Command-line options override config file settings
- [ ] Handle malformed config files gracefully with helpful error messages
- [ ] Add `--no-config` flag to ignore configuration files
- [ ] Update help text to mention configuration file support

## Configuration File Structure

```json
{
  "userAgent": "MyBot/1.0 (https://example.com/bot)",
  "cacheDir": "~/.cache/fedify",
  "defaultFormat": "json",
  "timeout": 30000,
  "verbose": false,
  "followRedirects": true
}
```

## Files to Create/Modify

- `cli/config.ts` - New configuration file handling logic
- `cli/utils.ts` - Add config loading utilities
- Multiple CLI command files - Update to use config values as defaults

## Example Usage

```bash
# Create config file
echo '{"userAgent": "MyBot/1.0", "defaultFormat": "json"}' > .fedifyrc

# Commands now use config defaults
fedify lookup @user@example.com  # Uses configured user-agent and format

# Override config with command line options
fedify lookup @user@example.com --format compact  # Overrides defaultFormat

# Ignore config file
fedify lookup @user@example.com --no-config
```

dahlia:
Hmm, it apparently depends on #258.

[...]

PR #349, opened 2025-08-03
Implement CLI Configuration Loading and Option Handling Improvements

devtedlee:
## Summary

This pull request introduces a configuration file system for the Fedify CLI, allowing users to set persistent options in `.fedifyrc` or `fedify.config.json`. It also centralizes and refactors command-line option handling to improve structural clarity, maintainability, and user experience by establishing a clear priority for settings.

## Related Issue

- Addresses feature #265

## Changes

- Created `packages/cli/src/config.ts` with an interface for configuration and a `loadConfig` function to read `.fedifyrc` and `fedify.config.json` files.
- Added unit tests for the configuration loading logic.
- Implemented a central `globalAction` in `mod.ts` to manage shared options and resource initialization (e.g., cache directory).
- Established a clear option precedence: Command-line flags > Config file settings > Hardcoded defaults.
- Introduced a `--no-config` flag to allow bypassing the configuration file loading.
- Relocated several global options (e.g., `--user-agent`) to their relevant subcommands (`lookup`, `nodeinfo`, `webfinger`).
- Centralized log level management based on `--debug` and `--verbose` flags within the `globalAction`.

## Benefits

- Improved Configurability: Users can define project-level settings in a configuration file, avoiding repetitive command-line flags and making the CLI easier to use.
- Enhanced Clarity: Moving options to specific subcommands makes the CLI's interface more intuitive and self-documenting.
- Increased Maintainability: Centralizing the logic for option parsing and merging into a single `globalAction` reduces code duplication and simplifies future development.
- Greater Flexibility & Testability: The `--no-config` flag enables running the CLI in an isolated state, which is crucial for scripting, automation, and reliable testing.
- Consistent Behavior: Shared resources and logging are initialized consistently across all commands, leading to more predictable outcomes.

## Additional Notes

This work was done incrementally, starting with adding the core config loading feature and then refactoring the option parsing logic for better structure. The changes affect the core CLI module and several subcommands, establishing a new, more robust foundation for how the CLI is configured.

## Focus on Review

- find out configs to apply: timeout, followRedirects, format


packages/cli/src/config.ts
```diff
@@ -0,0 +1,76 @@
...
+  const homeDir = Deno.env.get("HOME");
+
+  // search config file in current directory, priority arrays
+  const paths = [
+    join(currentDir, ".fedifyrc"),
+    join(currentDir, "fedify.config.json"),
+  ];
+  if (homeDir) {
+    paths.push(join(homeDir, ".fedifyrc"));
+    paths.push(join(homeDir, "fedify.config.json"));
```
dahlia: It would be great if it respects `XDG_CONFIG_HOME`.  See also <https://wiki.archlinux.org/title/XDG_Base_Directory>

packages/cli/src/lookup.ts
```diff
@@ -168,6 +172,7 @@ export const command = new Command()
     "-o, --output <file>",
     "Specify the output file path.",
   )
+  .option("--no-config [flag:boolean]", "Disable loading config file.")
```
dahlia: I think it shouldn't have its argument?

```suggestion
  .option("--no-config", "Disable loading config file.")
```

packages/cli/src/nodeinfo.ts
```diff
@@ -36,6 +40,7 @@ export const command = new Command()
     { conflicts: ["raw"] },
   )
   .option("-u, --user-agent <string>", "The custom User-Agent header value.")
+  .option("--no-config [flag:boolean]", "Disable loading config file.")
```
dahlia: Hmm, should we repeat this option for every subcommand?

devtedlee: this sub command also for type inference at each command. `globalOption` can't be type inference at each command.

packages/cli/src/lookup.ts
devtedlee: without argument, it's type inference show `config`, not `noConfig`. any other way to this type issue? 

image

dahlia: I think that's okay, isn't it? Use `!options.config` instead of `options.noConfig`.

devtedlee: Can I check my understand is right. 
I'll replace this to `.option("--config", "Enable loading config file.", { default: true })`
amm.. it's little bit strange I think.

oops. I got it now by cliffy doc

import { Command } from "https://deno.land/x/cliffy@v0.25.7/command/mod.ts";

new Command()
  .name("my-command")
  .option("--foo", "Enable foo")
  .option("--no-foo", "Disable foo (default)")
  .action(() => {
    // ...
  })
  .parse(Deno.args);

packages/cli/src/mod.ts
```diff
@@ -39,41 +57,42 @@ const command = new Command()
...
           {
             category: "localtunnel",
-            lowestLevel: "debug",
+            lowestLevel: logLevel,
             sinks: ["console", "file"],
           },
           {
             category: ["logtape", "meta"],
-            lowestLevel: "warning",
+            lowestLevel: logLevel,
```
dahlia: This should be `"warning"`.

packages/cli/src/nodeinfo.ts
dahlia: Hmm, yeah, that's maybe a limitation of Cliffy. That said, I think it's better to be a global option.

dahlia (changes requested):

packages/cli/src/config.ts
```diff
@@ -0,0 +1,79 @@
...
+  /** verbose mode */
+  verbose?: boolean;
+}
+
+export async function loadConfig(): Promise<Config> {
+  const currentDir = Deno.cwd();
+
+  // for more details, see https://wiki.archlinux.org/title/XDG_Base_Directory
+  const homeDir = Deno.env.get("XDG_CONFIG_HOME") ??
+    `${Deno.env.get("HOME")}/.config`;
```
dahlia: It might not work on Windows… As far as I know, Windows uses `UserProfile` instead of `HOME`.  See also <https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables>.

packages/cli/src/config.ts
```diff
@@ -0,0 +1,79 @@
...
+  cacheDir?: string;
+  /** verbose mode */
+  verbose?: boolean;
+}
+
+export async function loadConfig(): Promise<Config> {
+  const currentDir = Deno.cwd();
+
+  // for more details, see https://wiki.archlinux.org/title/XDG_Base_Directory
+  const homeDir = Deno.env.get("XDG_CONFIG_HOME") ??
```
dahlia: On Windows, it's better to use `AppData` instead of `XDG_CONFIG_HOME`. See also <https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables>.

dahlia (changes requested):

packages/postgres/src/mq.test.ts
```diff
@@ -31,6 +31,7 @@ test("PostgresMessageQueue", { skip: dbUrl == null }, async () => {
   try {
     const messages: string[] = [];
     const controller = new AbortController();
+    await mq.initialize();
```
dahlia: I think you should rebase your commits on the current *main* branch?

devtedlee: Cause of my mistake, already rebase many times. So duplicated commits exist. 😞

Anyway, I'll rebase it again thanks

devtedlee: rebase interactively and force push done.

packages/cli/src/nodeinfo.ts
```diff
@@ -237,7 +247,7 @@ export async function getFaviconUrl(
 ): Promise<URL> {
   const response = await fetch(url, {
     headers: {
-      "User-Agent": userAgent == null ? getUserAgent() : userAgent,
+      "User-Agent": userAgent ?? getUserAgent(),
```
dahlia, line 250: Don't we need to try `getSharedOption("userAgent", !options.config)` here as well?

packages/cli/src/config.ts
```diff
@@ -0,0 +1,102 @@
...
+
+const logger = getLogger(["fedify", "cli", "config"]);
+
+interface HttpConfig {
+  /** http header User-Agent */
+  userAgent?: string;
+  /** http request timeout */
+  timeout?: number;
+  /** auto follow redirects mode */
+  followRedirects?: boolean;
```
dahlia, line 12: Hmm… those two fields are apparently used nowhere?

packages/cli/src/config.ts
```diff
@@ -0,0 +1,102 @@
...
+ *   },
+ *   "cacheDir": "./cache",
+ *   "verbose": true
+ * }
+ */
+export interface Config {
+  /** HTTP related configuration */
+  http?: HttpConfig;
+  /** Output format configuration */
+  format?: FormatConfig;
```
dahlia, line 40: This field is not used either.

devtedlee: yes, I also noticed that at PR description 'focus on review' section. I don't know where to use this fields precisely. 

I think If we try to use this fields, we also modify other fedify related codes. like @fedify/fedify getDocumentLoader function

packages/cli/src/nodeinfo.ts
devtedlee: at line 95, when getFaviconUrl called, already use `getSharedOption` so I skipped to modify this code.

dahlia: Okay, fair enough!

packages/cli/src/config.ts
dahlia: See also the `writeObjectToStream()` function which is used in the *packages/cli/src/lookup.ts* file.

PR #352, opened 2025-08-05
Fixed kvCache for DocLoaders reading from kv store

allouis:
Fedify has several context documents built in and stored in-memory. By default the DocumentLoader will read from memory, but when it is wrapped with the kvCache, we end up reading from the kv store rather than memory which generates a lot of overhead for stores backed by an external db like Redis or MySQL.

## Summary

Improve performance when reading common contexts from the document loader

## Changes

- Updated kvCache wrapper to read from preloaded contexts

## Benefits

- This reduces network and disk overheads when parsing activitites and objects

## Checklist

- [ ] ~Did you write some relevant docs about this change (if it's a new feature)?~
- [ ] ~Did you write a regression test to reproduce the bug (if it's a bug fix)?~

## Additional Notes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@ Version 1.7.8
 
 To be released.
 
+ -  Updated `kvCache` wrapper to read from preloaded contexts rather than
+    from the KvStore. This saves network and disk overheads when parsing
+    activities and objects using the JSON-LD library.
+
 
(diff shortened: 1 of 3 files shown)


dahlia (approved):
Thank you!

dahlia:
Included in v1.7.8.

PR #364, opened 2025-08-10
Example for custom collection dispatchers

### LINKED ISSUE #337: Examples for custom collection dispatchers
### Summary

Recently, Custom Collection Dispatcher was added #310 . This allows users to easily dispatch various Collections of Objects. However, since it was added not long ago, it may still be unfamiliar to users. To solve this problem, writing examples using it will help lower the barrier to entry for users.

### Affected Docs

- Add `examples/custom-collection/` and inner files.
- Modify `examples/README.md`.

### What's Missing or Incorrect?

Examples of custom collection dispatcher are needed.

### Suggested Content or Structure

- Add `examples/custom-collection/` and inner files.
- Modify `examples/README.md`.

### Related Code / Feature

- Issue #310, PR #332 
- Custom Collection Dispatchers

sij411:
## Summary

Example for using custom collection dispatchers by printing posts and tagged posts 

## Related Issue

- closes #337 

## Changes

- Added `main.ts` : prints mock-up posts and tagged posts

## Benefits

It demonstrates users how to use custom collection dispatchers

## Checklist

## Additional Notes

Results look like this:
```sh
Custom collection data: {
  type: "Collection",
  id: "https://example.com/users/someone/tags/ActivityPub",
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    "https://w3id.org/security/data-integrity/v1",
    {
      toot: "http://joinmastodon.org/ns#",
      misskey: "https://misskey-hub.net/ns#",
      fedibird: "http://fedibird.com/ns#",
      ChatMessage: "http://litepub.social/ns#ChatMessage",
      sensitive: "as:sensitive",
      votersCount: "toot:votersCount",
      Emoji: "toot:Emoji",
      Hashtag: "as:Hashtag",
      quoteUrl: "as:quoteUrl",
      _misskey_quote: "misskey:_misskey_quote",
      quoteUri: "fedibird:quoteUri",
      emojiReactions: { "@id": "fedibird:emojiReactions", "@type": "@id" }
    }
  ],
  totalItems: 2,
  items: [
    {
      "@context": [
        "https://www.w3.org/ns/activitystreams",
        "https://w3id.org/security/data-integrity/v1",
        {
          toot: "http://joinmastodon.org/ns#",
          misskey: "https://misskey-hub.net/ns#",
          fedibird: "http://fedibird.com/ns#",
          sensitive: "as:sensitive",
          Emoji: "toot:Emoji",
          Hashtag: "as:Hashtag",
          quoteUrl: "as:quoteUrl",
          _misskey_quote: "misskey:_misskey_quote",
          quoteUri: "fedibird:quoteUri",
          emojiReactions: [Object]
        }
      ],
      id: "https://example.com/posts/post-1",
      type: "Note",
      content: "ActivityPub is a decentralized social networking protocol...",
      tag: [
        "https://example.com/tags/ActivityPub",
        "https://example.com/tags/Decentralization"
      ]
    },
    {
      "@context": [
        "https://www.w3.org/ns/activitystreams",
        "https://w3id.org/security/data-integrity/v1",
        {
          toot: "http://joinmastodon.org/ns#",
          misskey: "https://misskey-hub.net/ns#",
          fedibird: "http://fedibird.com/ns#",
          sensitive: "as:sensitive",
          Emoji: "toot:Emoji",
          Hashtag: "as:Hashtag",
          quoteUrl: "as:quoteUrl",
          _misskey_quote: "misskey:_misskey_quote",
          quoteUri: "fedibird:quoteUri",
          emojiReactions: [Object]
        }
      ],
      id: "https://example.com/posts/post-3",
      type: "Note",
      content: "WebFinger is a protocol for discovering information...",
      tag: "https://example.com/tags/ActivityPub"
    }
  ]
}
```

### DIFF
--- a/examples/custom-collections/main.ts
+++ b/examples/custom-collections/main.ts
@@ -43,18 +43,20 @@ const POSTS = [
   }),
 ];
 
-function getBookmarkedPosts(): Create[] {
-  return POSTS;
+function getPosts(): Promise<Create[]> {
+  return Promise.resolve(POSTS);
 }
 
-function getBookmarkCounts(): number {
-  return 5; // Total number of bookmarks for any user
+function getPostsCounts(): number {
+  return 5;
 }
 
-function getTaggedPostsByTag(tag: URL): Create[] {
-  return POSTS.filter((post) => {
-    return post.tagIds?.some((tagId) => tagId.toString() === tag.toString());
-  });
+function getTaggedPostsByTag(tag: string): Promise<Create[]> {
+  return Promise.resolve(
+    POSTS.filter((post) => {
+      return post.tagIds?.some((tagId) => tagId.toString() === tag);
+    }),
+  );
 }
 
 async function demonstrateCustomCollection() {
@@ -67,7 +69,7 @@ async function demonstrateCustomCollection() {
   });
 
   // Create a simple dispatcher function for testing
-  const dispatcher: CustomCollectionDispatcher<
+  const postsDispatcher: CustomCollectionDispatcher<
     Create,
     Record<string, string>,
     RequestContext<void>,
@@ -79,9 +81,7 @@ async function demonstrateCustomCollection() {
   ) => {
     if (values.handle !== "someone") return null;
 
-    // Get bookmarked posts for the user
-    const posts = await getBookmarkedPosts();
-
+    const posts = await getPosts();
     if (cursor != null) {
       const idx = parseInt(cursor);
       return {
@@ -94,13 +94,41 @@ async function demonstrateCustomCollection() {
     return { items: posts };
   };
 
+  const taggedPostsDispatcher: CustomCollectionDispatcher<
+    Create,
+    Record<string, string>,
+    RequestContext<void>,
+    void
+  > = async (
+    _ctx: { url: URL },
+    values: Record<string, string>,
+    cursor: string | null,
+  ) => {
+    if (values.handle !== "someone") return null;
+
+    const posts = await getTaggedPostsByTag(values.tag);
+
+    if (cursor != null) {
+      const idx = parseInt(cursor);
+      return {
+        items: [posts[idx]],
+        nextCursor: idx < posts.length - 1 ? (idx + 1).toString() : null,
+        prevCursor: idx > 0 ? (idx - 1).toString() : null,
+      };
+    }
+    return { items: posts };
+  };
+
   const counter: CustomCollectionCounter<Record<string, string>, void> = (
     _ctx: RequestContext<void>,
     values: Record<string, string>,
-  ) => values.handle === "someone" ? getBookmarkCounts() : 6;
+  ) => values.handle === "someone" ? getPostsCounts() : null;
 
-  const values = { handle: "someone" };
-  const posts = await dispatcher(context, values, null);
+  const values = {
+    handle: "someone",
+    tag: "https://example.com/tags/ActivityPub",
+  };
+  const posts = await postsDispatcher(context, values, null);
   console.log("All Posts for user: ", values.handle);
   console.log(posts);
 
@@ -108,9 +136,11 @@ async function demonstrateCustomCollection() {
   console.log("Count:", count);
 
   // Example of using a custom collection to get tagged posts
-  const tag = new URL("https://example.com/tags/ActivityPub");
-  const taggedPosts = getTaggedPostsByTag(tag);
-  console.log(`${tag.toString().split("/").pop()} Tagged Posts:`, taggedPosts);
+  const taggedPosts = await taggedPostsDispatcher(context, values, null);
+  console.log(
+    `${values.tag.toString().split("/").pop()} Tagged Posts:`,
+    taggedPosts,
+  );
 }
 
 if (import.meta.main) {
--- a/examples/custom-collections/main.ts
+++ b/examples/custom-collections/main.ts
@@ -1,12 +1,13 @@
 import {
   Create,
   createFederation,
-  CustomCollectionCounter,
   CustomCollectionDispatcher,
   MemoryKvStore,
   RequestContext,
 } from "@fedify/fedify";
-
+import {
+  handleCustomCollection,
+} from "../../packages/fedify/src/federation/handler.ts";
 import {
   createRequestContext,
 } from "../../packages/fedify/src/testing/context.ts";
@@ -43,14 +44,6 @@ const POSTS = [
   }),
 ];
 
-function getPosts(): Promise<Create[]> {
-  return Promise.resolve(POSTS);
-}
-
-function getPostsCounts(): number {
-  return 5;
-}
-
 function getTaggedPostsByTag(tag: string): Promise<Create[]> {
   return Promise.resolve(
     POSTS.filter((post) => {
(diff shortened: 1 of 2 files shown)


examples/custom-collections/main.ts
```diff
@@ -0,0 +1,118 @@
...
+  console.log("All Posts for user: ", values.handle);
+  console.log(posts);
+
+  const count = await counter(context, values);
+  console.log("Count:", count);
+
+  // Example of using a custom collection to get tagged posts
+  const tag = new URL("https://example.com/tags/ActivityPub");
+  const taggedPosts = getTaggedPostsByTag(tag);
+  console.log(`${tag.toString().split("/").pop()} Tagged Posts:`, taggedPosts);
```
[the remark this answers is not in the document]
ThisIsMissEm: This basically sums up my feedback, there's no call to federation's `setCollectionDispatcher()` and `setOrderedCollectionDispatcher()` methods here, but I'd expect there to be.

ThisIsMissEm (changes requested):
Should actually show using the dispatcher callbacks, so missing some `federation.set...` lines. The AI comments are also pretty good

examples/custom-collections/main.ts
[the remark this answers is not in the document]
sij411: Thanks for the review. I'll add them.

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,179 @@
...
+  CustomCollectionDispatcher,
+  MemoryKvStore,
+  RequestContext,
+} from "@fedify/fedify";
+import {
+  handleCustomCollection,
+} from "../../packages/fedify/src/federation/handler.ts";
+import {
+  createRequestContext,
+} from "../../packages/fedify/src/testing/context.ts";
```
dahlia: These APIs should be imported from `@fedify/fedify`.

dahlia (changes requested):

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,179 @@
...
+      }
+      return { items: posts };
+    },
+  ).setCounter(async (_ctx, values) => {
+    // Return the total count of bookmarked posts
+    const count = await getTaggedPostsCounts(values.tag);
+    return count;
+  });
+
+  const dispatcher: CustomCollectionDispatcher<
```
ThisIsMissEm: Where is this used? Looks like by here? handleCustomCollection

Is this needed or could we just make an actual request to fedify? (Well, inject one)

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,179 @@
...
+    async (
+      _ctx: { url: URL },
+      values: Record<string, string>,
+      cursor: string | null,
+    ) => {
+      if (values.handle !== "someone") return null; // Example handle check
+      const posts = await getTaggedPostsByTag(values.tag);
+
+      if (cursor != null) {
+        const idx = parseInt(cursor);
```
ThisIsMissEm:
```suggestion
        const idx = Number.parseInt(cursor, 10);
        
        If (Number.isNaN(idx)) // throw exception?
```

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,179 @@
...
+      return post.tagIds?.some((tagId) => tagId.toString() === tag);
+    }),
+  );
+}
+
+function getTaggedPostsCounts(tag: string): Promise<number> {
+  return Promise.resolve(
+    POSTS.filter((post) => {
+      return post.tagIds?.some((tagId) => tagId.toString() === tag);
+    }).length,
```
ThisIsMissEm: This should just call `getTaggedPostsByTag` then do length on that. No need to duplicate the logic

examples/custom-collections/main.ts
ThisIsMissEm: Like can we use `federation.fetch(..)` to test this in the demo?

examples/custom-collections/main.ts
ThisIsMissEm: What gemini said below is also a valid approach

ThisIsMissEm (changes requested):

examples/custom-collections/main.ts
sij411: I should fix some examples more since they're not included in `@fedify/fedify`

examples/custom-collections/main.ts
sij411: I'm afraid we can't use `federation.fetch(...)` with mock up data. I'll try with actual ones.

ThisIsMissEm: Oh? Why not..?

sij411: never mind it was routing problem (wrong `if` condition usage was the problem) 😅 It works with mockup data

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,123 @@
...
+      cursor: string | null,
+    ) => {
+      if (!values.userId || !values.tag) {
+        throw new Error("Missing userId or tag in values");
+      }
+      const posts = getTaggedPostsByTag(values.tag);
+      const items = posts.map((post) => (new Note({
+        id: new URL(`/posts/${post.id}`, _ctx.url),
+        content: post.content,
+      })));
```
ThisIsMissEm: Might be better to do this in the POSTS construction, since a `Create` activity doesn't usually have `content` and `tags`, those are normally on the `Note` that is being created

ThisIsMissEm (commented):
Minor note about Note and Create, but otherwise looking good.

examples/custom-collections/main.ts
sij411: Is it okay to use only `Note`

ThisIsMissEm: I think so? I'm not sure

sij411: @dahlia What do you think?

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,117 @@
+import { Create, createFederation, MemoryKvStore, Note } from "@fedify/fedify";
+import { print } from "ioredis";
```
ThisIsMissEm:
```suggestion
```

Unused import?

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,117 @@
...
+const POSTS = [
+  new Note({
+    id: new URL("https://example.com/posts/post-1"),
+    content: "ActivityPub is a decentralized social networking protocol...",
+    tags: [
+      new URL("https://example.com/tags/ActivityPub"),
+      new URL("https://example.com/tags/Decentralization"),
+    ],
+  }),
+  new Note({
```
ThisIsMissEm, line 14:
```suggestion

  new Note({
```

nit: inconsistent whitespace

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,117 @@
...
+    "/users/{userId}/tags/{tag}",
+    (
+      _ctx: { url: URL },
+      values: Record<string, string>,
+      cursor: string | null,
+    ) => {
+      if (!values.userId || !values.tag) {
+        throw new Error("Missing userId or tag in values");
+      }
+      const posts = getTaggedPostsByTag(values.tag);
```
ThisIsMissEm, line 71: Perhaps add a comment here:
```suggestion
      // Normally here you would look up posts from a database by user ID and tag name:
      const posts = getTaggedPostsByTag(values.tag);
```

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      if (cursor != null) {
+        const idx = Number.parseInt(cursor, 10);
+        if (Number.isNaN(idx)) {
+          throw new Error("Invalid cursor");
```
ThisIsMissEm: Perhaps:
```suggestion
          return { items: [], nextCursor: null, prevCursor: null };
```

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,117 @@
...
+      cursor: string | null,
+    ) => {
+      if (!values.userId || !values.tag) {
+        throw new Error("Missing userId or tag in values");
+      }
+      const posts = getTaggedPostsByTag(values.tag);
+
+      if (cursor != null) {
+        const idx = Number.parseInt(cursor, 10);
+        if (Number.isNaN(idx)) {
```
ThisIsMissEm:
```suggestion
        if (Number.isNaN(idx) || idx > posts.length || idx < 0) {
```

(not sure on the last condition, not sure if parseInt would ever return a negative number off top of head)

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,117 @@
...
+  federation.setCollectionDispatcher(
+    "TaggedPosts",
+    Note,
+    "/users/{userId}/tags/{tag}",
+    (
+      _ctx: { url: URL },
+      values: Record<string, string>,
+      cursor: string | null,
+    ) => {
+      if (!values.userId || !values.tag) {
```
ThisIsMissEm: We could also drop the check for `values.userId` since it's not relevant to this test case?

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,117 @@
...
+    },
+  ).setCounter(async (_ctx, values) => {
+    // Return the total count of tagged posts
+    const count = getTaggedPostsByTag(values.tag).length;
+    return count;
+  });
+
+  const response = await federation.fetch(
+    new Request(
+      "https://example.com/users/someone/tags/ActivityPub",
```
ThisIsMissEm: Else where in the documentation we recommend using identifiers for users instead of usernames, so perhaps:
```suggestion
      "https://example.com/users/123/tags/ActivityPub",
```

examples/custom-collections/main.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  console.log("Custom collection response status:", response.status);
+
+  if (response.ok) {
+    const jsonResponse = await response.json();
+    console.log("Custom collection data:", jsonResponse);
+  } else {
+    const errorText = await response.text();
+    console.log("Error response:", errorText);
+  }
```
ThisIsMissEm: I think this should probably be moved to the `if(import.meta.main)`?

ThisIsMissEm: or that and the fetch above?

ThisIsMissEm (commented):
I think Note is good on second pass, but I noticed some other things.

examples/custom-collections/main.ts
sij411: How come I didn't notice 😩 probably ghost writings

examples/custom-collections/main.ts
sij411: Oh yeah good catch. Thanks!

examples/custom-collections/main.ts
sij411: Mm only with main func sounds good? Handling both main and fetch might be redundant. I'll see. Huge thanks :)

examples/custom-collections/main.ts
dahlia: See also: <https://fedify.dev/manual/actor#actor-identifier-and-webfinger-username>.

examples/custom-collections/main.ts
dahlia: IMHO it's better to use only `Note` in the `POSTS` construction.

examples/custom-collections/main.ts
sij411: From my understanding, we should use usernames?  
> By default, Fedify assumes that the actor's identifier is the WebFinger username,

dahlia (approved):

PR #378, opened 2025-08-22
 feat(test): Add unit test for `fedify tunnel`

dodok8:
## Summary

Add unit test for `fedify tunnel`

## Related Issue

This is part of issue #259, to add tests to fedify/cli

## Changes

- refactor: extract tunnel creation logic into a separate function for testing
- dep: add @std/testing
- test: add comprehensive tests for tunnel functionality and error handling

## Benefits

unit test for `tunnel.ts`

## Checklist

- [ ] Did you add a changelog entry to the *CHANGES.md*?
- [ ] Did you write some relevant docs about this change (if it's a new feature)?
- [ ] Did you write a regression test to reproduce the bug (if it's a bug fix)?
- [x] Did you write some tests for this change (if it's a new feature)?
- [x] Did you run `deno task test-all` on your machine?

## Additional Notes

│ (fail) Activity.getObjects() [60819.63ms]
│ (fail) InProcessMessageQueue [5535.71ms]

These two tests failed on my machine on bun 1.2.18, WSL Ubuntu 24.04.2 LTㄴ.

PR #390, opened 2025-08-29
 Migrate cli `lookup` from Cliffy to Optique 

### LINKED ISSUE #374: CLI for Fedify 2.0: Cross-runtime transition and Optique adoption
The current Fedify CLI implementation has been working well for our Deno-focused ecosystem, but as we prepare for Fedify 2.0, we're hitting some architectural limitations that are becoming increasingly apparent. Right now, our CLI is built with [Cliffy](https://cliffy.io/), which means it only runs natively on Deno. For Node.js and Bun users who install via npm, they're actually downloading a compiled binary produced by `deno compile` rather than getting the actual JavaScript source. This feels somewhat awkward for what is fundamentally a JavaScript-based tool.

The compiled binary approach has served us reasonably well, but it creates a disconnect between the tool's JavaScript nature and how users actually consume it. Node.js developers expect to install JavaScript packages that run in their runtime, not pre-compiled executables. Additionally, Cliffy has some quirks that we've been working around—most notably the ANSI escape sequence issue that we had to fix in #257/#341, where it was outputting color codes even when stdout isn't a terminal. It also struggles with more complex CLI structures that we anticipate needing as Fedify's feature set expands.

For Fedify 2.0, we want to provide a more natural experience across all major JavaScript runtimes. This means the CLI should run natively on Deno, Node.js, and Bun without requiring compilation or runtime-specific workarounds. To address these needs, I've been developing [Optique](https://optique.dev/), a type-safe combinatorial CLI parser specifically designed with Fedify's requirements in mind. You can read more about the motivation and design principles behind Optique in the [*Why Optique?*](https://optique.dev/why) documentation. Optique handles cross-runtime compatibility out of the box and provides the composable architecture we need for building more sophisticated CLI interfaces.

The migration to Optique would not only solve our immediate cross-runtime needs but also position us well for the more complex CLI features we're planning for Fedify 2.0. The parser combinators approach should give us the flexibility to handle nested subcommands and reusable option groups in a way that's both maintainable and type-safe. I think this transition aligns well with our broader goals of making Fedify more accessible across the JavaScript ecosystem.

sij411:
## Summary

Migrated fedify/cli `lookup` and its test from [Cliffy](https://cliffy.io/) to [Optique](https://optique.dev/)

## Related Issue

- closes #374 

## Changes

- Created new package `cli-optique`. This will be a working directory for migrating entire `Cliffy` codebase to `Optique`.
- Re-wrote  `mod.ts`, `lookup.ts` and `lookup.test.ts` with `optique` 
- Replaced Deno API  to Node API in `lookup.ts`.
- Added placeholder scripts for development
	- `init.ts`, `nodeinfo.ts`, `tunnel.ts`, `webfinger.ts`, `inbox.ts`

## Benefits

With this migration, `fedify/cli` will run natively all major runtimes such as Node.js, Deno, and Bun. Also thanks to `optique` is type-safety and other features, future cli development and maintenance would be easier. 
## Checklist

## Additional Notes

- I import modules such as `docloader.ts` and 'tempserver.ts` from original `packages/cli`  
- CI failed due to relative import, I copied multiple scripts from original `cli/`


sij411:
Should I upgrade node version from 24.2.1 to 24.3.o (pnpm-lock.yaml)

dahlia:
No, you don't have to.

packages/cli-optique/deno.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@fedify/cli-optique",
```
dahlia: This should be renamed to `@fedify/cli` eventually (before merging this pull request).

packages/cli-optique/deno.json
```diff
@@ -0,0 +1,35 @@
...
+    "@cross/dir": "jsr:@cross/dir@^1.1.0",
+    "@optique/core": "jsr:@optique/core@^0.3.0",
+    "@optique/run": "jsr:@optique/run@^0.3.0",
+    "@std/fmt": "jsr:@std/fmt@^1.0.8",
+    "ora": "npm:ora@^8.2.0",
+    "@hongminhee/localtunnel": "jsr:@hongminhee/localtunnel@^0.2.0",
+    "icojs": "npm:icojs@^0.19.4",
+    "@jimp/wasm-webp": "npm:@jimp/wasm-webp@^1.6.0",
+    "@jimp/core": "npm:@jimp/core@^1.6.0",
+    "@std/fmt/colors": "jsr:@std/fmt@^0.224.0/colors"
```
dahlia: I believe some of them are no more necessary, e.g., Cliffy.

packages/cli-optique/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@fedify/cli-optique",
```
dahlia: This should be renamed to `@fedify/cli` eventually (before merging this pull request).

dahlia (commented):
Could you replace `console.log()`/`console.error()` calls with `print()`/`printError()` provided by Optique? See also: <https://optique.dev/concepts/messages#terminal-output>.

packages/cli-optique/deno.json
sij411: What about existing cli?

dahlia: That should be removed!

sij411: Alright. Then should I move all the other scripts into cli-optique/ ?

dahlia: No, that's okay! You should open some sub-issues although.

sij411: Got it.

dahlia:
Seems your commits have conflicts with the current *next* branch.

packages/cli/deno.json
```diff
@@ -51,18 +29,12 @@
...
-      ]
-    }
+    "runi": "deno run --allow-all src/mod.ts"
+  },
+  "pack": {
+    "command": "deno run -A scripts/pack.ts",
+    "dependencies": [
+      "codegen"
+    ]
   }
```
dahlia: I think it's in the wrong place. It should be next to the `"runi"` field.

packages/cli/src/mod.ts
2chanhaeng, line 1:
```ts
// ...
import inbox from "./inbox.ts";
import init from "./init.ts";
import lookup from "./lookup.ts";
import nodeinfo from "./nodeinfo.ts";
import tunnel from "./tunnel.ts";
import webfinger from "./webfinger.ts";

const cliList = [
  init,
  lookup,
  inbox,
  nodeinfo,
  tunnel,
  webfinger,
] as const;


const command = or(cliList.map((cli) => cli.command));

async function main() {
  const result = run(command, {
    programName: "fedify",
    help: "both",
  });
  await cliList.find((cli) => cli.command === result.command)?.run(result);
}

await main();
```

2chanhaeng (commented):
Maybe it would be better to change the modules to only return `command` and `run`, then write the code like this for better consistency?

packages/cli/src/mod.ts
2chanhaeng: The name `run` is already taken by "@optique/run", so maybe `main` is better?

sij411: Sorry but it seems `await cliList.find((cli) => cli.command === result.command)?.run(result)` wouldn't work because the variable `result`'s type is a parser. Since the actual functions that do the job aren't running, the cli result wouldn't be as same as we expected to be.  I agree with that approach would bring us simpler looks tho.
See also https://optique.dev/concepts/runners#high-level-execution-with-optique-run

dahlia: I don't think @2chanhaeng's approach would work as each parser has its own unique type; they are not homogeneous. In other words, they cannot be put together in an array in a type-safe way.

packages/cli/src/tunnel.ts
```diff
@@ -1,53 +1,27 @@
...
-    openTunnel,
-    ora,
-    console,
-    addSignalListener: Deno.addSignalListener,
-    exit: Deno.exit,
+export const tunnelCommand = command(
+  "tunnel",
+  object({
+    command: constant("tunnel"),
+    resources: argument(string({ metavar: "PORT" })),
```
dahlia: Port numbers are integers from 0 to 65,535:

```suggestion
    port: argument(integer({ metavar: "PORT", min: 0, max: 65_535 })),
```

packages/cli/package.json
```diff
@@ -51,5 +27,26 @@
...
+    "@fedify/fedify": "workspace:",
+    "@optique/core": "^0.3.0",
+    "@optique/run": "^0.3.0",
+    "cli-highlight": "^2.1.11",
+    "jimp": "^1.6.0",
+    "@jimp/core": "^1.6.0",
+    "@jimp/wasm-webp": "^1.6.0",
+    "ora": "^8.2.0",
+    "@cross/dir": "jsr:@cross/dir@^1.1.0",
+    "cli-highlight": "^2.1.11",
```
2chanhaeng: `cli-highlight` is duplicated.

packages/cli/package.json
```diff
@@ -51,5 +27,26 @@
...
+    "@hongminhee/localtunnel": "jsr:@hongminhee/localtunnel@^0.2.0",
```
dahlia: I ported this package to Node.js and Bun, and published it to npm (besides JSR):

```suggestion
    "@hongminhee/localtunnel": "^0.3.0",
```

packages/cli/deno.json
```diff
@@ -4,38 +4,21 @@
...
-    "@cliffy/prompt": "jsr:@cliffy/prompt@1.0.0-rc.4",
-    "@cliffy/table": "jsr:@cliffy/table@1.0.0-rc.4",
     "@cross/dir": "jsr:@cross/dir@^1.1.0",
-    "@david/dax": "jsr:@david/dax@^0.41.0",
+    "@david/dax": "jsr:@david/dax@^0.43.2",
+    "@optique/core": "jsr:@optique/core@^0.3.0",
+    "@optique/run": "jsr:@optique/run@^0.3.0",
+    "@std/fmt": "jsr:@std/fmt@^1.0.8",
+    "ora": "npm:ora@^8.2.0",
     "@hongminhee/localtunnel": "jsr:@hongminhee/localtunnel@^0.2.0",
```
dahlia:
```suggestion
    "@hongminhee/localtunnel": "jsr:@hongminhee/localtunnel@^0.3.0",
```

packages/cli/src/nodeinfo.ts
```diff
@@ -1,403 +1,34 @@
-import { Command } from "@cliffy/command";
-import { formatSemVer, getNodeInfo, getUserAgent } from "@fedify/fedify";
+import {
+  command as OQCommand,
+  constant,
+  type InferValue,
+  message,
+  object as OQObject,
+} from "@optique/core";
```
dahlia: Why are `command` and `object` aliased to `OQCommand` and `OQObject` in this module?

packages/cli/src/init.ts
```diff
@@ -1,1636 +1,26 @@
...
+  message,
+  object,
+} from "@optique/core";
+import { path } from "@optique/run";
+
+export const initCommand = command(
+  "init",
+  object({
+    command: constant("init"),
+    resources: argument(path({ metavar: "DIRECTORY" })),
```
dahlia: For metavariables, `DIR` is widely used. (I don't know why either though.)

```suggestion
    resources: argument(path({ metavar: "DIR" })),
```

packages/cli/src/log.ts
dahlia, line 1: I think this file shouldn't be deleted?

packages/cli/src/lookup.ts
```diff
@@ -14,75 +32,122 @@ import {
...
+
+const traverseOption = withDefault(
+  object({
+    traverse: flag("-t", "--traverse", {
+      description:
+        message`Traverse the given collection to fetch all items. If it is turned on, the argument cannot be multiple.`,
+    }),
+    suppressErrors: option("-S", "--suppress-errors", {
+      description:
+        message`Suppress partial errors while traversing the collection`,
```
dahlia: I know I'm picky:

```suggestion
        message`Suppress partial errors while traversing the collection.`,
```

packages/cli/src/lookup.ts
```diff
@@ -14,75 +32,122 @@ import {
...
+    }),
+  }),
+  { traverse: false } as const,
+);
+
+export const lookupCommand = merge(
```
2chanhaeng: You must wrap this with `command("webfinger", ...)` or CLI will not work.

2chanhaeng: You can test with `deno task cli lookup @2chanhaeng@hackers.pub` or other handles.

packages/cli/src/log.ts
sij411: The reason I deleted for now is I see no available options for logging in Optique `run()` from `@optique/run` . How do i log with optique?

packages/cli/src/lookup.ts
sij411: That's interesting. I run the script with `cd packages/cli/src && deno task run lookup {handle}` and it works fine for me. @dahlia Could you explain the reason behind this, please? [`command()`](https://github.com/dahlia/optique/blob/main/packages/core/src/parser.ts#L3131) looks a wrapper of parser for me...

packages/cli/src/lookup.ts
sij411: No, it's a good catch!

packages/cli/src/log.ts
dahlia: We have a global option named `-d`/`--debug` which decrease the `lowestLevel` to emit:

https://github.com/fedify-dev/fedify/blob/1f1074b0d8d664a71b4e353d86dd05d653f4fd07/packages/cli/src/mod.ts#L30-L60

Logs are not made by CLI itself, but *@fedify/fedify*, and these logs would appear from the `fedify` command with `-d`/`--debug` option.

sij411: From my understandings, I should implement a separate option somewhere in `mod.ts` ? Or do you have any optique examples I can look up?

packages/cli/src/lookup.ts
dahlia: @2chanhaeng's apparently right here. When you invoke `deno task run lookup {handle}`, the token `lookup` is treated as an argument, not a subcommand. Hence, currently the below command works as well:

```sh
deno task run {handle}  # No "lookup" subcommand here.
```

packages/cli/src/log.ts
dahlia: Hmm, I should add some examples to implement global options in Optique, but there's no example currently. Yeah, you should implement option `-d`/`--debug` in *mod.ts*.

sij411: Alright, I'll handle the logging tasks and pass the tunnel-related items to the person assigned to #378.

packages/cli/src/lookup.ts
sij411: Aha, got it. Thanks.

packages/cli/src/log.ts
```diff
@@ -33,9 +35,9 @@ export function getRecordingSink(): RecordingSink {
 
 export const recordingSink = getRecordingSink();
 
-export const logFile = Deno.env.get("FEDIFY_LOG_FILE");
+export const logFile = process.env["FEDIFY_LOG_FILE"];
 if (logFile != null) {
-  await Deno.mkdir(dirname(logFile), { recursive: true });
+  mkdir(dirname(logFile), { recursive: true });
```
dahlia: Shouldn't this be awaited?

```suggestion
  await mkdir(dirname(logFile), { recursive: true });
```

packages/cli/src/globals.ts
```diff
@@ -0,0 +1,43 @@
+import { message, object, option } from "@optique/core";
+import { configure, getConsoleSink } from "@logtape/logtape";
+import { getFileSink } from "@logtape/file";
+import { recordingSink } from "./log.ts";
+import { AsyncLocalStorage } from "node:async_hooks";
+import process from "node:process";
+
+export const debugOption = object("debug", {
```
dahlia: More appropriate labeling (it's shown on the `--help` message):

```suggestion
export const debugOption = object("Global options", {
```

packages/cli/src/lookup.ts
```diff
@@ -205,299 +262,279 @@ function wrapDocumentLoaderWithTimeout(
...
-      discardStdin: false,
-    }).start();
-    let server: TemporaryServer | undefined = undefined;
-    const baseDocumentLoader = await getDocumentLoader({
-      userAgent: options.userAgent,
-    });
-    const documentLoader = wrapDocumentLoaderWithTimeout(
-      baseDocumentLoader,
-      options.timeout,
+export async function runLookup(command: InferValue<typeof lookupCommand>) {
```
dahlia, line 277: We should call `configureLogging()` here if `-d`/`--debug` is present.

packages/cli/src/lookup.ts
```diff
@@ -14,75 +33,130 @@ import {
...
+);
+
+export const lookupCommand = command(
+  "lookup",
+  merge(
+    object({ command: constant("lookup") }),
+    traverseOption,
+    authorizedFetchOption,
+    debugOption,
+    object("lookup", {
```
dahlia: More appropriate labeling (it's shown on the `--help` message):

```suggestion
    object("Looking up options", {
```

packages/cli/src/tunnel.ts
dahlia: @sij411 This also should be addressed!

dahlia (approved):
Great work, thanks!

PR #392, opened 2025-08-30
Migrate from `LanguageTag` to `Intl.Locale` for representing language tags

### LINKED ISSUE #280: Migrate from @phensley/language-tag to `Intl.Locale` for Fedify 2.0
## Problem

Currently, Fedify uses the third-party `@phensley/language-tag` package and its `LanguageTag` class to represent language tags and locales. For Fedify 2.0, we should migrate to the standardized `Intl.Locale` class which is part of the web platform and provides better compatibility across different JavaScript environments.

## Benefits

- **Standard API**: `Intl.Locale` is a web standard, ensuring better long-term compatibility
- **Reduced dependencies**: Eliminates the need for the `@phensley/language-tag` third-party package
- **Better platform support**: Native support across Deno, Node.js, and Bun environments
- **Performance**: Native implementation is likely more performant than userland code

## Tasks

- [ ] Audit current usage of `@phensley/language-tag` and `LanguageTag` throughout the codebase
- [ ] Replace `LanguageTag` with `Intl.Locale` in all relevant code
- [ ] Update type definitions and interfaces that reference `LanguageTag`
- [ ] Update tests to use `Intl.Locale` instead of `LanguageTag`
- [ ] Remove `@phensley/language-tag` dependency from package.json
- [ ] Update documentation and examples to reflect the new API
- [ ] Verify compatibility across all supported environments (Deno, Node.js, Bun)

## Breaking Changes

This will be a breaking change for Fedify 2.0 as any public APIs that currently accept or return `LanguageTag` will now use `Intl.Locale` instead.

## Migration Guide

Users will need to:
- Replace `LanguageTag` imports with `Intl.Locale`
- Update any code that creates or manipulates language tags to use the `Intl.Locale` constructor and methods
- Review any custom code that depends on `LanguageTag`-specific methods or properties

## Target

This should be completed as part of the Fedify 2.0 release.

Palcimer:
## Summary

Migrate from `LanguageTag` to `Intl.Locale` for representing language tags

## Related Issue

- closes #280 

## Changes

- Replaced `LanguageTag` with `Intl.Locale` in the `LanguageString` class. 
- Updated corresponding ScalarType (`fedify:langTag`)
- Updated tests
- Removed dependency for @phensley/language-tag

## Benefits

- Enhanced performance and compatibility through the use of native Intl.Locale API instead of a third-party package.

## Checklist

## Additional Notes

- Renamed the property that holds the language tag from `language` to `locale` to eliminate the awkward repetition in the expression `this.language.language`, and I wonder if the renaming was appropriate.
- Updating `CHANGES.md`


packages/fedify/src/codegen/type.ts
```diff
@@ -155,7 +155,7 @@ const scalarTypes: Record<string, ScalarType> = {
     encoder(v) {
       return `{
         "@value": ${v}.toString(),
-        "@language": ${v}.language.compact(),
+        "@language": ${v}.locale.language,
```
dahlia: You should use `baseName` here instead of `language`, because it loses other information like region or script for non-trivial locale code, e.g., `ko-KR`, `zh-Hant-CN`:

```ts
> const locale = new Intl.Locale("ko-KR");
undefined
> locale.language
"ko"
> locale.baseName
"ko-KR"
```

packages/fedify/src/codegen/type.ts
```diff
@@ -262,22 +262,22 @@ const scalarTypes: Record<string, ScalarType> = {
...
-      return `${v} instanceof LanguageTag`;
+      return `${v} instanceof Intl.Locale`;
     },
     encoder(v) {
-      return `{ "@value": ${v}.compact() }`;
+      return `{ "@value": ${v}.language }`;
     },
     compactEncoder(v) {
-      return `${v}.compact()`;
+      return `${v}.language`;
```
dahlia: You should use `baseName` instead here, too.

packages/fedify/src/runtime/langstr.ts
```diff
@@ -1,36 +1,34 @@
-import { type LanguageTag, parseLanguageTag } from "@phensley/language-tag";
-
 /**
  * A language-tagged string which corresponds to the `rdf:langString` type.
  */
 export class LanguageString extends String {
-  readonly language: LanguageTag;
+  readonly locale: Intl.Locale;
```
dahlia, line 6: I do like this renaming, but we should write an changelog entry about this change.

Also, we need a JSDoc comment here!

packages/fedify/src/runtime/langstr.ts
```diff
@@ -1,36 +1,34 @@
...
+      ? new Intl.Locale(language)
       : language;
   }
 
   [Symbol.for("Deno.customInspect")](
     inspect: typeof Deno.inspect,
     options: Deno.InspectOptions,
   ): string {
-    return `<${this.language.compact()}> ${inspect(this.toString(), options)}`;
+    return `<${this.locale.language}> ${inspect(this.toString(), options)}`;
```
dahlia: You should use `baseName` instead here, too.

packages/fedify/src/runtime/langstr.ts
```diff
@@ -1,36 +1,34 @@
...
   }
 
   [Symbol.for("nodejs.util.inspect.custom")](
     _depth: number,
     options: unknown,
     inspect: (value: unknown, options: unknown) => string,
   ): string {
-    return `<${this.language.compact()}> ${inspect(this.toString(), options)}`;
+    return `<${this.locale.language}> ${inspect(this.toString(), options)}`;
```
dahlia: You should use `baseName` instead here, too.

packages/fedify/src/vocab/vocab.test.ts
```diff
@@ -890,7 +889,7 @@ const sampleValues: Record<string, any> = {
     0x30, 0x65, 0x59, 0x76, 0xf0, 0xb0, 0xdb, 0x78, 0x0d, 0xf5, 0xc1, 0xad,
     0x3f, 0xbd, 0xf3, 0x07,
   ]),
-  "fedify:langTag": parseLanguageTag("en"),
+  "fedify:langTag": new Intl.Locale("en"),
```
dahlia: It would be better if we can test whether it works well with non-trivial locale code, e.g., `en-Latn-US`:

```suggestion
  "fedify:langTag": new Intl.Locale("en-Latn-US"),
```

packages/fedify/src/runtime/langstr.ts
```diff
@@ -1,36 +1,34 @@
...
-  readonly language: LanguageTag;
+  readonly locale: Intl.Locale;
 
   /**
    * Constructs a new `LanguageString`.
    * @param value A string value written in the given language.
-   * @param language The language of the string.  If a string is given, it will
-   *                 be parsed as a `LanguageTag`.
+   * @param locale The language of the string.  If a string is given, it will
+   *                 be parsed as a Intl.Locale object.
```
dahlia:
```suggestion
   *                 be parsed as a `Intl.Locale` object.
```

packages/fedify/src/codegen/__snapshots__/class.test.ts.snap
dahlia, line 1: For your information: You can update them by running `deno task test -- --update` inside the *packages/fedify/* directory.

packages/fedify/src/vocab/__snapshots__/vocab.test.ts.snap
dahlia, line 1: For your information: You can update them by running `deno task test -- --update` inside the *packages/fedify/* directory.

dahlia (changes requested):

dahlia (approved):
Looks good to me.  Thanks!

PR #394, opened 2025-08-31
Fixed SvelteKit integration hook types

### LINKED ISSUE #271: SvelteKit integration hook has the wrong type
The [SvelteKit hook](https://github.com/fedify-dev/fedify/blob/main/fedify/x/sveltekit.ts) in `@fedify/fedify/x/sveltekit` has the type:

```ts
type RequestEvent = {
  request: Request
}

type HookParams = {
  event: RequestEvent 
  //     ^ covariant
  resolve: (event: RequestEvent) => Promise<Response>
  //               ^ contravariant
}

fedifyHook(...): (params: HookParams) => Promise<Response>
```

Notice two points here:

- The `RequestEvent` type is defined locally and is a subtype of [the actual type in SvelteKit](https://svelte.dev/docs/kit/@sveltejs-kit#RequestEvent)
- `RequestEvent` is *invariant* in the type of `fedifyHook(...)`, so the type of `fedifyHook(...)` is not a subtype nor a supertype of [`Handle`, the handler hook type in SvelteKit](https://svelte.dev/docs/kit/@sveltejs-kit#RequestEvent).

Indeed, TypeScript complains if you try to mark it as a `Handle`:
```ts
import { fedifyHook } from '@fedify/fedify/x/svelte'
import type { Handle } from '@sveltejs/kit'

const handleActivityPub = fedifyHook(federation, makeContextData) as Handle
//                                                                ^^^^^^^^^
// TS2352: Conversion of type '(params: HookParams) => Promise<Response>' to type 
// 'Handle' may be a mistake because neither type sufficiently overlaps with the other. 
// If this was intentional, convert the expression to 'unknown' first.
```

I suppose the most straightforward way to solve this is to add `@sveltejs/kit` to devDependencies and then use the actual `Handle` type from there, something like:

```ts
import type { Handle, RequestEvent } from '@sveltejs/kit'

export function fedifyHook<TContextData>(
  federation: Federation<TContextData>,
  createContextData: (
    event: RequestEvent,
  ) => TContextData | Promise<TContextData>,
): Handle {
  return async ({ event, resolve }) => {
    // -- snip --
  };
}
```

---

PS. I see the AI labeler adding more and more labels to my issue every time I edit it. You may want to fix that too!

2chanhaeng:
## Summary

Fixed SvelteKit integration hook types.

## Related Issue

- fixes #271

## Changes

- `packages/sveltekit`

## Benefits

Now users who use SvelteKit integration can get proper hook types.

## Checklist

- ~~[ ] Did you write some relevant docs about this change (if it's a new feature)?~~
- ~~[ ] Did you write some tests for this change (if it's a new feature)?~~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -116,6 +116,8 @@ To be released.
  -  Created SvelteKit integration as the *@fedify/sveltekit* package.
     Separated from `@fedify/fedify/x/sveltekit` to improve modularity and
     reduce bundle size.  [[#375] by Chanhaeng Lee]
+ -  Fixed SvelteKit integration hook types to correctly infer the request
+    and response types in hooks.  [ by Chanhaeng Lee]
 
 
 Version 1.8.8
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -117,7 +117,10 @@ To be released.
     Separated from `@fedify/fedify/x/sveltekit` to improve modularity and
     reduce bundle size.  [[#375] by Chanhaeng Lee]
  -  Fixed SvelteKit integration hook types to correctly infer the request
-    and response types in hooks.  [ by Chanhaeng Lee]
+    and response types in hooks.  [[#271], [#394] by Chanhaeng Lee]
+
+[#271]: https://github.com/fedify-dev/fedify/pull/271
+[#394]: https://github.com/fedify-dev/fedify/pull/394
 
 
 Version 1.8.8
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -116,6 +116,8 @@ To be released.
  -  Created SvelteKit integration as the *@fedify/sveltekit* package.
     Separated from `@fedify/fedify/x/sveltekit` to improve modularity and
     reduce bundle size.  [[#375] by Chanhaeng Lee]
+ -  Fixed SvelteKit integration hook types to correctly infer the request
+    and response types in hooks.  [ by Chanhaeng Lee]
```
dahlia: Please fill your pull request number here!

PR #401, opened 2025-09-01
CLI: Add webfinger command using Optique

### LINKED ISSUE #398: CLI: Add `webfinger` command using Optique
### Summary

Add `webfinger` command to cli with [Optique](https://optique.dev/) instead of cliffy.  

### Problem

For now, `webfinger.ts` has an empty function only. It should be implemented to run as same as original cli with cliffy does, but with Optique. Also it should be cross-runtime so existing Deno API should be replaced to Node API.

### Proposed Solution

1. Add [`webfinger.ts`](https://github.com/fedify-dev/fedify/blob/main/packages/cli/src/webfinger.ts) 

### Alternatives Considered

.

### Scope / Dependencies

- `fedify/packages/cli/src/mod.ts`
- `fedify/packages/cli/src/webfinger.ts`

2chanhaeng:
## Summary

Add webfinger command using Optique.

## Related Issue

- closes #398

## Changes

- `packages/cli/webfinger.ts`

## Benefits

Now Fedify 2.0 lookup webfinger using Optique

## Checklist

  I will add it after #390
- [ ] ~~Did you write some relevant docs about this change (if it's a new feature)?~~
- [ ] ~~Did you write a regression test to reproduce the bug (if it's a bug fix)?~~

## Additional Notes

Merge after #390


packages/cli/src/webfinger.test.ts
```diff
@@ -0,0 +1,85 @@
+import { lookupWebFinger } from "@fedify/fedify/webfinger";
+import { run } from "@optique/run";
```
dahlia: For unit testing, you'll probably want rather `parse()` from `@optique/core/parser`. See also [*Runners and execution*](https://optique.dev/concepts/runners) section in the docs.

2chanhaeng: Oh, that's the problem! Thanks!!!

dahlia (changes requested):
It would be great if we can utilize [Optique's rich message formatter](https://optique.dev/concepts/messages#terminal-output).

2chanhaeng:
> It would be great if we can utilize [Optique's rich message formatter](https://optique.dev/concepts/messages#terminal-output).

For that, we need a spinner library that can replace `ora` and work with `optique`.

packages/cli/src/webfinger/error.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @returns {string} A descriptive error message.
+ */
+export const getErrorMessage = (resource: string, error: unknown): Message =>
+  error instanceof InvalidHandleError
+    ? message`Invalid handle format: ${error.handle}`
+    : error instanceof NotFoundError
+    ? message`Resource not found: ${error.resource}`
+    : error instanceof Error
+    ? message`Error looking up WebFinger for ${resource}: ${error.message}`
+    : message`Error looking up WebFinger for ${resource}: ${String(error)}`;
```
dahlia: It's trivial, but as `printError()` prepends `Error: ` by default, it might sound unnatural: `Error: Error looking up WebFinger for ...: ...`.

```suggestion
    ? message`Failed to look up WebFinger for ${resource}: ${error.message}`
    : message`Failed to look up WebFinger for ${resource}: ${String(error)}`;
```

2chanhaeng: Changed at ede5d077b028af898e00afaf5ec4677d64ef5f9c!

dahlia (approved):

Issue #119, opened 2024-08-23
Webfinger in Fedify should support Mastodon's special `rel= 'http://ostatus.org/schema/1.0/subscribe'` and `template`


mattes3:
When you webfinger an account on a Mastodon instance, you get a list of links, one of them has a template:

```json
{
  "subject":"acct:sundogplanets@mastodon.social",
  "aliases":["https://mastodon.social/@sundogplanets","https://mastodon.social/users/sundogplanets"],
  "links":[
    {"rel":"http://webfinger.net/rel/profile-page","type":"text/html","href":"https://mastodon.social/@sundogplanets"},
    {"rel":"self","type":"application/activity+json","href":"https://mastodon.social/users/sundogplanets"},
    {"rel":"http://ostatus.org/schema/1.0/subscribe","template":"https://mastodon.social/authorize_interaction?uri={uri}"},
    {"rel":"http://webfinger.net/rel/avatar","type":"image/jpeg","href":"https://files.mastodon.social/accounts/avatars/108/194/428/751/240/116/original/f1eb397ab1659575.jpeg"}
  ]
}
```

The link with the template can be used to simplify a federated `follow` request. When you replace the `{uri}` parameter in the string by a Fediverse handle like `username@domain`, the link redirects you to their profile page as seen from the perspective of the other Mastodon server, e.g. mastodon.social in this case.

On that `mirrored` profile page, you can simply click the `Follow` button, and you're done.

I wanted to make this work with Fedify so that other people can easily follow me on my app, but I couldn't figure out how to add such a template-based URL to the `urls` collection of my actor object, so that Fedify's webfinger endpoint would show it.

The problem was that the `Link` class in the vocabulary does not support the (Mastodon-only) `template` field. Could you add this in Fedify, as a non-standard feature?

You can see a good application of this mechanism in Guppe: https://github.com/immers-space/guppe/blob/7d44a6df88a4a60e05ad6360aac6835e291b0ced/web/src/views/Profile.vue#L88

Guppe supports this "reverse follow" to make it easy to follow a gup.pe group. It would be great if Fedify-based apps would support this out of the box.

Does this all make sense? 😀

dahlia:
In fact, I've been working on a feature that would allow you to customize WebFinger responses, but it won't happen anytime soon.

dodok8:
I'll work on this. This is related to <https://github.com/fedify-dev/botkit/pull/14>.

mattes3:
[...]

PR #403, opened 2025-09-01
type: change `ResourceDescriptor` and `Link` interfaces

### LINKED ISSUE #402: type: extend `Link` type and `ResourceDescriptor`
### Summary

- Add a new type or add a 'template' field to `Link`
- Make `href` field in `Link` optional

### Problem

Before handling #119, we need to extend our type definition of the return type of `lookupWebFinger()`, `ResourceDescriptor` that supports non-RFC standard, [OStatus 1.0 Draft 2](https://www.w3.org/community/ostatus/wiki/images/9/93/OStatus_1.0_Draft_2.pdf).

And [According to the RFC 7033 Section 4.4.4.3](https://datatracker.ietf.org/doc/html/rfc7033#section-4.4.4.3), href is also OPTIONAL, so it needed to be changed.

### Proposed Solution

Add type like OStatusSubscribeLink

```typescript
export interface ResourceDescriptor {
  subject?: string;
  aliases?: string[];
  properties?: Record<string, string>;
  links?: Array<Link | OStatusSubscribeLink>;
}

/**
 * References a link. See also
 * [OStatus 1.0 Draft 2](https://www.w3.org/community/ostatus/wiki/images/9/93/OStatus_1.0_Draft_2.pdf)
 */
export interface OStatusSubscribeLink {
  rel: "http://ostatus.org/schema/1.0/subscribe";
  /**
   * A URI template (RFC 6570) that can be used to construct URIs by
   * substituting variables. Used primarily for subscription endpoints
   * where parameters like account URIs need to be dynamically inserted.
   */
  template: string;
}
```

### Alternatives Considered

```typescript
/**
 * Describes a resource.  See also
 * [RFC 7033 section 4.4](https://datatracker.ietf.org/doc/html/rfc7033#section-4.4).
 */
export interface ResourceDescriptor {
  subject?: string;
  aliases?: string[];
  properties?: Record<string, string>;
  links?: Link[];
}

/**
 * Represents a link.  See also
 * [RFC 7033 section 4.4.4](https://datatracker.ietf.org/doc/html/rfc7033#section-4.4.4).
 */
export interface Link {
  rel: string;
  type?: string;
  href?: string;
  titles?: Record<string, string>;
  properties?: Record<string, string>;
  template?: string;
}
```

### Scope / Dependencies

- `packages/cli/webfinger.ts`
- `packages/fedify/src/webfinger/*`(especially `jrd.ts`)

dodok8:
## Summary

Change ResourceDescriptor and Link interfaces to support [OStatus 1.0 Draft 2](https://www.w3.org/community/ostatus/wiki/images/9/93/OStatus_1.0_Draft_2.pdf)

## Related Issue

- closes #402 
- related #119 

## Changes

- Add `OStatusSubscribeLink`
- Change `href` field in `Link` Optional
- Change `links` field of `ResourceDescriptor` to `Array<Link | OStatusSubscribeLink>`

## Benefits

Describe the advantages or improvements brought by these changes.
Explain how these changes affect the project, users, or performance.

## Checklist

## Additional Notes
It's a breaking change about `ctx.lookupWebFinger()`.

### DIFF
--- a/docs/manual/context.md
+++ b/docs/manual/context.md
@@ -435,7 +435,7 @@ The returned WebFinger document contains links to various resources associated
 with the account, such as profile pages, ActivityPub actor URIs, and more:
 
 ~~~~ typescript twoslash
-import { type Context, isLink } from "@fedify/fedify";
+import { type Context, isLink, isOStatusSubscribeLink } from "@fedify/fedify";
 const ctx = null as unknown as Context<void>;
 // ---cut-before---
 const webfingerData = await ctx.lookupWebFinger("acct:fedify@hollo.social");
@@ -452,6 +452,16 @@ if (activityPubActorLink?.href) {
   const actor = await ctx.lookupObject(activityPubActorLink.href);
   // Work with the actor...
 }
+
+// Get subscription template URI
+const activityPubSubscribeLink = webfingerData?.links?.find((link) =>
+  isOStatusSubscribeLink(link)
+);
+
+if (activityPubSubscribeLink?.template) {
+  // Work with remote follow URI...
+}
+
 ~~~~
 
 > [!NOTE]


packages/fedify/src/webfinger/jrd.ts
```diff
@@ -57,3 +57,52 @@ export interface Link {
...
+ */
+export interface OStatusSubscribeLink {
+  rel: "http://ostatus.org/schema/1.0/subscribe";
+  /**
+   * A URI template (RFC 6570) that can be used to construct URIs by
+   * substituting variables. Used primarily for subscription endpoints
+   * where parameters like account URIs need to be dynamically inserted.
+   */
+  template: string;
+}
```
dahlia, line 73: Instead of declaring a separate `OStatusSubscribeLink` type, how about adding a `template` field to the existing `Link` type? I think the `template` field might be useful even when the `rel` field is not `"http://ostatus.org/schema/1.0/subscribe"`.

dodok8: This way seems right; however, this PR and branch are mainly about separating types, so I'll close this PR and make a new PR after adding the `template` field and fixing the following problems.

PR #410, opened 2025-09-07
Add inbox command using Optique

### LINKED ISSUE #396: CLI: Add `inbox` command using Optique
### Summary

Add `inbox` command to cli with [Optique](https://optique.dev/) instead of cliffy.  

### Problem

For now, `inbox.ts` has an empty function only. It should be implemented to run as same as original cli with cliffy does, but with Optique. Also it should be cross-runtime so existing Deno API should be replaced to Node API. 

### Proposed Solution

1. Add [`inbox.tsx`](https://github.com/fedify-dev/fedify/blob/main/packages/cli/src/inbox.tsx)
2. Add [`inbox.test.ts`](https://github.com/fedify-dev/fedify/blob/main/packages/cli/src/inbox.test.ts) 
3. Add [`inbox/`](https://github.com/fedify-dev/fedify/blob/main/packages/cli/src/inbox)
 
 
### Alternatives Considered

.

### Scope / Dependencies

- `fedify/packages/cli/src/mod.ts`
- `fedify/packages/cli/src/inbox.tsx`
- `fedify/packages/cli/src/inbox.test.ts`
- `fedify/packages/cli/src/log.ts`
- `fedify/packages/cli/src/table.ts`
- `fedify/packages/cli/src/inbox/entry.ts`
- `fedify/packages/cli/src/inbox/rendercode.ts`
- `fedify/packages/cli/src/inbox/view.tsx`

sij411:
## Summary

Add inbox command with Optique

## Related Issue

- closes #396 

## Changes

- Added inbox command and related scripts

## Benefits

Interfaces for the command option and option validations are removed.

## Checklist

- [x] Did you run `deno taskt test-all` on your machine?

## Additional Notes

Removed redundant inbox.test.ts. The tests within were for option combinations that the Optique parser now handles.


packages/cli/package.json
```diff
@@ -46,6 +46,11 @@
     "@cross/dir": "jsr:@cross/dir@^1.1.0",
     "@hongminhee/localtunnel": "^0.3.0",
     "@std/fmt/colors": "jsr:@std/fmt@^0.224.0/colors",
-    "@david/dax": "jsr:@david/dax@^0.43.2"
+    "@david/dax": "jsr:@david/dax@^0.43.2",
+    "cli-table3": "^0.6.5",
+    "hono": "jsr:@hono/hono@^4.8.3",
```
dahlia: We should avoid JSR packages in *package.json*:

```suggestion
    "hono": "^4.8.3",
```

dahlia (changes requested):

packages/cli/src/inbox.tsx
```diff
@@ -0,0 +1,536 @@
...
+  Image,
+  isActor,
+  lookupObject,
+  MemoryKvStore,
+  PUBLIC_COLLECTION,
+  type Recipient,
+} from "@fedify/fedify";
+import { getLogger } from "@logtape/logtape";
+import * as colors from "@std/fmt/colors";
+import { parse } from "@std/semver";
```
dahlia: We probably don't need *@std/semver* here; you can use `parseSemVer()` from `@fedify/fedify/nodeinfo` IIRC.

dahlia (changes requested):
In order to make it runnable on Node.js, we need to get rid of `Deno.serve()` from *packages/cli/src/tempserver.ts* and utilize [srvx](https://srvx.h3.dev/) instead.

packages/cli/src/inbox/rendercode.ts
```diff
@@ -0,0 +1,68 @@
...
+    code += `${capitalize(key)}: ${value}\n`;
+  }
+  let body: string;
+  try {
+    body = await request.text();
+  } catch (_) {
+    body = "[Failed to decode body; it may be binary.]";
+  }
+  code += `\n${body}`;
+  return code;
```
notJoon, line 20: This pattern seems to be used in `renderResponse` as well. How about reducing the duplication?

notJoon (commented):
looks good overall

packages/cli/src/inbox/rendercode.ts
sij411: Yea it's true. I have two ideas on my mind for now, 1) To separate that patterns and have 3 functions (`render(), renderRequest(), renderResponse()` 2) To combine two functions and have if conditions to process each type `Request|Response`. But I think they're quite similar to original one. It does reduce the duplications, but that doesn't sound necessary. Do you have specific idea on your mind?

dahlia: Hmm, I think we can probably hold off on refactoring the duplicate code for now and just move forward. @sij411 Could you create a separate issue for removing it? I think it would make for a good first issue.

dahlia (approved):

packages/cli/src/inbox/rendercode.ts
sij411: Got it.

PR #427, opened 2025-09-11
Change NodeInfo `software.version` field type from `SemVer` to `string`

### LINKED ISSUE #366: Change NodeInfo `software.version` field type from `SemVer` to `string`
## Summary

The NodeInfo specification does not require the `software.version` field to follow Semantic Versioning, but Fedify currently enforces this by using the `SemVer` type for this field. This causes compatibility issues with legitimate software that uses different versioning schemes (see #353 for a real-world example with snac).

## Current Situation

Currently, the `NodeInfo` interface defines `software.version` as:

```typescript
interface NodeInfo {
  software: {
    name: string;
    version: SemVer;  // <-- This is the problem
    // ...
  };
  // ...
}
```

This forces all NodeInfo implementations to use Semantic Versioning, which is not required by the [NodeInfo 2.1 specification](https://nodeinfo.diaspora.software/schema.html). The spec simply states that `software.version` is a string without any format requirements.

## Proposed Change

In Fedify 2.0, change the type of `software.version` from `SemVer` to `string`:

```typescript
interface NodeInfo {
  software: {
    name: string;
    version: string;  // <-- Plain string, as per spec
    // ...
  };
  // ...
}
```

## Breaking Change Impact

This is a breaking change that will affect:

1. **NodeInfo dispatchers** — Code that creates `NodeInfo` objects will need to change from:
   ```typescript
   version: { major: 1, minor: 0, patch: 0 }
   ```
   to:
   ```typescript
   version: "1.0.0"
   ```

2. **NodeInfo consumers** — Code that reads `NodeInfo.software.version` will receive a string instead of a `SemVer` object, affecting any code that accesses `version.major`, `version.minor`, or `version.patch`.

## Related Issues

- #353 — The original bug report that highlighted this design issue
- #365 — PR implementing the immediate fix (best-effort parsing improvement)

## References

- https://github.com/fedify-dev/fedify/issues/353#issuecomment-3157248800
- https://github.com/fedify-dev/fedify/pull/365#issuecomment-3175216071

## Alternative Considered

We considered supporting both `SemVer` and `string` via a union type (`SemVer | string`), but this would:
- Add complexity to the API
- Still not be spec-compliant (the spec says it's a string, period)
- Create confusion about which type to use when

Therefore, a clean break to `string` in a major version is the preferred approach.

dodok8:
## Summary

Change NodeInfo `software.version` field type from `SemVer` to `string`

## Related Issue

- fixes #366 

## Changes

- Changes the behavior of `parseNodeInfo`: assign `String(data.version)` if version field exists.
- Change related test.
- Change related CLI tools and codes.
- Removed `Semver` related things - `parserSemver` and `formatSemver`
- Change related document.

## Benefits

The NodeInfo specification does not require the `software.version` field to follow Semantic Versioning. Now we can handle non-SemVer version text in the `software.version` field both server-side and client-side.

## Checklist


dahlia:
@dodok8 Could you rebase your commits on the current *next* branch?

dahlia:
I believe you should drop the first two commits: 30bd3bef49da3da4bbe1f1b83a8e1d42d6d34064 and acf745ec67736c44fc873001553f8d144f177013. They apparently are included by mistake.

PR #428, opened 2025-09-11
Fix TypeScript 5.9 compatibility issues with `BufferSource` types

### LINKED ISSUE #423: Fix TypeScript 5.9 compatibility issues with `BufferSource` types
### Problem

GitHub Actions workflows are failing due to type errors after Deno automatically upgraded to version 2.5, which includes TypeScript 5.9.2. The type errors occur when passing `Uint8Array<ArrayBufferLike>` to functions expecting `BufferSource`.

### Error details

The following type errors are occurring in `packages/fedify/src/runtime/key.ts`:

```
Argument of type 'Uint8Array<ArrayBufferLike>' is not assignable to parameter of type 'BufferSource'.
  Type 'Uint8Array<ArrayBufferLike>' is not assignable to type 'ArrayBufferView<ArrayBuffer>'.
    Types of property 'buffer' are incompatible.
      Type 'ArrayBufferLike' is not assignable to type 'ArrayBuffer'.
        Type 'SharedArrayBuffer' is missing the following properties from type 'ArrayBuffer': 
        resizable, resize, detached, transfer, transferToFixedLength
```

Affected lines:
- Line 42: `crypto.subtle.importKey()` with "pkcs8" format
- Line 114: `crypto.subtle.importKey()` with "spki" format  
- Line 122: `crypto.subtle.importKey()` with "raw" format

### Root cause

- Deno 2.5 upgraded TypeScript from 5.8.3 to 5.9.2
- TypeScript 5.9 includes changes to lib.d.ts that make `ArrayBufferLike` and `ArrayBuffer` types incompatible
- The new `ArrayBuffer` type in TypeScript 5.9 includes additional properties that `SharedArrayBuffer` doesn't have

### Solution

Instead of downgrading Deno, we should fix the type compatibility issues properly. The `Uint8Array` instances need to be cast or ensured to use `ArrayBuffer` instead of `ArrayBufferLike` when passed to Web Crypto API methods.

### References

- [Deno 2.5 release notes](https://deno.com/blog/v2.5#v8-140-and-typescript-592)
- [TypeScript 5.9 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#libdts-changes)

dodok8:
## Summary

It resolves type issues with `BufferSource` types related to `ArrayBufferLike`

## Related Issue

- closes #423 

## Changes

- Change type definition of `spki`
- Now it makes a shallow copy of `Uint8Array<ArrayBufferLike>` using `slice()` to make `Uint8Array<ArrayBuffer>`
 
## Benefits

Now, `@fedify/fedify` supports Deno 2.5.0.

## Checklist

## Additional Notes

The `federation/kv.test.ts` is failing on local.

```typescript
  await t.step("set() & get()", async () => {
    await store.set(["foo", "bar"], "foobar");
    assertEquals(await store.get(["foo", "bar"]), "foobar");
    assertEquals(await store.get(["foo"]), undefined);

    await store.set(["foo", "baz"], "baz", {
      ttl: Temporal.Duration.from({ seconds: 0 }),
    });
    assertEquals(await store.get(["foo", "baz"]), undefined);
  });
```

`assertEquals(await store.get(["foo", "baz"]), undefined);` failed with value `"baz"`, which means the `ttl` options don't work properly. It's another issue anyway; it makes the test fail.

### DIFF
--- a/packages/fedify/src/runtime/key.ts
+++ b/packages/fedify/src/runtime/key.ts
@@ -27,14 +27,13 @@ const algorithms: Record<
  */
 export async function importSpki(pem: string): Promise<CryptoKey> {
   pem = pem.replace(/(?:-----(?:BEGIN|END) PUBLIC KEY-----|\s)/g, "");
-  let spki: Uint8Array;
+  let spki: Uint8Array<ArrayBuffer>;
   try {
     spki = decodeBase64(pem);
   } catch (_) {
     throw new TypeError("Invalid PEM-SPKI format.");
   }
-  const spkiArrayBuffer = spki.slice().buffer;
-  const pki = PublicKeyInfo.fromBER(spkiArrayBuffer);
+  const pki = PublicKeyInfo.fromBER(spki);
   const oid = pki.algorithm.algorithmId;
   const algorithm = algorithms[oid];
   if (algorithm == null) {
@@ -42,7 +41,7 @@ export async function importSpki(pem: string): Promise<CryptoKey> {
   }
   return await crypto.subtle.importKey(
     "spki",
-    spkiArrayBuffer,
+    spki,
     algorithm,
     true,
     ["verify"],


CHANGES.md
```diff
@@ -61,6 +61,9 @@ To be released.
  -  Added `Federatable.setWebFingerLinksDispatcher()` method to set additional
     links to WebFinger.  [[#119], [#407] by HyeonseoKim]
 
+ -  Fixed TypeScript 5.9 compatibility issues with `BufferSource` types in
+    `runtime/key.ts` operations.  [[#423], [#428] by Hyeonseo Kim]
+
```
dahlia: Since this change doesn't affect users (it's inside API surface), you don't have to (and shouldn't) write a changelog about it.

PR #431, opened 2025-09-13
CommonJS support for npm packages

### LINKED ISSUE #429: CommonJS support for better NestJS integration
While Fedify claims to officially support NestJS, the reality is that using them together requires enabling Node.js's `--experimental-require-module` flag, and even then, developers face several runtime issues that make the integration brittle.

The core problem stems from Fedify being an ESM-only package while NestJS primarily operates in the CommonJS ecosystem. This mismatch creates what's known as the [“dual package hazard”](https://nodejs.org/en/learn/modules/publishing-a-package#the-dual-package-hazard). For instance, when both Fedify and a NestJS application depend on `@js-temporal/polyfill`, they end up loading different versions of the module—Fedify loads the ESM version while NestJS loads the CJS version. This means that when you create a `Temporal.Instant` instance in your NestJS code and pass it to Fedify, the `instanceof Temporal.Instant` checks inside Fedify fail, even though you're technically passing the correct type. This is just one example of the subtle bugs that plague the current integration.

These issues significantly degrade the developer experience, especially in enterprise environments where NestJS is commonly used. Such environments typically avoid experimental Node.js features and prioritize stability. If we want to claim first-class NestJS support, we need to meet developers where they are, not force them into using experimental features and dealing with mysterious runtime failures.

Since we're already using tsdown for our build process, adding CJS support should be relatively straightforward. We could provide dual packages (both ESM and CJS) from the same TypeScript source, allowing NestJS users to import Fedify without any special flags or workarounds. This approach to [publishing both ESM and CJS](https://antfu.me/posts/publish-esm-and-cjs) would eliminate the module resolution issues and make Fedify truly compatible with the broader Node.js ecosystem, not just the ESM-forward portion of it.

I believe this change would make Fedify much more accessible to the NestJS community and remove a significant barrier to adoption in enterprise settings.

dahlia:
### Summary

- All npm-published packages now support both [ESM] and [CommonJS] module formats
- Resolves dual package hazard issues with NestJS applications
- Eliminates the need for Node.js's [`--experimental-require-module`] flag

### Changes

- Configure tsdown to build both ESM and CJS formats for npm packages
- Update package.json exports for dual package support
- Add proper type definitions for both module systems  
- Update documentation to recommend ESM while noting CommonJS support

### Test plan

- [x] Verify CJS imports work correctly in Node.js applications
- [ ] Test ESM imports continue to work as expected
- [x] Confirm NestJS integration works without experimental flags

Resolves #429

[ESM]: https://nodejs.org/api/esm.html
[CommonJS]: https://nodejs.org/docs/latest/api/modules.html
[`--experimental-require-module`]: https://nodejs.org/docs/latest/api/modules.html#loading-ecmascript-modules-using-require

### DIFF
--- a/packages/fedify/deno.json
+++ b/packages/fedify/deno.json
@@ -20,7 +20,6 @@
   },
   "imports": {
     "@cfworker/json-schema": "npm:@cfworker/json-schema@^4.1.1",
-    "@hugoalh/http-header-link": "jsr:@hugoalh/http-header-link@^1.0.2",
     "@multiformats/base-x": "npm:@multiformats/base-x@^4.0.1",
     "@opentelemetry/api": "npm:@opentelemetry/api@^1.9.0",
     "@opentelemetry/semantic-conventions": "npm:@opentelemetry/semantic-conventions@^1.27.0",
--- a/packages/fedify/package.json
+++ b/packages/fedify/package.json
@@ -151,7 +151,6 @@
   },
   "dependencies": {
     "@cfworker/json-schema": "^4.1.1",
-    "@hugoalh/http-header-link": "^1.0.2",
     "@js-temporal/polyfill": "^0.5.1",
     "@logtape/logtape": "catalog:",
     "@multiformats/base-x": "^4.0.1",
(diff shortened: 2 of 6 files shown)


ThisIsMissEm:
I'm not sure the package exports are right reading through: https://hirok.io/posts/package-json-exports

dahlia:
I understand that when using [nested conditions](https://nodejs.org/api/packages.html#nested-conditions), putting a `require`/`import` condition inside a `types` condition works the same way as putting a `types` condition inside a `require`/`import` condition.

PR #434, opened 2025-09-13
Lift up checks to notAcceptable to the middleware

ThisIsMissEm:
## Summary

Following discussion in discourse, this PR shows the triggering of `notAcceptable` happening from the middleware before any of the dispatchers are triggered, since they should only be triggered if they would actually be acceptable and lead to a response.

This means that when you're passing through requests via content-negotiation, the fedify dispatcher is not fired if it's not a `Accept` content-type that fedify would actually handle.

This would be a major breaking change, hence targeting `next`

## Related Issue

There doesn't seem to be related issues reported, however, it's a common point of confusion for integraters "why is my fedify dispatcher firing for this html request?"

## Changes

- Lifted check for Accept header content-type up to middleware
- Moved `acceptsJsonLd` to `negotiation.ts` (from `handle.ts`)
- Removed old notAcceptable tests from `handler.test.ts` — I'm still not 100% sure how these tests were working.

## Benefits

Describe the advantages or improvements brought by these changes.
Explain how these changes affect the project, users, or performance.

## Checklist

## Additional Notes


ThisIsMissEm:
This actually seems to mess with a tonne of the tests, because most of them don't set the `Accept` header, so I'm not sure how to proceed there.

packages/fedify/src/federation/middleware.test.ts
```diff
@@ -963,6 +974,222 @@ test({
...
+          "accept": "application/ld+json",
+          "content-type": "application/ld+json",
+        },
+        body: JSON.stringify(createFixture),
+      }),
+      rsaPrivateKey2,
+      rsaPublicKey2.id!,
+    );
+
+    let response = await federation.fetch(
```
ThisIsMissEm: These requests are all returning 401 and I'm not sure why:
```
 packages/fedify test:bun: (pass) Federation.createContext() [178.00ms]
packages/fedify test:bun: {
packages/fedify test:bun:   response: Response (39 bytes) {
packages/fedify test:bun:     ok: false,
packages/fedify test:bun:     url: "",
packages/fedify test:bun:     status: 401,
packages/fedify test:bun:     statusText: "",
packages/fedify test:bun:     headers: Headers {
packages/fedify test:bun:       "content-type": "text/plain; charset=utf-8",
packages/fedify test:bun:       "vary": "Accept",
packages/fedify test:bun:     },
packages/fedify test:bun:     redirected: false,
packages/fedify test:bun:     bodyUsed: false,
packages/fedify test:bun:     Blob (39 bytes)
packages/fedify test:bun:   },
packages/fedify test:bun:   inbox: [],
packages/fedify test:bun: }
```

I can't seem to find what exactly is triggering the 401s.

packages/fedify/src/federation/middleware.test.ts
```diff
@@ -35,18 +38,7 @@ import {
...
-  Create,
-  type CryptographicKey,
-  Invite,
-  Multikey,
-  Note,
-  Object,
-  Offer,
-  Person,
-} from "../vocab/vocab.ts";
+import * as vocab from "../vocab/vocab.ts";
```
ThisIsMissEm, line 41: If you're wondering why I had to make this change, I needed to use `Object.fromEntries` whilst debugging.

dahlia (approved):
Good job! Thanks for your contribution.

PR #442, opened 2025-09-21
CI/CD: Cancel concurrent workflow runs

ThisIsMissEm:
## Summary

This prevents multiple pushes from blocking up the workflow runner queue

## Related Issue

No related issue to my knowledge, observed during #434 

## Changes

- Uses the workflow concurrency controls to cancel prior workflow runs automatically.

## Benefits

This improves the performance of the workflow runners by reducing their workload when people push multiple commits in quick succession to a branch.

## Checklist

## Additional Notes

Docs: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,6 +1,10 @@
 name: build
 on: [push, pull_request_target]
 
+concurrency:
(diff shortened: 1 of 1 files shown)


dahlia (approved):
Thanks!

PR #446, opened 2025-09-26
Add TypeScript support for RFC 6570 URI Template expressions (e.g., {+identifier})

### LINKED ISSUE #426: Add TypeScript support for RFC 6570 URI Template expressions (e.g., `{+identifier}`)
### Summary

Currently, Fedify's TypeScript types only allow simple string expansion (`{identifier}`) for URI template parameters, but the runtime already supports all RFC 6570 expression types. This causes type errors when using other valid RFC 6570 expressions like `{+identifier}` (reserved string expansion), which is the recommended solution for URI-containing identifiers.

### Background

This issue was discovered in #416 where a user needed to use `{+identifier}` to properly handle URI identifiers with special characters. While the runtime works correctly, TypeScript throws errors because the current type definitions are too restrictive.

**Current type constraint:**
```typescript
type ParamPath<Param extends string> = `${string}{${Param}}${string}`;
```

**Runtime support:** Already supports all RFC 6570 expressions via `uri-template-router` and `url-template` libraries.

### Expected behavior

TypeScript should accept all valid RFC 6570 expression types:
- `{identifier}` - Simple string expansion
- `{+identifier}` - Reserved string expansion (recommended for URI identifiers)
- `{#identifier}` - Fragment expansion
- `{.identifier}` - Label expansion with dot-prefix
- `{/identifier}` - Path segments
- `{;identifier}` - Path-style parameters
- `{?identifier}` - Query component
- `{&identifier}` - Query continuation

### Current behavior

Using any expression other than simple string expansion causes TypeScript errors:

```typescript
// ❌ Type error
federation.setFollowersDispatcher(
  `/users/{+identifier}/followers`, // Type error here
  async (ctx, identifier, cursor) => { /* ... */ },
)
```

### Proposed solution

1. **Create RFC 6570 expression type helper:**
   ```typescript
   type Rfc6570Expression<Param extends string> =
     | `{${Param}}`     // Simple string expansion
     | `{+${Param}}`    // Reserved string expansion
     | `{#${Param}}`    // Fragment expansion
     | `{.${Param}}`    // Label expansion with dot-prefix
     | `{/${Param}}`    // Path segments
     | `{;${Param}}`    // Path-style parameters
     | `{?${Param}}`    // Query component
     | `{&${Param}}`;   // Query continuation
   ```

2. **Update ParamPath type:**
   ```typescript
   type ParamPath<Param extends string> = `${string}${Rfc6570Expression<Param>}${string}`;
   ```

3. **Update documentation** to clarify when to use different expression types (especially `{+identifier}` for URI identifiers).

### Implementation details

**Files to modify:**
- `packages/fedify/src/federation/federation.ts` (lines ~1180) - Update type definitions
- Documentation to include RFC 6570 expression usage examples

**Testing:**
- Verify all existing simple expansion usage continues to work
- Add test cases for different RFC 6570 expression types
- Ensure runtime behavior remains unchanged (type-only change)

### Impact

- **Breaking change:** No - this is additive type improvement
- **Runtime impact:** None - runtime already supports these expressions
- **Documentation:** Update needed to explain when to use different expressions

### Related issues

- #416 - Original issue that revealed this limitation
- Related to upstream `uri-template-router` issue: https://github.com/awwright/uri-template-router/issues/11

sij411:
## Summary

Added TypeScript support for RFC 6570 URI Template expressions (e.g., {+identifier})

## Related Issue

- closes #426 

## Changes

- Removed `ParamPath` related codes since `tsc` can't handle too many type combinations 
- Added RFC 6570 on `Param` type

## Benefits

Performance & Type Safety: Reduced TypeScript compilation time by
optimizing parameter support from 6 to 3, while maintaining type
safety and removing 143 lines of complex code.

Enhanced Collection Handling: Added new setCollectionDispatchers and
setOrderedCollectionsDispatchers methods supporting up to 5
parameters for improved federated application flexibility.

Standards Compliance: Implemented RFC 6570 URI template support and
updated handler documentation for better interoperability and
developer experience.

## Checklist

## Additional Notes


packages/fedify/src/federation/federation.ts
```diff
@@ -454,9 +473,125 @@ export interface Federatable<TContextData> {
...
+  //     TParam,
+  //     RequestContext<TContextData>,
+  //     TContextData
+  //   >,
+  // ): CustomCollectionCallbackSetters<
+  //   TParam,
+  //   RequestContext<TContextData>,
+  //   TContextData
+  // >;
+
```
2chanhaeng: This code looks unnecessary.

2chanhaeng (commented):
LGTM!
I suggest to replace `(new (...args: any[]) => TObject) & { typeId: URL }` with `ConstructorWithTypeId` that declare in `federation.ts`.
And you don't have to declare every interface by the number of parameters if you declare following type(you can name it more proper):
```ts
type ParamPath<TParam extends string> =
  | `${string}${
      Rfc6570Expression<TParam>
    }${string}${
      Rfc6570Expression<TParam>
    }${string}${
      Rfc6570Expression<TParam>
    }${string}${
      Rfc6570Expression<TParam>
    }${string}`
  | `${string}${
      Rfc6570Expression<TParam>
    }${string}${
      Rfc6570Expression<TParam>
    }${string}${
      Rfc6570Expression<TParam>
    }${string}`
  | `${string}${
      Rfc6570Expression<TParam>
    }${string}${
      Rfc6570Expression<TParam>
    }${string}`
  | `${string}${Rfc6570Expression<TParam>}${string}`;
```

packages/fedify/src/federation/federation.ts
sij411: The main reason I didn't use `ConstructorWithTypeId` in many dispatchers is to maintain the consistency in codebase. Previously, it was used in only custom collection dispatchers not others. I'll take a look and see what's the best design for these. Thanks.

2chanhaeng: The reason `ConstructorWithTypeId` isn't widely used is because I recently created it while implementing `CustomCollection`. @daliah, could you take a look?

dahlia: @2chanhaeng Could you file an issue to define and apply the `ConstructorWithTypeId` type where it's applicable? I think doing this in the current pull request is a bit out of scope.

2chanhaeng: Done! Check #447 please!

dahlia (approved):

Issue #359, opened 2025-08-07
The `@fedify/relay` package for ActivityPub relays


dahlia:
### Summary

Create a new package `@fedify/relay` that provides a comprehensive implementation of ActivityPub relay protocols based on the [FEP-ae0c](https://w3id.org/fep/ae0c) specification. This package should enable developers to easily create and operate relay servers that facilitate content distribution across the fediverse, providing separate implementations for Mastodon-style and LitePub-style relay protocols through a common interface.

### Problem

ActivityPub relays are crucial infrastructure components that help small instances participate effectively in the federated social network by acting as intermediary servers that distribute public content without requiring individual actor-following relationships. Currently, there is no standardized, easy-to-use implementation available for developers who want to create relay servers using modern JavaScript/TypeScript tooling.

The existing relay implementations in the ecosystem are often written in different languages (Rust, Go, Python) or are tightly coupled to specific platforms. This creates barriers for developers who want to build relay functionality into their Fedify-based applications or create standalone relay servers with the same developer experience that Fedify provides for ActivityPub servers.

[FEP-ae0c](https://w3id.org/fep/ae0c) documents two popular relay protocols that are widely deployed in the fediverse: Mastodon-style relays (which use LD signatures and follow the Public collection) and LitePub-style relays (which use bidirectional following relationships). These protocols have fundamental differences in their subscription mechanisms, activity publishing methods, and signature requirements, making it impractical to support both protocols simultaneously in a single relay instance. A comprehensive relay package should provide separate implementations for each protocol while maintaining a unified developer interface.

### Proposed Solution

The `@fedify/relay` package should provide a common `Relay` interface implemented by protocol-specific classes: `MastodonRelay` and `LitePubRelay`. Both classes should follow the same architectural patterns as Fedify's `Federation` object, each offering a `fetch()` method that can be integrated with any web framework or runtime. The package should internally leverage Fedify's existing infrastructure for ActivityPub message handling, HTTP signatures, WebFinger support, and key–value storage.

The common interface should enable developers to work with either relay type through consistent APIs while allowing each implementation to handle protocol-specific requirements appropriately. Here's an example of how the API might look:

```typescript
import { MastodonRelay, LitePubRelay, type Relay } from "@fedify/relay";
import { MemoryKvStore } from "@fedify/fedify";

// Common interface for both relay types
interface Relay {
  fetch(request: Request, options?: { contextData?: any }): Promise<Response>;
[...]

PR #471, opened 2025-10-19
Add LitePub relay implementation

sij411:
## Summary

  This PR adds a LitePub-compatible ActivityPub relay implementation to the
  `@fedify/relay` package, complementing the existing Mastodon relay implementation.

## Related Issue

#359 
 
## Changes

  - Added LitePubRelay class in packages/relay/src/relay.ts:321-585 that implements
  the LitePub relay protocol
  - Implemented LitePub-specific activity handling:
    - Wraps incoming Create, Update, Move, and Delete activities in Announce
  activities before forwarding
    - Uses Application actor type instead of Mastodon's Service type
    - Sends announcements to PUBLIC_COLLECTION for broader visibility
  - Added comprehensive test suite for LitePubRelay in
  packages/relay/src/relay.test.ts:455-832 covering:
    - Relay initialization and configuration
    - Actor endpoint behavior (Application type)
    - Followers collection management
    - Subscription handler integration
    - Follow/Undo activity processing
    - Multiple follower support

 ## Benefits

  - Protocol compatibility: Enables Fedify relay instances to work with
  LitePub-compatible fediverse software, which expects activities to be wrapped in
  Announce rather than directly forwarded
  - Broader reach: Using PUBLIC_COLLECTION as the audience for announces ensures
  better visibility across the fediverse
  - Developer choice: Developers can now choose between Mastodon-style
  (MastodonRelay) and LitePub-style (LitePubRelay) relay implementations based on
  their target ecosystem
  - Consistent API: Both relay implementations share the same Relay interface, making
   it easy to switch between protocols

 ## Checklist

  - [ ] Did you add a changelog entry to the CHANGES.md?
  - [x] Did you run deno task test-all on your machine?

  ## Additional Notes

  Key differences between LitePub and Mastodon relay protocols:

  1. Actor type: LitePub uses Application, Mastodon uses Service
  2. Activity forwarding:
    - LitePub wraps activities in Announce objects with relay as actor
    - Mastodon forwards activities directly using ctx.forwardActivity()
  3. Audience: LitePub sends to PUBLIC_COLLECTION for maximum visibility

  The implementation follows the same subscription management and key pair handling
  as MastodonRelay, ensuring consistent behavior for core relay functionality like
  follower management and subscription approvals.

### DIFF
--- a/packages/relay/src/relay.ts
+++ b/packages/relay/src/relay.ts
@@ -62,10 +62,11 @@ export interface Relay {
   setSubscriptionHandler(handler: SubscriptionRequestHandler): this;
 }
 
-type LitePubRelayFollower = {
-  actor: unknown;
-  state: string;
-};
+interface LitePubRelayFollower {
+  readonly actor: unknown;
+  readonly state: string;
+}
+
 /**
  * A Mastodon-compatible ActivityPub relay implementation.
  * This relay follows Mastodon's relay protocol for maximum compatibility


dahlia:
@sij411 Could you rebase your commits on the latest *next* branch?

packages/relay/src/relay.ts
```diff
@@ -59,6 +62,10 @@ export interface Relay {
   setSubscriptionHandler(handler: SubscriptionRequestHandler): this;
 }
 
+type LitePubRelayFollower = {
+  actor: string;
+  state: string;
+};
```
dahlia: I think the `actor` field should be `unknown` so that it has a JSON object?

```suggestion
interface LitePubRelayFollower {
  readonly actor: unknown;
  readonly state: string;
};
```

sij411: Oops got it

sij411:
> By the way, `LitePubRelay` should be tested with actual Pleroma/Akkoma instances.

I'll let you know test result when it's done. maybe in few days.

packages/relay/src/relay.ts
dahlia: I think you should address this?

packages/relay/src/relay.ts
```diff
@@ -322,9 +329,341 @@ export class LitePubRelay implements Relay {
...
+
+    this.#federation.setFollowingDispatcher(
+      "/users/{identifier}/following",
+      (_ctx, _identifier) => {
+        return {
+          items: [],
+        };
+      },
+    )
+      .setCounter((_ctx, _identifier) => 0);
```
dahlia: Since LitePub Relay protocol requires both ends follow each other, this following dispatcher should return the same collection as the followers dispatcher.

sij411: Does following dispatcher need Counter?

dahlia: I don't think so.

packages/relay/src/relay.ts
sij411: Resolved with commit `020b6605`

PR #479, opened 2025-11-06
Add `deno task test-init` for testing `fedify init` and fixed `fedify init` CLI with it

### LINKED ISSUE #461: Add tests for `init` CLI  and fix the command
### Summary

We need tests for `init` CLI and fix the command

### Problem

When the version was updated to 2.0, the `init` CLI not only applied `optique` but also underwent code refactoring. However, this process introduced a problem in the code, preventing it from running. Since there are too many cases, it's inefficient for a human to check each one individually, so a test code is needed for this. This test code would consume too many resources to run during the deployment process. Therefore, tests are added assuming only the scenario where the user runs them directly on their local machine.

### Proposed Solution

- Add test for `init` CLI 
- Fix `init` CLI

### Alternatives Considered

_No response_

### Scope / Dependencies

- `packages/cli/src/init`

2chanhaeng:
## Summary

Add `deno task test-init` for testing `fedify init` and fixed `fedify init` CLI with it

## Related Issue

- closes #461

## Changes

- `packages/src/cli/init`
- `docs/cli.md`

## Benefits

The `test-init` task makes it easy that contributors who want to extend the `fedify init` CLI to test their changes.

## Checklist

- [ ] ~~Did you write some tests for this change (if it's a new feature)?~~

## Additional Notes

`fedify test-init` CLI was developed solely for use in local environments. CI/CD environments were not taken into consideration.

`fedify test-init` CLI can't test `bun`, `yarn`, `Next.js` for now.
The


docs/cli.md
```diff
@@ -225,6 +224,73 @@ This option works with all other initialization options, allowing you to preview
...
+------------------------------------------------------
+
+*This command is available since Fedify 2.0.0.*
+
+> [!WARNING]
+> `bun`, `yarn` runtime and `Next.js` web framework are not supported in this 
+> command yet.
+
+> [!NOTE]
+> You need to install and run Redis server, PostgreSQL server, or RabbitMQ
```
sij411: I'd recommend adding information on how to prepare beforehand, including Redis and PostgreSQL installation methods depending on different system environments.

packages/cli/src/init/test/lookup.ts
```diff
@@ -0,0 +1,253 @@
...
+    sendLookup,
+  );
+
+  printMessage`    Lookup ${result ? "successful" : "failed"} for ${
+    values([wf, pm, kv, mq])
+  }!`;
+  if (!result) {
+    printMessage`    Check out these files for more details:
+      ${join(dir, "out.txt")}
+      ${join(dir, "err.txt")}\n`;
```
dodok8, line 92:
```suggestion
    printMessage`    Check out these files for more details:
      ${join(dir, "out.txt")} and 
      ${join(dir, "err.txt")}\n`;
```

It's difficult to discern that error messages are being output to two different files. It would be better to include a word like `and` to explicitly distinguish them.

packages/cli/src/init/test/lookup.ts
```diff
@@ -0,0 +1,253 @@
...
+    // Common patterns for port detection
+    const portPatterns = [
+      /listening on.*:(\d+)/i,
+      /server.*:(\d+)/i,
+      /port\s*:?\s*(\d+)/i,
+      /localhost:(\d+)/i,
+      /0\.0\.0\.0:(\d+)/i,
+      /127\.0\.0\.1:(\d+)/i,
+      /https?:\/\/[^:]+:(\d+)/i,
+    ];
```
dodok8, line 212: These pattern heuristics catches redis port first. For example, when I run `test-init -p deno -w hono -k redis -m amqp -h`, it has output about `.env` file like below:

```
Package manager: "deno";
  Web framework: "hono";
  Key–value store: "redis";
  Message queue: "amqp";

Note that you probably want to edit the ".env" file.
It currently contains the following values:

  AMQP_URL='amqp://localhost'
  REDIS_URL='redis://localhost:6379'

To start the server, run the following command:

  `deno task dev`

Then, try look up an actor from your server:

  `fedify lookup http://localhost:8000/users/john`
    
Start by editing the src/federation.ts file to define your federation!
Server started at http://0.0.0.0:8000
```

And line 206(`/localhost:(\d+)/i`) catches localhost:6379 first, So it try to lookup 6379 instead of 8000.

docs/cli.md
2chanhaeng: Changed at f6360d880c7e73e3dd283351a9c1266320fc3ec9!

packages/cli/src/init/test/lookup.ts
2chanhaeng: Fixed in 5ac1b5d2eecccd9b039e3279f18c3ab4603e5b4d!

packages/cli/src/init/test/lookup.ts
2chanhaeng: Fixed in 5ac1b5d2eecccd9b039e3279f18c3ab4603e5b4d!

dahlia:
Could you resolve conflicts?

docs/cli.md
```diff
@@ -225,6 +224,81 @@ This option works with all other initialization options, allowing you to preview
...
+This will save outputs and errors to text files instead of creating a 
+temporary project.
+
+~~~~ sh
+fedify test-init -d
+~~~~
+
+If you did not specify `-h` or `-d` option, it will run both options by 
+default.
+
```
dahlia: This section documents `fedify test-init` as a public CLI command, but it has been moved to a Deno task (`deno task test-init`) and is no longer exposed in the public CLI.

Since this is now a contributor-only tool, this entire section should be removed from the public CLI documentation. The usage is already properly documented in *CONTRIBUTING.md*.

CHANGES.md
```diff
@@ -126,6 +126,12 @@ To be released.
     `-t`/`--traverse` option, allowing users to traverse multiple collections
     in a single command.  [[#408], [#449] by Jiwon Kwon]
 
+ -  Added `test-init` task to test `fedify init` command in `/packages/cli`.  
+    This task create test server with various combinations of options and test
+    the server using `fedify lookup` command.  This command is for
+    contributors who make integration with various frameworks.  [[#461] by
+    Chanhaeng Lee]
+
```
dahlia: This entry can be removed. The `test-init` task is a contributor-only development tool, not a user-facing feature, so it doesn't belong in the public changelog.

docs/cli.md
2chanhaeng: Fixed at a931db1f02f6ee51ca1af181057c60d40688a294!

CHANGES.md
2chanhaeng: Fixed at a931db1f02f6ee51ca1af181057c60d40688a294!

PR #486, opened 2025-12-01
Fix flaky WebFinger test by mocking fetch (Issue #480)

### LINKED ISSUE #480: Flaky test: `lookupSingleWebFinger` depends on external server (hollo.social)
The test for `lookupSingleWebFinger` in [`packages/cli/src/webfinger/mod.test.ts`](https://github.com/fedify-dev/fedify/blob/af353cc21820f8d8c983e91f2498d9ea5fd9870d/packages/cli/src/webfinger/mod.test.ts#L73-L79) is flaky because it depends on the external remote server `hollo.social`. When this server is temporarily unavailable, the test fails with:

```
Error [NotFoundError]: Resource not found: @fedify@hollo.social
    at lookupSingleWebFinger (/home/runner/work/fedify/fedify/packages/cli/src/webfinger/action.ts:28:5)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  resource: '@fedify@hollo.social'
}
```

### Current implementation

The test ([lines 73-79](https://github.com/fedify-dev/fedify/blob/af353cc21820f8d8c983e91f2498d9ea5fd9870d/packages/cli/src/webfinger/mod.test.ts#L73-L79)) makes real HTTP requests to external servers:

```typescript
test("Test lookupSingleWebFinger", async () => {
  const aliases = (await Array.fromAsync(
    RESOURCES,
    (resource) => lookupSingleWebFinger({ resource }),
  )).map((w) => w?.aliases?.[0]);
  assert.deepEqual(aliases, ALIASES);
});
```

Where `RESOURCES` includes ([lines 9-12](https://github.com/fedify-dev/fedify/blob/af353cc21820f8d8c983e91f2498d9ea5fd9870d/packages/cli/src/webfinger/mod.test.ts#L9-L12)):
- `@hongminhee@hackers.pub`
- `@fedify@hollo.social`

### Problem

This approach has several issues:

1. **Unreliability**: External servers may be temporarily unavailable, causing intermittent test failures
2. **Network dependency**: Tests require internet connectivity and DNS resolution
3. **Slowness**: Real HTTP requests significantly slow down the test suite
4. **Brittleness**: Changes to the external server's configuration or data can break tests unexpectedly

### Proposed solution

The test should be refactored to use one of the following approaches:

1. **Mock server**: Spin up a local HTTP server that responds to WebFinger queries with known test data
2. **Mock HTTP client**: Mock the underlying HTTP fetch calls to return pre-determined responses
3. **Use @fedify/testing utilities**: Leverage existing testing utilities from the `@fedify/testing` package if available

The ideal solution would be to create a minimal local ActivityPub/WebFinger server for testing purposes, ensuring tests are:
- Fast (no real network I/O)
- Reliable (no external dependencies)
- Deterministic (known test data)

### Additional context

- This issue only affects the `next` branch (Fedify 2.0)
- The test is located at [`packages/cli/src/webfinger/mod.test.ts:73`](https://github.com/fedify-dev/fedify/blob/af353cc21820f8d8c983e91f2498d9ea5fd9870d/packages/cli/src/webfinger/mod.test.ts#L73-L79)
- The implementation being tested is in [`packages/cli/src/webfinger/action.ts`](https://github.com/fedify-dev/fedify/blob/af353cc21820f8d8c983e91f2498d9ea5fd9870d/packages/cli/src/webfinger/action.ts)

ShilpaG813:
Summary

This PR removes the external network calls in the WebFinger test and replaces them with a mock fetch.
The original test depended on hackers.pub and hollo.social, which caused flaky failures.
Now the test is stable, offline, and fully deterministic.

Related Issue

closes #480

Changes

Added a local globalThis.fetch mock inside the WebFinger test.

Returned fixed WebFinger JRD responses from the mock.

Removed the old network-dependent test logic.

Restored the original fetch after the test.

Benefits

Test no longer fails due to external server downtime.

Test runs faster and reliably in CI.

No internet connection is required for running the test.

Checklist

 Added changelog entry (not required for test-only changes)

 Wrote docs (not needed — test fix)

 Added/updated tests

 Ran deno task test-all (CI will run tests)

Additional Notes

This change affects only the test file and does not modify production code.

### DIFF
--- a/packages/cli/src/webfinger/mod.test.ts
+++ b/packages/cli/src/webfinger/mod.test.ts
@@ -10,7 +10,6 @@ const RESOURCES = [
   "@hongminhee@hackers.pub",
   "@fedify@hollo.social",
 ];
-
 const ALIASES = [
   "https://hackers.pub/ap/actors/019382d3-63d7-7cf7-86e8-91e2551c306c",
   "https://hollo.social/@fedify",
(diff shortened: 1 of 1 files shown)


dahlia (approved):
Thanks for your contribution!

PR #490, opened 2025-12-12
refactor(relay): implement factory pattern with Mastodon and LitePub relay support

sij411:
## Summary

This PR refactors the `@fedify/relay` package to introduce a factory pattern that supports multiple relay implementations, specifically **Mastodon** and **LitePub** relay protocols. The changes improve code organization, maintainability, and extensibility for future relay implementations.

### Key Changes

- **Factory Pattern**: Introduced `createRelay()` factory function that instantiates relay implementations based on type (`"mastodon"` or `"litepub"`)
- **Separate Relay Implementations**:
  - `MastodonRelay`: Implements Mastodon's relay protocol with immediate follower acceptance
  - `LitePubRelay`: Implements LitePub's relay protocol with reciprocal follow pattern (pending → accepted states)
- **Refactored Core Module**: Extracted common relay infrastructure into `relay.ts` with shared:
  - Actor dispatcher for relay server actor
  - Key pair management (RSA + Ed25519)
  - Followers/Following collection dispatchers
- **Comprehensive Test Coverage**: Added dedicated test suites for both implementations:
  - `mastodon.test.ts`: 896 lines of tests covering Mastodon relay behavior
  - `litepub.test.ts`: 896 lines of tests covering LitePub relay behavior with reciprocal follows
  - Removed old monolithic `relay.test.ts`

### Protocol Differences

**Mastodon Relay**:
- Immediately accepts Follow activities and adds followers to the active list
- Simpler subscription flow without reciprocal follow requirement
- Suitable for permissive relay scenarios

**LitePub Relay**:
- Implements reciprocal follow pattern (relay follows back the follower)
- Two-state follower lifecycle: `pending` → `accepted`
- Follower is only activated after accepting the relay's Follow activity
- More secure and controlled subscription process

### Architecture Improvements

- **Modular Design**: Each relay type is self-contained with its own file and tests
- **Extensibility**: Easy to add new relay implementations (e.g., Akkoma, Pleroma variants)
- **Shared Infrastructure**: Common functionality centralized in `relay.ts` and `relayBuilder`
- **Type Safety**: Strong TypeScript typing with proper interfaces (`RelayOptions`, `LitePubFollower`)

### Migration Guide

**Before**:
```typescript
import { MastodonRelay, LitePubRelay } from "@fedify/relay";
const mastodonRelay = new MastodonRelay(options);
const litepubRelay = new LitePubRelay(options);
```

**After**:
```typescript
import { createRelay } from "@fedify/relay";

// For Mastodon-compatible relay
const mastodonRelay = createRelay("mastodon", options);

// For LitePub-compatible relay
const litepubRelay = createRelay("litepub", options);
```

### Files Changed

- `packages/relay/src/relay.ts`: Refactored to shared infrastructure (771 → 161 lines)
- `packages/relay/src/mastodon.ts`: New Mastodon relay implementation (191 lines)
- `packages/relay/src/litepub.ts`: New LitePub relay implementation (297 lines)
- `packages/relay/src/mastodon.test.ts`: New comprehensive test suite (784 lines)
- `packages/relay/src/litepub.test.ts`: New comprehensive test suite (896 lines)
- `packages/relay/src/relay.test.ts`: Removed old monolithic tests (1128 lines deleted)
- `packages/relay/src/mod.ts`: Updated exports to include new relay classes

### Test Plan

- [x] All Mastodon relay tests pass (784 lines of test coverage)
- [x] All LitePub relay tests pass (896 lines of test coverage)
- [x] Factory function correctly instantiates both relay types
- [x] Follower management works correctly for both protocols
- [x] Activity forwarding (Create, Update, Delete, Move, Announce) tested for both
- [x] Subscription approval/rejection flows tested
- [x] Reciprocal follow pattern tested for LitePub
- [x] HTTP signature verification works across both implementations

### Breaking Changes

This is a **breaking change** that requires updates to existing relay usage code:
- The `Relay` class constructor is replaced with `createRelay()` factory function
- Users must specify relay type: `"mastodon"` or `"litepub"`
- Default behavior changes based on selected relay type

### Related Commits

- 36b65745: Add relay factory pattern with federation builder
- 03ea8aab: Add create factory function for mastodon and litepub
- 1e7c16f2: Add mastodon relay and its test
- 0e174116: Add litepub relay and its test
- 7902d022: Delete old relay test

### DIFF
--- a/packages/relay/src/mod.ts
+++ b/packages/relay/src/mod.ts
@@ -7,7 +7,6 @@
  *
  * @module
  */
-export { BaseRelay } from "./base.ts";
 export { relayBuilder } from "./builder.ts";
 export { createRelay } from "./factory.ts";
 export { LitePubRelay } from "./litepub.ts";


packages/relay/src/litepub.ts
```diff
@@ -0,0 +1,298 @@
...
+          }
+        })
+        .on(Create, async (ctx, create) => {
+          const sender = await create.getActor(ctx);
+          const excludeBaseUris = sender?.id ? [new URL(sender.id)] : [];
+
+          const announce = new Announce({
+            id: new URL(`/announce#${crypto.randomUUID()}`, ctx.origin),
+            actor: ctx.getActorUri(RELAY_SERVER_ACTOR),
+            object: create.objectId,
```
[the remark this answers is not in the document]
sij411: The spec explicitly says:
  1. Announce the object (like a Note), not the activity (like Create)
  2. Use a URI reference (activity.objectId) rather than embedding the full object
  
  Close this conversation

packages/relay/src/relay.ts
```diff
@@ -48,648 +45,118 @@ export interface RelayOptions {
...
-  readonly domain: string;
-
-  fetch(request: Request): Promise<Response>;
-  setSubscriptionHandler(handler: SubscriptionRequestHandler): this;
-}
-
-interface LitePubRelayFollower {
+export interface RelayFollower {
   readonly actor: unknown;
   readonly state: string;
```
dahlia: The `state` field is typed as string, but the codebase only uses `"pending"` and `"accepted"` as values. Consider using a string literal union type for better type safety:

```suggestion
  readonly state: "pending" | "accepted";
```

This would catch typos at compile time and make the valid states self-documenting.

packages/relay/src/mod.ts
```diff
@@ -9,5 +9,13 @@
...
 // Export relay functionality here
-export type { Relay, RelayOptions } from "./relay.ts";
-export { LitePubRelay, MastodonRelay } from "./relay.ts";
+export {
+  createRelay,
+  RELAY_SERVER_ACTOR,
+  type RelayFollower,
+  type RelayOptions,
+  type SubscriptionRequestHandler,
+} from "./relay.ts";
```
dahlia: `Relay` interface and `RelayType` are defined in *relay.ts* but not re-exported from *mod.ts*. Users who want to type-annotate relay instances or accept relay types as parameters would benefit from having these exported:


```suggestion
export {
  createRelay,
  RELAY_SERVER_ACTOR,
  type Relay,
  type RelayFollower,
  type RelayOptions,
  type RelayType,
  type SubscriptionRequestHandler,
} from "./relay.ts";
```

This enables usage like:

```ts
import { type Relay, type RelayType } from "@fedify/relay";

async function startRelay(type: RelayType, options: RelayOptions): Promise<void> {
  const relay: Relay = createRelay(type, options);
  // … omitted …
}
```

packages/relay/package.json
dahlia: I don't think we use this line for `@fedify/relay`?

packages/relay/src/litepub.ts
```diff
@@ -0,0 +1,302 @@
...
+            const updatedFollowers = followers.filter((id) =>
+              id !== activity.actorId?.href
+            );
+            await ctx.data.kv.set(["followers"], updatedFollowers);
+            await ctx.data.kv.delete(["follower", activity.actorId?.href]);
+          } else {
+            console.warn(
+              "Unsupported object type ({type}) for Undo activity: {object}",
+              { type: activity?.constructor.name, object: activity },
+            );
```
dahlia: Warnings should be logged using a LogTape logger.

packages/relay/src/relay.ts
```diff
@@ -48,648 +45,118 @@ export interface RelayOptions {
...
+
+relayBuilder.setActorDispatcher(
+  "/users/{identifier}",
+  async (ctx, identifier) => {
+    if (identifier !== RELAY_SERVER_ACTOR) return null;
+    const keys = await ctx.getActorKeyPairs(identifier);
+    return new Application({
+      id: ctx.getActorUri(identifier),
+      preferredUsername: identifier,
+      name: "ActivityPub Relay",
```
dahlia: This shouldn't be hard-coded like this. Could we make it configurable (at least with a default value)?

sij411:
> ### On code duplication
> I notice there's quite a bit of duplicated code between `LitePubRelay` and `MastodonRelay`. For example:
> 
> * The class fields, `constructor`, and `fetch()` method are nearly identical
> * The validation logic in the `Follow` handler (null checks, public follow check, etc.) is the same
> * The `Undo` handler is completely identical
> * The `Accept`/`Reject` activity creation follows the same pattern
> * Within `LitePubRelay` itself, the `Create`, `Update`, `Move`, `Delete`, and `Announce` handlers repeat almost the same pattern
> 
> Now, coincidental duplication—where code happens to look similar but serves different purposes—is often fine to leave as-is. But the duplication here appears to be essential rather than coincidental: it's the same logic serving the same purpose, just copied in multiple places.
> 
> The problem with this kind of duplication is that when you need to fix or change the logic, you have to update multiple places at once. That's tedious, but more importantly, it's easy to accidentally fix one place and forget the other—leading to bugs. In fact, the `published` timestamp issue [that Gemini flagged earlier](https://github.com/fedify-dev/fedify/pull/490#discussion_r2612994958) is a good example: some handlers had it, others didn't, precisely because of this repeated pattern.
> 
> There are various ways to extract common logic (e.g., an abstract base class, shared helper functions, etc.). I'll leave the specific approach up to you, but it would be great to consolidate the places where the same logic is copy-pasted.

Got it. I'd like to implement some helper functions to remove duplicated logics since that feels like the most fedify-like way. What do you think about this approach?

dahlia:
Apparently the *deno.lock* file seems broken. Try to remove it and run `deno task install` to regenerate it.

deno.lock
```diff
@@ -2644,6 +2644,8 @@
         "flat-cache"
       ]
     },
+=======
+>>>>>>> ac923ced (Fix deno.lock)
```
dahlia: I think your lock file is corrupt due to these conflict lines. Please remove the *deno.lock* file at all, and then run `deno task install` to regenerate one!

sij411: oh my i was fixing it

sij411:
Merry Christmas 🎅 I refactored Relay classes and the package's structure as @dahlia reviewed previously.

## Relay Package Refactoring (80858b6f → ac2ada3b)

  Major refactoring to eliminate code duplication between MastodonRelay and LitePubRelay implementations using Template Method pattern.

  Stats: 15 commits, 12 files changed, 541 insertions(+), 489 deletions(-)

  ---
  ## Key Changes

  1. New Architecture

  - Added BaseRelay abstract class (src/base.ts) - common functionality for both relay types
  - Both MastodonRelay and LitePubRelay now extend BaseRelay
  - Each implements setupInboxListeners() for protocol-specific behavior

  2. Code Extracted into Modules

  - types.ts: Type definitions and constants
  - builder.ts: FederationBuilder setup and shared dispatchers
  - follow.ts: Follow/Undo activity validation and response handling
  - factory.ts: createRelay() factory function
  - base.ts: Abstract base class

  3. Code Reduction

  - MastodonRelay: -125 lines (uses direct forwarding)
  - LitePubRelay: -180 lines (uses Announce wrapping)
  - Deleted old monolithic relay.ts (283 lines)

  4. Documentation

  - Updated README.md to reflect new structure
  - Fixed package.json typo

  ---
  Review Checklist

  Must verify:
  - Public API remains compatible
  - Mastodon and LitePub protocols still work correctly
  - No circular dependencies introduced
  - Error handling preserved
  - Tests cover both implementations

  Key files to review:
  1. src/base.ts - abstract base class
  2. src/mastodon.ts & src/litepub.ts - protocol implementations
  3. src/factory.ts - factory pattern
  4. src/builder.ts & src/follow.ts - extracted helpers
  
  ---
  
  p.s.
  
  - I left Litepub specific Accept inbox listener codes without extracting it with helper functions because only Litepub needs to handle Accept responses. Not sure on this is the right decision?
  
- For minor fixes check out this [commit](https://github.com/fedify-dev/fedify/pull/490/commits/80858b6fe9b807e06baa935b268728a9102052dd)

packages/relay/src/mod.ts
```diff
@@ -7,7 +7,15 @@
  *
  * @module
  */
-
-// Export relay functionality here
-export type { Relay, RelayOptions } from "./relay.ts";
-export { LitePubRelay, MastodonRelay } from "./relay.ts";
+export { BaseRelay } from "./base.ts";
```
dahlia: I don't think this needs to be re-exported here. IMHO it's better to keep it internal.

packages/relay/src/litepub.ts
```diff
@@ -0,0 +1,170 @@
...
+          await ctx.data.kv.set(
+            ["follower", followerActor.id.href],
+            updatedFollowerData,
+          );
+
+          // Update followers list
+          const followers = await ctx.data.kv.get<string[]>(["followers"]) ??
+            [];
+          followers.push(followerActor.id.href);
+          await ctx.data.kv.set(["followers"], followers);
```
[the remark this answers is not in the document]
dahlia: No, the `KvStore` interface in Fedify does *not* provide an `atomic()` method. Instead, we recently got a `list()` method. We could utilize it instead.

packages/relay/src/mastodon.ts
```diff
@@ -0,0 +1,103 @@
...
+          if (this.options.subscriptionHandler) {
+            approved = await this.options.subscriptionHandler(ctx, follower);
+          }
+
+          if (approved) {
+            // Mastodon-specific: immediately add to followers list
+            const followers = await ctx.data.kv.get<string[]>(["followers"]) ??
+              [];
+            followers.push(follower.id.href);
+            await ctx.data.kv.set(["followers"], followers);
```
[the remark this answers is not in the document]
dahlia: Same here. We should a `KvStore.list()` method instead of `atomic()` which is non-existent.

packages/relay/src/follow.ts
```diff
@@ -0,0 +1,105 @@
...
+  if (activity instanceof Follow) {
+    if (activity.id == null || activity.actorId == null) return;
+
+    const followers = await ctx.data.kv.get<string[]>(["followers"]) ?? [];
+    const updatedFollowers = followers.filter((id) =>
+      id !== activity.actorId?.href
+    );
+
+    await ctx.data.kv.set(["followers"], updatedFollowers);
+    await ctx.data.kv.delete(["follower", activity.actorId?.href]);
```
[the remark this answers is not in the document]
dahlia: Fair enough, except that we don't have such method. We could use the `KvStore.list()` method instead though.

sij411:
Alright, since those race condition problems seem bit out of this PR scope, I'll be handling in another PR. I'll keep those in mind and resolve conversations for now.

packages/relay/src/mod.ts
sij411: Fair, resolved with the latest [commit](https://github.com/fedify-dev/fedify/pull/490/commits/46b317c682766558f8b0913d41ed2469b125a96d)

PR #500, opened 2025-12-22
Add `list()` method to `KvStore` interface

### LINKED ISSUE #498: Add optional `list()` operation to `KvStore`
### Motivation

The current `KvStore` interface provides basic key–value operations (`get`, `set`, `delete`) and an optional `cas()` for compare-and-swap. However, there is no way to enumerate keys matching a certain prefix, which limits the kinds of data structures that can be efficiently built on top of `KvStore`.

A concrete example is the distributed trace storage proposed in #497. To store trace data efficiently, we want to write each activity record under its own key (e.g., `["traces", traceId, spanId]`) and later retrieve all records for a given trace by scanning keys with prefix `["traces", traceId]`. Without a `list()` operation, the only alternative is to store all records for a trace as a single list value and use `cas()` for concurrent appends—this works but becomes inefficient as the list grows and may suffer from contention under high write loads.

Beyond trace storage, a `list()` operation would enable other use cases such as:

- Enumerating all cached public keys for a given actor
- Listing all pending outbox messages for debugging
- Implementing cache invalidation by prefix

Most key–value store backends already support prefix scanning natively, so exposing this capability through the `KvStore` interface is straightforward.

### Proposed interface

The `list()` method should be added as an *optional* method in Fedify 1.10.0 to maintain backward compatibility with existing `KvStore` implementations:

```typescript
interface KvStore {
  get<T = unknown>(key: KvKey): Promise<T | undefined>;
  set(key: KvKey, value: unknown, options?: KvStoreSetOptions): Promise<void>;
  delete(key: KvKey): Promise<void>;

  // Optional (since 1.8.0)
  cas?: (
    key: KvKey,
    expectedValue: unknown,
    newValue: unknown,
    options?: KvStoreSetOptions,
  ) => Promise<boolean>;

  // Optional (since 1.10.0)
  list?: (
    options: KvStoreListOptions,
  ) => AsyncIterable<KvStoreListEntry>;
}

interface KvStoreListOptions {
  prefix: KvKey;
}

interface KvStoreListEntry {
  key: KvKey;
  value: unknown;
}
```

The method returns an `AsyncIterable` to support backends that may need to paginate results internally. Callers can use `for await...of` to iterate through entries:

```typescript
for await (const entry of kv.list({ prefix: ["traces", traceId] })) {
  console.log(entry.key, entry.value);
}
```

### Implementation for existing backends

Each official `KvStore` implementation should add support for `list()`. Most backends already provide native prefix scanning capabilities that can be directly leveraged:

- `MemoryKvStore`: Filter in-memory keys by prefix
- `RedisKvStore`: Use `SCAN` with a pattern
- `PostgresKvStore` and `SqliteKvStore`: Use `LIKE` queries
- `DenoKvStore`: Delegate to Deno KV's built-in `list()` method

The implementation should be straightforward for all backends since prefix scanning is a common operation supported by most storage systems.

### Migration path

In Fedify 1.10.0, `list()` will be optional. Code that depends on `list()` should check for its presence and fall back to alternative approaches (such as `cas()`-based list management) when it is not available.

In Fedify 2.0.0, `list()` will become a required method on the `KvStore` interface. This gives `KvStore` implementers approximately one major version cycle to add support.

### Related issues

This is a prerequisite for #497 (distributed trace storage for the debug dashboard).

dahlia:
## Summary

This PR adds an optional `list()` method to the `KvStore` interface for enumerating entries by key prefix, as proposed in #498.

Closes #498.

## Changes from the original proposal

The original proposal suggested:

```typescript
list?: (options: KvStoreListOptions) => AsyncIterable<KvStoreListEntry>;

interface KvStoreListOptions {
  prefix: KvKey;
}
```

However, during implementation, I simplified the API to:

```typescript
list?(prefix?: KvKey): AsyncIterable<KvStoreListEntry>;
```

Key differences:

1. **Removed `KvStoreListOptions` interface**: Since the options object only contained a single `prefix` property, passing the prefix directly as an argument is simpler and more ergonomic.

2. **Made prefix optional**: When `prefix` is omitted or an empty array, `list()` returns all entries in the store.  This is useful for debugging and administrative purposes.

## Implementations

Implemented `list()` in all official `KvStore` backends:

| Backend | Strategy |
|---------|----------|
| `MemoryKvStore` | Filter in-memory keys by prefix |
| `SqliteKvStore` | `LIKE` query with JSON key pattern |
| `PostgresKvStore` | Array slice comparison (`key[1:n] = prefix`) |
| `RedisKvStore` | `SCAN` with pattern matching + key deserialization |
| `DenoKvStore` | Delegate to Deno KV's built-in `list()` |
| `WorkersKvStore` | KV `list()` with JSON key prefix pattern |

## Bonus: Cloudflare Workers testing infrastructure

While implementing `WorkersKvStore.list()`, I noticed that the existing tests in `@fedify/cfworkers` were completely skipped (they checked for `navigator.userAgent === "Cloudflare-Workers"`).

I set up proper testing using Vitest with `@cloudflare/vitest-pool-workers`, which runs tests in a simulated Cloudflare Workers runtime environment. This includes:

- *wrangler.jsonc* with KV namespace and Queue bindings
- *vitest.config.ts* with Workers pool configuration
- Separate test files for `WorkersKvStore` and `WorkersMessageQueue`
- Mock `Temporal.Duration` objects (Workers runtime doesn't support Temporal)

## Test plan

- [x] `MemoryKvStore.list()` tests
- [x] `SqliteKvStore.list()` tests
- [x] `PostgresKvStore.list()` tests (requires PostgreSQL)
- [x] `RedisKvStore.list()` tests (requires Redis)
- [x] `DenoKvStore.list()` tests
- [x] `WorkersKvStore.list()` tests
- [x] Documentation updated in *docs/manual/kv.md*
- [x] CHANGES.md updated

### DIFF
--- a/packages/cfworkers/src/mod.ts
+++ b/packages/cfworkers/src/mod.ts
@@ -90,40 +90,22 @@ export class WorkersKvStore implements KvStore {
    * @since 1.10.0
    */
   async *list(prefix?: KvKey): AsyncIterable<KvStoreListEntry> {
+    let pattern: string;
+    let exactKey: string | null = null;
+
     if (prefix == null || prefix.length === 0) {
       // Empty prefix: list all entries
       // JSON encoded keys start with '[', so prefix with '[' to match all arrays
-      let cursor: string | undefined;
-      do {
-        const result = await this.#namespace.list<KvMetadata>({
-          prefix: "[",
-          cursor,
-        });
-        cursor = result.list_complete ? undefined : result.cursor;
-
-        for (const keyInfo of result.keys) {
-          const metadata = keyInfo.metadata as KvMetadata | undefined;
-          if (metadata?.expires != null && metadata.expires < Date.now()) {
-            continue;
-          }
-
-          const value = await this.#namespace.get(keyInfo.name, "json");
-          if (value == null) continue;
-
-          yield {
-            key: JSON.parse(keyInfo.name) as KvKey,
-            value,
-          };
-        }
-      } while (cursor != null);
+      pattern = "[";
     } else {
       // Keys are JSON encoded: '["prefix","a"]'
       // Pattern to match keys starting with prefix: '["prefix",' matches children
-      // Also check for exact match: '["prefix"]'
-      const exactKey = this.#encodeKey(prefix);
-      const childrenPattern = JSON.stringify(prefix).slice(0, -1) + ",";
+      exactKey = this.#encodeKey(prefix);
+      pattern = JSON.stringify(prefix).slice(0, -1) + ",";
+    }
 
-      // First, check if the exact prefix key exists
+    // First, check if the exact prefix key exists
+    if (exactKey != null) {
       const { value, metadata } = await this.#namespace.getWithMetadata(
         exactKey,
         "json",
@@ -134,36 +116,36 @@ export class WorkersKvStore implements KvStore {
           (metadata as KvMetadata).expires! >= Date.now())
       ) {
         yield {
-          key: prefix,
+          key: prefix!,
           value,
         };
       }
-
-      // Then list all keys starting with prefix
-      let cursor: string | undefined;
-      do {
-        const result = await this.#namespace.list<KvMetadata>({
-          prefix: childrenPattern,
-          cursor,
-        });
-        cursor = result.list_complete ? undefined : result.cursor;
-
-        for (const keyInfo of result.keys) {
-          const metadata = keyInfo.metadata as KvMetadata | undefined;
-          if (metadata?.expires != null && metadata.expires < Date.now()) {
-            continue;
-          }
-
-          const value = await this.#namespace.get(keyInfo.name, "json");
-          if (value == null) continue;
-
-          yield {
-            key: JSON.parse(keyInfo.name) as KvKey,
-            value,
-          };
-        }
-      } while (cursor != null);
     }
+
+    // List all keys matching the pattern
+    let cursor: string | undefined;
+    do {
+      const result = await this.#namespace.list<KvMetadata>({
+        prefix: pattern,
+        cursor,
+      });
+      cursor = result.list_complete ? undefined : result.cursor;
+
+      for (const keyInfo of result.keys) {
+        const metadata = keyInfo.metadata as KvMetadata | undefined;
+        if (metadata?.expires != null && metadata.expires < Date.now()) {
+          continue;
+        }
+
+        const value = await this.#namespace.get(keyInfo.name, "json");
+        if (value == null) continue;
+
+        yield {
+          key: JSON.parse(keyInfo.name) as KvKey,
+          value,
+        };
+      }
+    } while (cursor != null);
   }
 }
 


packages/cfworkers/src/mod.ts
```diff
@@ -83,6 +84,87 @@ export class WorkersKvStore implements KvStore {
...
+          if (value == null) continue;
+
+          yield {
+            key: JSON.parse(keyInfo.name) as KvKey,
+            value,
+          };
+        }
+      } while (cursor != null);
+    }
+  }
```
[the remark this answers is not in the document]
dahlia: While extracting a helper function would reduce duplication, I'd prefer to keep the current inline approach for now since:

1. The duplicated block is relatively small (~10 lines)
2. Keeping it inline makes the control flow easier to follow
3. The `bind(this)` pattern for generator functions adds complexity

I'll keep this in mind for future refactoring if the method grows more complex.

packages/redis/src/kv.ts
```diff
@@ -104,4 +109,88 @@ export class RedisKvStore implements KvStore {
...
+          const encodedValue = await this.#redis.getBuffer(key);
+          if (encodedValue == null) continue;
+          yield {
+            key: this.#deserializeKey(key),
+            value: this.#codec.decode(encodedValue),
+          };
+        }
+      } while (cursor !== "0");
+    }
+  }
```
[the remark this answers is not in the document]
dahlia: Same reasoning as above; I'll keep the inline approach for now to maintain readability, but will consider extracting if the logic becomes more complex.

packages/denokv/src/mod.ts
```diff
@@ -93,6 +94,35 @@ export class DenoKvStore implements KvStore {
...
+    // Then list all keys starting with the prefix
+    const entries = this.#kv.list({ prefix: prefixToUse });
+    for await (const entry of entries) {
+      if (entry.value == null) continue;
+      yield {
+        key: entry.key as KvKey,
+        value: entry.value,
+      };
+    }
+  }
```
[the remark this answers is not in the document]
dahlia: I investigated this and found that Deno KV's `list({ prefix })` does *not* include the exact prefix key; it only returns strictly longer keys.

For example, `kv.list({ prefix: ["exact"] })` returns `["exact", "child1"]` but *not* `["exact"]` itself.

So the explicit check is actually necessary to include the exact prefix key in results. I added a regression test in 2ce6a42478d9a04f2f359c7adcfc866b9d4b32b4 to verify this behavior.

docs/manual/kv.md
```diff
@@ -683,6 +692,77 @@ async cas(
...
+
+*This API is available since Fedify 1.10.0.*
+
+If your storage backend supports prefix scanning, you can implement the
+`~KvStore.list()` method.  This method allows you to enumerate all entries
+whose keys start with a given prefix.  This is useful for implementing
+batch operations or iterating over related entries.
+
+~~~~ typescript twoslash
+import type { KvStore, KvKey, KvStoreSetOptions } from "@fedify/fedify";
```
sij411: Line 705-706 don't have to be separated if you want to use all four types in this code block.

packages/cfworkers/src/mod.ts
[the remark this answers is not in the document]
2chanhaeng: Then how about :
```suggestion
    let pattern = "[";
    if (prefix != null && prefix.length !== 0) {
      // Keys are JSON encoded: '["prefix","a"]'
      // Pattern to match keys starting with prefix: '["prefix",' matches children
      // Also check for exact match: '["prefix"]'
      const exactKey = this.#encodeKey(prefix);

      // First, check if the exact prefix key exists
      // ...
      pattern = JSON.stringify(prefix).slice(0, -1) + ",";
    }
    // Then list all keys starting with prefix
    let cursor: string | undefined;
    do {
      const result = await this.#namespace.list<KvMetadata>({
        prefix: pattern,
        cursor,
      });
      // ...
    } while (cursor != null);
```

packages/redis/src/kv.ts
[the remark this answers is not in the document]
2chanhaeng: Then how about:

```suggestion
    let pattern: string;
    if (prefix == null || prefix.length === 0) {
      // Empty prefix: scan for all keys with the key prefix
      pattern = `${prefixKeyStr}*`;
    } else {
      const prefixKey = this.#serializeKey(prefix);
      const prefixKeyFullStr = typeof prefixKey === "string"
        ? prefixKey
        : new TextDecoder().decode(new Uint8Array(prefixKey));

      // First, check if the exact prefix key exists
      const exactValue = await this.#redis.getBuffer(prefixKey);
      if (exactValue != null) {
        yield {
          key: prefix,
          value: this.#codec.decode(exactValue),
        };
      }

      // Then scan for all keys starting with prefix::
      pattern = `${prefixKeyFullStr}::*`;
    }
    let cursor = "0";
    do {
      // ...
    } while (cursor !== "0");
```

packages/fedify/src/federation/kv.ts
```diff
@@ -148,4 +175,33 @@ export class MemoryKvStore implements KvStore {
...
+      // Check prefix match
+      if (prefix != null) {
+        if (key.length < prefix.length) continue;
+        let matches = true;
+        for (let i = 0; i < prefix.length; i++) {
+          if (key[i] !== prefix[i]) {
+            matches = false;
+            break;
+          }
+        }
```
2chanhaeng: This is good but how about:
```suggestion
        const matches = prefix.every((p, i) => p === key[i]);
```

packages/fedify/src/federation/kv.test.ts
```diff
@@ -34,4 +35,80 @@ test("MemoryKvStore", async (t) => {
...
+    await store.set(["prefix", "b"], "value-b");
+    await store.set(["prefix", "nested", "c"], "value-c");
+    await store.set(["other", "x"], "value-x");
+    await store.set(["prefix"], "exact-match");
+
+    // Test: list with prefix
+    const entries: { key: KvKey; value: unknown }[] = [];
+    for await (const entry of store.list!(["prefix"])) {
+      entries.push(entry);
+    }
```
2chanhaeng: This pattern can probably be simplified as follows:
```suggestion
    const entries = await Array.fromAsync(store.list!([/* prefix */]));
```

It is also found on 59-62, 81-84, 102-105.

packages/denokv/src/mod.ts
```diff
@@ -93,6 +94,35 @@ export class DenoKvStore implements KvStore {
...
+   * {@inheritDoc KvStore.list}
+   * @since 1.10.0
+   */
+  async *list(prefix?: KvKey): AsyncIterable<KvStoreListEntry> {
+    const prefixToUse = prefix ?? [];
+
+    // First, check if the exact prefix key exists (only if prefix is specified)
+    if (prefix != null && prefix.length > 0) {
+      const exactEntry = await this.#kv.get(prefix);
+      if (exactEntry.value != null) {
```
2chanhaeng: This excluding null value pattern isn't found in other integrations and is only visible here. Unless there's a specific issue with DenoKV, I think it would be appropriate to either modify `@fedify/fedify`'s `federation/kv.ts` to apply to all KV stores, or simply remove it here as well.

packages/redis/src/kv.ts
```diff
@@ -104,4 +109,88 @@ export class RedisKvStore implements KvStore {
     const serializedKey = this.#serializeKey(key);
     await this.#redis.del(serializedKey);
   }
+
+  #deserializeKey(redisKey: string): KvKey {
+    const prefixStr = typeof this.#keyPrefix === "string"
+      ? this.#keyPrefix
+      : new TextDecoder().decode(new Uint8Array(this.#keyPrefix));
```
2chanhaeng: This decoding buffer redis key pattern duplicate third times(114-116, 128-130, 158-160). It would look better if you made it into a function.

2chanhaeng:
LGTM!

sij411 (approved):
It looks good to me :)

packages/fedify/src/federation/kv.ts
async3619, line 101: just curiosity) may I ask why did you use `Record<...>` instead of `Map<K, V>` (or some containers that supports hash comparison)? that would prevent you to use type conversion when loop through `#values`.

It would be like this:

```ts
#values = new KeyValueContainer<KvKey, [...]>(/* comparator if needed... */)
```

and then:

```ts
  async *list(prefix?: KvKey): AsyncIterable<KvStoreListEntry> {
    const now = Temporal.Now.instant();
    for (const [key, entry] of Object.entries(this.#values)) {
	//          ^^^ Now type conversion is not needed \o/
```

sij411: i think Map approach just hides type conversions with extra classes?

sij411: I mean type conversions not appearing inside of that for loop, but gonna happen in another file anyway. It's just my intuitions tho.

dahlia: Good question! The reason I used `Record<string, ...>` with JSON-encoded keys instead of `Map<KvKey, ...>` is that JavaScript's `Map` uses [SameValueZero algorithm](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#key_equality) for key comparison, which is essentially `===` (strict equality) for objects. Since `KvKey` is an array (`KvKeyComponent[]`), two arrays with identical contents are still considered different objects:

```javascript
const key1 = ["user", "123"];
const key2 = ["user", "123"];
console.log(key1 === key2); // false - different references

const map = new Map();
map.set(key1, "value");
map.get(key2); // undefined - key2 is treated as a different key
```

This means if we used `Map<KvKey, ...>`, we wouldn't be able to retrieve or delete values using a different array instance with the same contents. By serializing keys to JSON strings, we get value-based equality comparison that works correctly for our use case.

PR #501, opened 2025-12-23
Separate vocabularies to `@fedify/vocab`

### LINKED ISSUE #437: Extract vocab from @fedify/fedify into @fedify/vocab
### Summary

Currently all the vocabulary lives in the following:
```
import * as vocab from "@fedify/fedify/vocab";
// or
import { Object as ASObject } from "@fedify/fedify";
```

To support reusing and perhaps more easily extending the vocabs (#207), we should probably extract the vocab code to two packages:
- `@fedify/vocab` which includes all of `@fedify/fedify/vocab`
- `@fedify/vocab-tools` which has the codegen for vocabs and supports compiling the yaml files to a vocab package. 

i.e., `GENPATH=vocab-$(deno eval \"console.log(crypto.randomUUID());\").ts && deno run --allow-read --allow-write --check src/codegen/main.ts src/vocab/ ../runtime/ src/vocab/$GENPATH && deno fmt src/vocab/$GENPATH && mv src/vocab/$GENPATH src/vocab/vocab.ts && deno cache src/vocab/vocab.ts && deno check src/vocab/vocab.ts`

Would be a thing that you could run from like `npx @fedify/vocab-tools compile` or something

Along with this, we should, if we haven't already, deprecate pulling in vocab types directly from the `@fedify/fedify` exports.

### Target Area

vocab & codegen

### Motivation

Improve the ability to use vocabs outside of the Fedify code (e.g., in ActivityPub API client code)

### Risk Assessment

_No response_

### Benefits

_No response_

2chanhaeng:
## Summary

Separate vocabularies to `@fedify/vocab`.

## Related Issue

- closes #437

## Changes

This PR has changes across the entire codebase. This is because the Vocabulary API, which was separated in this PR, was being used throughout the codebase. However, since most of the changes occur in import statements, there shouldn't be any complex changes.

## Checklist

- [x] ~~Did you write some relevant docs about this change (if it's a new feature)?~~

## Additional Notes

In issue #437, it was suggested to completely separate `@fedify/fedify/vocab`. However, `@fedify/fedify/vocab` references `@fedify/fedify/webfinger`. Therefore, to fully separate `@fedify/vocab`, `webfinger` must also be separated. @dahlia  suggested that separating `webfinger` would be burdensome, so I implemented it based on the proposal to only separate the generated `vocab.ts`.

### DIFF
--- a/packages/fedify/deno.json
+++ b/packages/fedify/deno.json
@@ -101,12 +101,13 @@
       ]
     },
     "pnpm:build-vocab-runtime": {
-      "command": "cd ../vocab-runtime && pnpm build"
+      "command": "pnpm run --filter vocab-runtime build"
     },
     "pnpm:build-vocab-tools": {
-      "command": "cd ../vocab-tools && pnpm build"
+      "command": "pnpm run --filter vocab-tools build"
     },
     "pnpm:build-vocab": {
+      "command": "pnpm run --filter vocab build",
       "dependencies": [
         "pnpm:build-vocab-runtime",
         "pnpm:build-vocab-tools"
--- a/packages/fedify/package.json
+++ b/packages/fedify/package.json
@@ -155,11 +155,11 @@
     "wrangler": "^4.17.0"
   },
   "scripts": {
-    "build": "deno task codegen && pnpm run --filter vocab-runtime build && pnpm run --filter vocab-tools build && tsdown",
+    "build": "deno task codegen && deno task pnpm:build-vocab && tsdown",
     "prepack": "pnpm run build",
     "prepublish": "pnpm run build",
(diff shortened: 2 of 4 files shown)


2chanhaeng:
As I wrote in the Additional Notes, I originally planned to separate the entire `@fedify/fedify/vocab` into `@fedify/vocab`, but I couldn't do it because of the dependency on `@fedify/fedify/webfinger`.
After discussing this issue with @dahlia, we agreed that it's not a good idea to only separate the generated Vocabulary API code into `@fedify/vocab`. Instead, we both agreed that it's better to also separate `@fedify/fedify/webfinger` into `@fedify/webfinger` and follow the original plan. So, I close this PR and rework on a new PR that separates `@fedify/webfinger` and `@fedify/vocab`.

PR #502, opened 2025-12-23
Add `FedifySpanExporter` for distributed trace storage

### LINKED ISSUE #497: Distributed trace storage for the debug dashboard
### Background

With PR #483, Fedify now has comprehensive OpenTelemetry instrumentation that records span events containing full activity JSON payloads. This enables developers to build custom observability tools by implementing a `SpanExporter` that captures ActivityPub activities as they flow through the system.

However, the example `FedifyDebugExporter` shown in the documentation stores captured activities in an in-memory `Map`. This works fine when the entire application runs in a single process, but Fedify applications commonly run in distributed environments where:

- The web server handling HTTP requests runs on different nodes than the background workers processing the message queue.
- Multiple worker nodes may process queued messages in parallel.
- The debug dashboard itself may run on yet another node.

In such environments, a custom `SpanExporter` with in-memory storage cannot aggregate traces across nodes. Each node would only see its own spans, making it impossible to view the complete picture of a distributed trace.

### Proposed solution

We should provide a `SpanExporter` implementation that persists trace data to Fedify's `KvStore`, which is already designed to be shared across nodes. This allows all nodes in a distributed deployment to write to the same storage, and the debug dashboard can query this shared storage to display complete traces.

The key insight is that OpenTelemetry's context propagation (which Fedify already implements via the `traceContext` field in `InboxMessage`, `OutboxMessage`, and `FanoutMessage`) ensures that all spans belonging to the same logical request share the same trace ID, even when they execute on different nodes. By storing spans keyed by their trace ID, we can later retrieve all activities that occurred within a single request.

A rough sketch of the API might look like:

```typescript
import { createFederation } from "@fedify/fedify";
import { RedisKvStore } from "@fedify/redis";
import { FedifySpanExporter } from "@fedify/fedify/otel";
import { NodeTracerProvider, SimpleSpanProcessor } from "@opentelemetry/sdk-trace-node";

const kv = new RedisKvStore(redis);

// Create the exporter that writes to KvStore
const fedifyExporter = new FedifySpanExporter(kv, {
  ttl: Temporal.Duration.from({ hours: 1 }),
});

const tracerProvider = new NodeTracerProvider();
tracerProvider.addSpanProcessor(new SimpleSpanProcessor(fedifyExporter));

const federation = createFederation({
  kv,
  tracerProvider,
});

// Later, in the debug dashboard endpoint:
app.get("/debug/traces/:traceId", async (req, res) => {
  const activities = await fedifyExporter.getActivitiesByTraceId(req.params.traceId);
  res.json(activities);
});

app.get("/debug/traces", async (req, res) => {
  const recentTraces = await fedifyExporter.getRecentTraces({ limit: 100 });
  res.json(recentTraces);
});
```

### Storage strategy

There are two possible approaches for storing trace data, depending on what operations the `KvStore` supports.

When the `KvStore` provides a `list()` operation (see #498), each activity record can be stored under its own key using a pattern like `["traces", traceId, spanId]`. Retrieving all activities for a trace becomes a prefix scan over `["traces", traceId]`. This approach is simple and avoids any concurrency issues since each span writes to a unique key.

When the `KvStore` only provides `cas()` but not `list()`, activities must be stored as a list under a single key `["traces", traceId]`. Appending a new activity requires reading the existing list, adding the new item, and writing it back using compare-and-swap to handle concurrent writes. This works but is less efficient and may experience contention under high load.

If the `KvStore` provides neither `list()` nor `cas()`, the exporter should throw an error at construction time with a clear message explaining which `KvStore` capabilities are required.

### Data model

Each stored record should contain enough information to reconstruct the activity flow:

```typescript
interface TraceActivityRecord {
  traceId: string;
  spanId: string;
  parentSpanId?: string;
  direction: "inbound" | "outbound";
  activityType: string;
  activityId?: string;
  activityJson: string;
  verified?: boolean;
  timestamp: string;
  // For outbound activities
  inboxUrl?: string;
}
```

The exporter should extract this information from the span events added in PR #483, specifically `activitypub.activity.received` and `activitypub.activity.sent`.

### Relationship to other issues

This issue depends on #498 (adding optional `list()` operation to `KvStore`), which should be implemented first. While the exporter can fall back to `cas()` for `KvStore` implementations that lack `list()`, the `list()` operation provides a more robust and efficient solution.

This issue is part of the larger effort to build a real-time ActivityPub debug dashboard (originally discussed in #234 and #323).

dahlia:
### Summary

This PR implements `FedifySpanExporter`, a new OpenTelemetry `SpanExporter` that persists ActivityPub activity traces to a `KvStore`. This enables distributed tracing support across multiple nodes in a Fedify deployment, which is essential for building debug dashboards that can show complete request flows across web servers and background workers.

Closes #497

### Usage

The exporter captures activity data from OpenTelemetry span events (`activitypub.activity.received` and `activitypub.activity.sent`) and stores them with full trace context preserved:

```typescript
import { MemoryKvStore } from "@fedify/fedify";
import { FedifySpanExporter } from "@fedify/fedify/otel";
import {
  BasicTracerProvider,
  SimpleSpanProcessor,
} from "@opentelemetry/sdk-trace-base";

const kv = new MemoryKvStore();
const exporter = new FedifySpanExporter(kv, {
  ttl: Temporal.Duration.from({ hours: 1 }),
});

const provider = new BasicTracerProvider();
provider.addSpanProcessor(new SimpleSpanProcessor(exporter));
```

The stored traces can then be queried for display in a debug dashboard:

```typescript
// Get all activities for a specific trace
const activities = await exporter.getActivitiesByTraceId(traceId);

// Get recent traces with summary information
const recentTraces = await exporter.getRecentTraces({ limit: 100 });
```

### Storage strategies

The exporter supports two storage strategies depending on the `KvStore` capabilities. When the `list()` method is available (preferred), it stores individual records with keys like `[prefix, traceId, spanId]`. When only `cas()` is available, it uses compare-and-swap operations to append records to arrays stored per trace.

### Changelog

For a detailed list of added types and interfaces, see the diff in *CHANGES.md*.

### Test plan

- [x] Unit tests for `FedifySpanExporter` class (22 test cases)
- [x] Tests pass in Deno environment (`deno task test`)
- [x] Tests pass in Node.js environment (`pnpm test`)

### DIFF
--- a/packages/fedify/src/otel/exporter.test.ts
+++ b/packages/fedify/src/otel/exporter.test.ts
@@ -124,7 +124,9 @@ test("FedifySpanExporter", async (t) => {
     const kv = new MemoryKvStore();
     const exporter = new FedifySpanExporter(kv);
 
-    const activityJson = JSON.stringify({
+    const traceId = "trace123";
+    const spanId = "span456";
+    const activity = {
       "@context": "https://www.w3.org/ns/activitystreams",
       type: "Create",
       id: "https://example.com/activities/123",
@@ -133,11 +135,12 @@ test("FedifySpanExporter", async (t) => {
         type: "Note",
         content: "Hello!",
       },
-    });
+    };
+    const activityJson = JSON.stringify(activity);
 
     const span = createMockSpan({
-      traceId: "trace123",
-      spanId: "span456",
+      traceId,
+      spanId,
       name: "activitypub.inbox",
       events: [
         createActivityReceivedEvent({
@@ -155,16 +158,13 @@ test("FedifySpanExporter", async (t) => {
       });
     });
 
-    const activities = await exporter.getActivitiesByTraceId("trace123");
+    const activities = await exporter.getActivitiesByTraceId(traceId);
     assertEquals(activities.length, 1);
-    assertEquals(activities[0].traceId, "trace123");
-    assertEquals(activities[0].spanId, "span456");
+    assertEquals(activities[0].traceId, traceId);
+    assertEquals(activities[0].spanId, spanId);
     assertEquals(activities[0].direction, "inbound");
-    assertEquals(activities[0].activityType, "Create");
-    assertEquals(
-      activities[0].activityId,
-      "https://example.com/activities/123",
-    );
+    assertEquals(activities[0].activityType, activity.type);
+    assertEquals(activities[0].activityId, activity.id);
     assertEquals(activities[0].activityJson, activityJson);
     assertEquals(activities[0].verified, true);
   });
@@ -175,23 +175,27 @@ test("FedifySpanExporter", async (t) => {
       const kv = new MemoryKvStore();
       const exporter = new FedifySpanExporter(kv);
 
-      const activityJson = JSON.stringify({
+      const traceId = "trace789";
+      const spanId = "span012";
+      const inboxUrl = "https://example.com/users/alice/inbox";
+      const activity = {
         "@context": "https://www.w3.org/ns/activitystreams",
         type: "Follow",
         id: "https://myserver.com/activities/789",
         actor: "https://myserver.com/users/bob",
         object: "https://example.com/users/alice",
-      });
+      };
+      const activityJson = JSON.stringify(activity);
 
       const span = createMockSpan({
-        traceId: "trace789",
-        spanId: "span012",
+        traceId,
+        spanId,
         name: "activitypub.send_activity",
         events: [
           createActivitySentEvent({
             activityJson,
-            inboxUrl: "https://example.com/users/alice/inbox",
-            activityId: "https://myserver.com/activities/789",
+            inboxUrl,
+            activityId: activity.id,
           }),
         ],
       });
@@ -203,20 +207,14 @@ test("FedifySpanExporter", async (t) => {
         });
       });
 
-      const activities = await exporter.getActivitiesByTraceId("trace789");
+      const activities = await exporter.getActivitiesByTraceId(traceId);
       assertEquals(activities.length, 1);
-      assertEquals(activities[0].traceId, "trace789");
-      assertEquals(activities[0].spanId, "span012");
+      assertEquals(activities[0].traceId, traceId);
+      assertEquals(activities[0].spanId, spanId);
       assertEquals(activities[0].direction, "outbound");
-      assertEquals(activities[0].activityType, "Follow");
-      assertEquals(
-        activities[0].activityId,
-        "https://myserver.com/activities/789",
-      );
-      assertEquals(
-        activities[0].inboxUrl,
-        "https://example.com/users/alice/inbox",
-      );
+      assertEquals(activities[0].activityType, activity.type);
+      assertEquals(activities[0].activityId, activity.id);
+      assertEquals(activities[0].inboxUrl, inboxUrl);
     },
   );
 


packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,560 @@
...
+  /**
+   * Gets recent traces with summary information.
+   *
+   * @param options Options for the query.
+   * @returns An array of trace summaries.
+   */
+  async getRecentTraces(
+    options?: GetRecentTracesOptions,
+  ): Promise<TraceSummary[]> {
+    const summaryPrefix: KvKey = [...this.#keyPrefix, "_summaries"] as KvKey;
```
async3619: trivial)

```suggestion
    const summaryPrefix = [...this.#keyPrefix, "_summaries"] as KvKey;
```

async3619 (approved):

packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,560 @@
...
+   * @throws {Error} If the KvStore supports neither `list()` nor `cas()`.
+   */
+  constructor(kv: KvStore, options?: FedifySpanExporterOptions) {
+    if (kv.list == null && kv.cas == null) {
+      throw new Error(
+        "KvStore must support either list() or cas() method to use " +
+          "FedifySpanExporter. Consider using a KvStore implementation " +
+          "that supports these operations, such as MemoryKvStore, " +
+          "RedisKvStore, PostgresKvStore, or SqliteKvStore.",
+      );
```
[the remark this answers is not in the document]
dahlia: It's fine as is, because `DenoKvStore` and `WorkersKvStore` is only usable on specific runtime platforms.

packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,589 @@
...
+          return;
+        }
+      }
+    }
+
+    // Fallback to non-atomic set if CAS is not available or fails
+    const existing = await this.#kv.get<T>(key);
+    const newValue = transform(existing);
+    await this.#kv.set(key, newValue, options);
+  }
```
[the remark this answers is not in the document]
dahlia: Intentionally keeping the fallback behavior. Since this is a debugging/tracing tool, throwing errors on CAS failures would disrupt the user experience. Losing some trace data occasionally is preferable to crashing the debug dashboard. The data is for debugging purposes, not critical business data.

packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,589 @@
...
+        if (await this.#kv.cas(key, existing, newValue, options)) {
+          return;
+        }
+      }
+    }
+
+    // Fallback to non-atomic set if CAS is not available or fails
+    const existing = await this.#kv.get<T>(key);
+    const newValue = transform(existing);
+    await this.#kv.set(key, newValue, options);
```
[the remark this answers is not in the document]
dahlia: Intentionally keeping the fallback behavior. Since this is a debugging/tracing tool, throwing errors on CAS failures would disrupt the user experience. Losing some trace data occasionally is preferable to crashing the debug dashboard. The data is for debugging purposes, not critical business data.

packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,589 @@
...
+
+    // Sort by timestamp descending (most recent first)
+    summaries.sort((a, b) => b.timestamp.localeCompare(a.timestamp));
+
+    if (options?.limit != null) {
+      return summaries.slice(0, options.limit);
+    }
+
+    return summaries;
+  }
```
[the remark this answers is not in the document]
dahlia: Returning an empty array is intentional for the same reason as above—this is a debugging tool where graceful degradation is preferred over throwing errors. Users can still query specific traces by ID even without `list()` support.

packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,589 @@
...
+  async getRecentTraces(
+    options?: GetRecentTracesOptions,
+  ): Promise<TraceSummary[]> {
+    const summaryPrefix = [...this.#keyPrefix, "_summaries"] as KvKey;
+    const summaries: TraceSummary[] = [];
+
+    if (this.#kv.list != null) {
+      for await (const entry of this.#kv.list(summaryPrefix)) {
+        summaries.push(entry.value as TraceSummary);
+      }
```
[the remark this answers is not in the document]
dahlia: This is a known limitation. Skipping summary creation for CAS-only stores would require additional complexity, and the storage overhead is minimal. We may revisit this in a future iteration if needed.

packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,589 @@
...
+    await this.#setWithCasRetry<TraceActivityRecord[]>(
+      key,
+      (existing) => {
+        const records = existing ?? [];
+        records.push(record);
+        return records;
+      },
+      options,
+    );
+    await this.#updateTraceSummary(record, options);
```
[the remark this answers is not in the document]
dahlia: Consolidating these into a single atomic operation would require significant refactoring. Given that this is trace data for debugging purposes, occasional inconsistencies are acceptable.

packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,591 @@
...
+
+    // Sort by timestamp descending (most recent first)
+    summaries.sort((a, b) => b.timestamp.localeCompare(a.timestamp));
+
+    if (options?.limit != null) {
+      return summaries.slice(0, options.limit);
+    }
+
+    return summaries;
+  }
```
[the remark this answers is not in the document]
dahlia: Returning an empty array is intentional. Since this is a debugging/tracing tool, graceful degradation is preferred over throwing errors—users can still query specific traces by ID using `getActivitiesByTraceId()` even without `list()` support.

However, I agree that documenting this limitation would be helpful. I'll add a note to the documentation mentioning that `getRecentTraces()` requires a `KvStore` with `list()` support.

packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,592 @@
...
+      for (const record of records) {
+        storeOperations.push(this.#storeRecord(record));
+      }
+    }
+    const results = await Promise.allSettled(storeOperations);
+    const rejected = results.filter(
+      (r): r is PromiseRejectedResult => r.status === "rejected",
+    );
+    if (rejected.length > 0) {
+      throw rejected[0].reason;
```
sij411: Just a question/suggestions: what if users need to see all rejected items not just the first one if `rejected` stores multiple items?

packages/fedify/src/otel/exporter.test.ts
```diff
@@ -0,0 +1,881 @@
+import { assertEquals, assertThrows } from "@std/assert";
```
2chanhaeng, line 1: To reduce Fedify's dependency on JSR, wouldn't it be better to use `node:assert` instead of `@std/assert`?

packages/fedify/src/otel/exporter.test.ts
```diff
@@ -0,0 +1,881 @@
...
+    assertEquals(activities[0].spanId, "span456");
+    assertEquals(activities[0].direction, "inbound");
+    assertEquals(activities[0].activityType, "Create");
+    assertEquals(
+      activities[0].activityId,
+      "https://example.com/activities/123",
+    );
+    assertEquals(activities[0].activityJson, activityJson);
+    assertEquals(activities[0].verified, true);
+  });
```
2chanhaeng: It would be better to store the test expected value in a variable instead of hardcoding it.
```suggestion
    const activity = { /* ... */ };
    const activityJson = JSON.stringify(activity);
    const spanOptions = { /* ... */ };
    const span = createMockSpan(spanOptions);

    await new Promise<void>((resolve) => {
      exporter.export([span], (result) => {
        assertEquals(result.code, 0); // SUCCESS
        resolve();
      });
    });

    const activities = await exporter.getActivitiesByTraceId("trace123");
    assertEquals(activities.length, 1);
    assertEquals(activities[0].traceId, spanOptions.traceId);
    assertEquals(activities[0].spanId, spanOptions.spanId);
    assertEquals(activities[0].direction, "inbound");
    assertEquals(activities[0].activityType, activity.type);
    assertEquals(activities[0].activityId, activity.id);
    assertEquals(activities[0].activityJson, activityJson);
    assertEquals(activities[0].verified, true);

```

packages/fedify/src/otel/exporter.ts
dahlia: Good point! I'll use [`AggregateError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError) to include all rejected reasons.

dahlia: Addressed in e2a18075fe4224f3fcbe2ec8fdd73a75d2de675e!

packages/fedify/src/otel/exporter.test.ts
2chanhaeng, line 1: I think it would be better to include an example of how to properly use the second parameter `resultCallback` of `FedifySpanExporter.export`.

packages/fedify/src/otel/exporter.ts
```diff
@@ -0,0 +1,592 @@
...
+    this.#exportAsync(spans)
+      .then(() => resultCallback({ code: ExportResultCode.SUCCESS }))
+      .catch((error) => {
+        getLogger(["fedify", "otel", "exporter"]).error(
+          "Failed to export spans to KvStore: {error}",
+          { error },
+        );
+        resultCallback({ code: ExportResultCode.FAILED });
+      });
+  }
```
2chanhaeng, line 258: I think making `export` an async function is better than passing a callback to it.

packages/fedify/src/otel/exporter.test.ts
dahlia: The project convention uses `@std/assert` across all test files (35+ files). Keeping it consistent with the existing codebase.

packages/fedify/src/otel/exporter.test.ts
dahlia: The test file (*exporter.test.ts*) demonstrates resultCallback usage patterns. Additionally, the main documentation in *opentelemetry.md* shows how to configure the exporter with OpenTelemetry SDK, where the callback handling is managed by the SDK itself. Do you think we need more explicit examples in the documentation?

packages/fedify/src/otel/exporter.ts
dahlia: The callback-based signature is required by OpenTelemetry's `SpanExporter` interface. The interface defines `export(spans, resultCallback)` where the result must be communicated through the callback rather than a return value. This is a constraint from the OpenTelemetry SDK design, not a choice we can change. The internal implementation uses `#exportAsync()` which is async, but the public `export()` method must conform to the interface contract.

packages/fedify/src/otel/exporter.test.ts
dahlia: Addressed in 85d7c09ec2751a0f10bd12409bc0d42cb63a762b!

2chanhaeng: Well, I commented on this test, but how about applying it to all the other tests?

packages/fedify/src/otel/exporter.test.ts
2chanhaeng: Oh, I missed that! 😅

2chanhaeng (approved):

PR #512, opened 2025-12-30
Add relay manual

sij411:
## Summary

Added a manual for `@fedify/relay`. Related to #359 

## Notes

Could you check if this documentation contains all information for building relay servers? I'm open to take opinions or recommendations.

### DIFF
--- a/docs/manual/relay.md
+++ b/docs/manual/relay.md
@@ -5,11 +5,11 @@ description: >-
 ---
 
 Relay
-============
+=====
 
 *This API is available since Fedify 2.0.0.*
 
-Fedify provides the `@fedify/relay` package for building [ActivityPub relay
+Fedify provides the *@fedify/relay* package for building [ActivityPub relay
 servers]—services that forward activities between instances without requiring
 individual actor-following relationships.
 
@@ -19,7 +19,7 @@ individual actor-following relationships.
 Setting up a relay server
 -------------------------
 
-First, install the `@fedify/relay` package.
+First, install the *@fedify/relay* package.
 
 ::: code-group
 
@@ -27,8 +27,16 @@ First, install the `@fedify/relay` package.
 deno add @fedify/relay
 ~~~~
 
-~~~~ sh [Node.js]
-npm add @fedify/relay
+~~~~ sh [npm]
+npm add @fedify/relay @hono/node-server
+~~~~
+
+~~~~ sh [pnpm]
+pnpm add @fedify/relay @hono/node-server
+~~~~
+
+~~~~ sh [Yarn]
+yarn add @fedify/relay @hono/node-server
 ~~~~
 
 ~~~~ sh [Bun]
@@ -39,7 +47,9 @@ bun add @fedify/relay
 
 Then create a relay using the `createRelay()` function.
 
-~~~~ typescript twoslash
+::: code-group
+
+~~~~ typescript twoslash [Deno]
 import { createRelay } from "@fedify/relay";
 import { MemoryKvStore } from "@fedify/fedify";
 
@@ -48,7 +58,7 @@ const relay = createRelay("mastodon", {
   domain: "relay.example.com",
   name: "My ActivityPub Relay",
   subscriptionHandler: async (ctx, actor) => {
-    // Approve all subscriptions 
+    // Approve all subscriptions
     return true;
   },
 });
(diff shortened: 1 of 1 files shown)


docs/manual/relay.md
```diff
@@ -0,0 +1,368 @@
...
+const kv = null as unknown as KvStore;
+// ---cut-before---
+import { isRelayFollower } from "@fedify/relay";
+
+for await (const entry of kv.list(["follower"])) {
+  if (isRelayFollower(entry.value)) {
+    console.log(`Valid follower in state: ${entry.value.state}`);
+  }
+}
+~~~~
```
dahlia: The *Managing followers* section exposes internal implementation details (key patterns like `["follower", actorId]` and the raw `RelayFollower` data structure). This approach has a few concerns:

1. *Encapsulation*: Exposing the internal key schema makes it part of the public API contract
2. *Maintenance burden*: Changing the storage structure later would become a breaking change
3. *Incomplete API*: Direct `KvStore` access isn't a proper API—it lacks type safety and discoverability

I think we have two options:

- *Option A*: Remove this section from the documentation for now, and add it back once we have proper APIs like `relay.listFollowers()`, `relay.getFollower(actorId)`, etc.

- *Option B*: Add these methods to `BaseRelay` first, then document the proper API instead of raw `KvStore` access.

What do you think?

sij411: Option B sounds good, I will be working on that with another PR. (`relay.listFollowers()` and `relay.getFollower(actorId)`) How's that sound?

dahlia: Good!

sij411: Resolved with #514

docs/manual/relay.md
```diff
@@ -0,0 +1,429 @@
...
+
+>>>>>>> cbb51460362eba2c9ccf5d639f9d25899bad64a4
```
dahlia: It seems you'd mistaken this during resolving conflicts?

sij411: oh right, it will be removed!

docs/manual/relay.md
```diff
@@ -0,0 +1,427 @@
...
+ -  `actorId`: The actor's ID (URL) as a string
+ -  `actor`: The validated `Actor` object
+ -  `state`: Either `"pending"` or `"accepted"`
+
+> [!NOTE]
+> The `listFollowers()` method requires a `KvStore` implementation that
+> supports listing by prefix (Redis, PostgreSQL, SQLite, Deno KV all support
+> this).
+
+=======
```
dahlia: You should get rid of this line as well!

docs/manual/relay.md
```diff
@@ -0,0 +1,427 @@
...
+:   Callback to approve or reject subscription requests. See
+    [*Handling subscriptions*](#handling-subscriptions). To create an open relay that accepts all subscriptions:
+
+    ~~~~ typescript
+    subscriptionHandler: async (ctx, actor) => true
+    ~~~~
+
+`documentLoaderFactory`
+:   A factory function for creating a document loader to fetch remote
+      ActivityPub objects. See [*Getting a `Federation`
```
dahlia: This single line is indented incorrectly.

dahlia (changes requested):
Thank you for this comprehensive documentation for `@fedify/relay`! The overall structure is excellent and covers the essential topics well.

However, I found a few issues that need to be addressed before merging:

### Documentation build fails due to Twoslash errors

Running `pnpm build` in the *docs/* directory fails with the following errors:

```
[2307] Cannot find module '@fedify/relay' or its corresponding type declarations.
[7006] Parameter 'ctx' implicitly has an 'any' type.
[7006] Parameter 'actor' implicitly has an 'any' type.
```

The `@fedify/relay` package isn't available in the docs build environment. You'll need to add the package to the docs dependencies. For the implicit `any` errors, you may need to add explicit type annotations in the code examples.

### Markdown style guide issues

There are some places that don't follow the [*Markdown style guide*](https://github.com/fedify-dev/fedify/blob/6cb1cf020716accbcfe3f274b7e01b1aff41ce15/AGENTS.md#markdown-style-guide) described in *AGENTS.md*. Please review the guide and update the formatting accordingly.

---

Once these issues are resolved, the documentation looks great!

docs/manual/relay.md
```diff
@@ -280,46 +280,6 @@ Each follower entry contains:
...
-// ---cut-before---
-import { isRelayFollower } from "@fedify/relay";
-
-for await (const entry of kv.list(["follower"])) {
-  if (isRelayFollower(entry.value)) {
-    console.log(`Valid follower in state: ${entry.value.state}`);
-  }
-}
-~~~~
-
```
dahlia: The *Querying followers* and *Validating follower objects* subsections were removed in this commit, but I believe these examples are valuable for users who need to work with follower data directly through `KvStore`.

Could you please restore these examples? They demonstrate important use cases:

 -  How to iterate over followers using `kv.list<RelayFollower>(["follower"])`
 -  How to validate follower objects using `isRelayFollower()` type guard

If there were Twoslash type errors with these examples, we can work together to fix them rather than removing the examples entirely.

dahlia (changes requested):
There are some places that don't follow the *Markdown style guide* described in *AGENTS.md*. Please review the guide and update the formatting accordingly.

> [!TIP]
> If you're using Claude Code, you can ask it to reformat *docs/manual/relay.md* according to the *Markdown style guide* in *AGENTS.md*—it should handle this well!

docs/manual/relay.md
sij411: Unfortunately, that was the result of asking CC to use AGENTS.md to follow markdown styles. Welp, it seems that didn't work very well though. 

The problem could be solved with adding `isRelayFollower`. However, whendo people use directly querying follower over public APIs?

dahlia: Actually, I take that back. Since there's already a *Listing all followers* subsection using `relay.listFollowers()`, the *Querying followers* subsection would be redundant. And `isRelayFollower()` isn't exported anyway, so removing both subsections was the right call. Sorry for the confusion!

dahlia (dismissed):

PR #514, opened 2025-12-31
Add relay follower query API and refactor types

sij411:
## Summary

Adds public API methods to query relay followers and refactors the type system for better type safety and maintainability.

Related to: https://github.com/fedify-dev/fedify/pull/512#discussion_r2653133284

### New Public API Methods

- `listFollowers(): AsyncIterableIterator<RelayFollower>` - Lists all followers of the relay
- `getFollower(actorId: string): Promise<RelayFollower | null>` - Gets a specific follower by actor ID

### Type System Refactor

**Two-type design for clarity:**

- **`RelayFollowerData`** (internal): Storage format with JSON-LD actor data (`actor: unknown`)
- **`RelayFollower`** (public): API format with validated Actor instances (`actor: Actor`, includes `actorId`)

**Why separate types?**
- KV stores require JSON-serializable data (can't store class instances)
- Deserialization happens at read time
- Public API provides typed Actor objects for better DX

### Implementation Details

- Added `parseFollowerData()` helper to centralize validation and deserialization
- Both API methods use the same validation logic (no duplication)
- Enhanced tests to verify typed Actor properties

## Benefits

- Users can query relay followers without accessing KV store directly  
- Type-safe Actor instances in public API  
- Clear separation between storage and API layers  
- No code duplication in validation logic

### DIFF
--- a/packages/relay/src/base.ts
+++ b/packages/relay/src/base.ts
@@ -71,9 +71,19 @@ export abstract class BaseRelay {
    *
    * @example
    * ```ts
+   * import { createRelay } from "@fedify/relay";
+   * import { MemoryKvStore } from "@fedify/fedify";
+   *
+   * const relay = createRelay("mastodon", {
+   *   kv: new MemoryKvStore(),
+   *   domain: "relay.example.com",
+   *   subscriptionHandler: async (ctx, actor) => true,
+   * });
+   *
    * for await (const follower of relay.listFollowers()) {
    *   console.log(`Follower: ${follower.actorId}`);
    *   console.log(`State: ${follower.state}`);
+   *   console.log(`Actor: ${follower.actor.name}`);
    * }
    * ```
    *
@@ -97,9 +107,21 @@ export abstract class BaseRelay {
    *
    * @example
    * ```ts
-   * const follower = await relay.getFollower("https://mastodon.example.com/users/alice");
+   * import { createRelay } from "@fedify/relay";
+   * import { MemoryKvStore } from "@fedify/fedify";
+   *
+   * const relay = createRelay("mastodon", {
+   *   kv: new MemoryKvStore(),
+   *   domain: "relay.example.com",
+   *   subscriptionHandler: async (ctx, actor) => true,
+   * });
+   *
+   * const follower = await relay.getFollower(
+   *   "https://mastodon.example.com/users/alice"
+   * );
    * if (follower) {
    *   console.log(`State: ${follower.state}`);
+   *   console.log(`Actor: ${follower.actor.preferredUsername}`);
    * }
    * ```
    *
--- a/packages/relay/src/base.ts
+++ b/packages/relay/src/base.ts
@@ -1,8 +1,9 @@
 import type { Federation, FederationBuilder } from "@fedify/fedify";
-import type {
-  RelayFollower,
-  RelayFollowerEntry,
-  RelayOptions,
+import { isActor, Object as APObject } from "@fedify/fedify/vocab";
+import {
+  isRelayFollowerData,
+  type RelayFollower,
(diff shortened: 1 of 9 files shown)


packages/relay/src/mod.ts
dahlia: Unrelated to this specific pull request, but do we need to export this? Apparently it exposes its internal details.

packages/relay/src/base.ts
dahlia: How about having a public interface named `Relay` (with methods like `listFollowers()`, `getFollower()`, and `fetch()`), and keep `BaseRelay` internal?

packages/relay/src/base.ts
```diff
@@ -1,5 +1,10 @@
 import type { Federation, FederationBuilder } from "@fedify/fedify";
-import type { RelayOptions } from "./types.ts";
+import { isActor, Object as APObject } from "@fedify/fedify/vocab";
+import {
+  isRelayFollowerData,
```
dahlia, line 4: When do we need this (as a public API)?

dahlia (changes requested):

packages/relay/src/base.ts
sij411: I used them to validate it's type Actor or not. Doesn't it need to be validated?

packages/relay/src/mod.ts
sij411: Fair enough.

packages/relay/src/base.ts
sij411: Good I'll be working on that

packages/relay/src/base.ts
dahlia: Hmm, users shouldn't be accessible to the raw data which require validation; they should be able to access to only valid data (which are already parsed).

dahlia: Oh, it wasn't *mod.ts*. It's my mistake.

sij411: I'm confused. The function that parse raw data itself is internal. 

```ts
// base.ts
 /**
   * Helper method to parse and validate follower data from storage.
   * Deserializes JSON-LD actor data and validates it.
   *
   * @param actorId The actor ID of the follower
   * @param data Raw data from KV store
   * @returns RelayFollower object if valid, null otherwise
   * @internal
   */
  private async parseFollowerData(
    actorId: string,
    data: unknown,
  ): Promise<RelayFollower | null> {
    if (!isRelayFollowerData(data)) return null;

    const actor = await APObject.fromJsonLd(data.actor);
    if (!isActor(actor)) return null;

    return {
      actorId,
      actor,
      state: data.state,
    };
  }
  ```
  
  And this validator for `RelayFollowerData` which validates when it store data to kv store, also internal.
  
  ```ts
// follow.ts
  /**
 * Type predicate to check if a value is valid RelayFollowerData from KV store.
 * Validates the storage format (JSON-LD), not the deserialized Actor instance.
 *
 * @param value The value to check
 * @returns true if the value is a RelayFollowerData
 * @internal
 */
export function isRelayFollowerData(
  value: unknown,
): value is RelayFollowerData {
  if (!value || typeof value !== "object") return false;
  const obj = value as Record<string, unknown>;
  return (
    "actor" in obj &&
    "state" in obj &&
    typeof obj.state === "string" &&
    (obj.state === "pending" || obj.state === "accepted")
  );
}
```

Eventually, public API only provides valid data as far as I designed. 🤔

dahlia (approved):
Is there any more work to add to this pull request? If not, can we take it out of draft?

packages/relay/src/base.ts
dahlia: Yeah, I was wrong! 😂

PR #517, opened 2026-01-03
Extract vocab from `@fedify/fedify` into `@fedify/vocab`

### LINKED ISSUE #437: Extract vocab from @fedify/fedify into @fedify/vocab

### LINKED ISSUE #519: `@fedify/fixture` Node.js tests fail due to missing `@js-temporal/polyfill` dependency
### Description

Node.js tests in `packages/fixture` fail because the `@js-temporal/polyfill` package cannot be found.

### Steps to reproduce

```bash
deno task test:node
# or
cd packages/fixture && pnpm test
```

### Error message

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@js-temporal/polyfill' 
imported from /home/.../packages/fixture/dist/docloader.test.js
```

### Cause

The built output in `packages/fixture/dist/` imports `@js-temporal/polyfill`, but this package is not listed as a dependency in `packages/fixture/package.json`.

### Solution

Add `@js-temporal/polyfill` to the dependencies in `packages/fixture/package.json`.

2chanhaeng:
## Summary

Extract vocab from `@fedify/fedify` into `@fedify/vocab`.

## Related Issue

- closes #437
- closes #519 (fd8450e355a01c9fc8d778b4730f7660ae811d02)

## Changes

- Almost of code that using Vocabulary API

## Benefits

Now the users who only need the Vocabulary API can install just `@fedify/vocab` and use it without having to install all of `@fedify/fedify`.

## Checklist

- [x] ~~Did you write some relevant docs about this change (if it's a new feature)?~~

## Additional Notes

While working on this issue, `@fedify/fedify/webfinger` also becomes `@fedify/webfinger`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -250,16 +250,6 @@ To be released.
 [#466]: https://github.com/fedify-dev/fedify/issues/466
 [#478]: https://github.com/fedify-dev/fedify/pull/478
 
-### @fedify/fixture
-
- -  Created Fedify test fixture utilities as the *@fedify/fixture* package.
-    This package provides reusable test fixtures for Fedify applications,
-    simplifying the setup of common test scenarios.
-    [[#508], [#516] by ChanHaeng Lee]
-
-[#508]: https://github.com/fedify-dev/fedify/issues/508
-[#516]: https://github.com/fedify-dev/fedify/pull/516
-
 ### @fedify/webfinger
 
  -  Created WebFinger utilities as the *@fedify/webfinger* package.
--- a/docs/manual/webfinger.md
+++ b/docs/manual/webfinger.md
@@ -248,3 +248,130 @@ if (activityPubActorLink?.href) {
 > which automatically performs WebFinger lookups when given a handle.
 > Use `~Context.lookupWebFinger()` when you need the raw WebFinger data or
 > want more direct control over the lookup process.
+
+
+Using WebFinger client
+----------------------
+
+The `@fedify/webfinger` package provides a standalone WebFinger client that can
+be used independently of the main Fedify framework.  This is useful when you
+only need WebFinger lookup functionality without the full ActivityPub
+federation capabilities.
+
+
+### Installation
+
+The `@fedify/webfinger` package is available on [JSR] and [npm].  You can
+install it using the following command:
+
+[JSR]: https://jsr.io/@fedify/webfinger
+[npm]: https://www.npmjs.com/package/@fedify/webfinger
+
+::: code-group
+
+~~~~ bash [Deno]
+deno add jsr:@fedify/webfinger
+~~~~
+
+~~~~ bash [npm]
+npm add @fedify/webfinger
+~~~~
+
+~~~~ bash [pnpm]
+pnpm add @fedify/webfinger
+~~~~
+
+~~~~ bash [Yarn]
+yarn add @fedify/webfinger
+~~~~
+
+~~~~ bash [Bun]
+bun add @fedify/webfinger
+~~~~
+
+:::
+
+
+### Looking up a WebFinger resource
+
+You can look up a WebFinger resource using the `lookupWebFinger()` function:
+
+~~~~ typescript twoslash
+import { lookupWebFinger } from "@fedify/webfinger";
+
+// Look up by acct: URI
+const result = await lookupWebFinger("acct:alice@example.com");
+
+// Look up by URL
+const result2 = await lookupWebFinger("https://example.com/users/alice");
+~~~~
+
+
+### Working with the result
+
+The result is a `ResourceDescriptor` object containing the subject, aliases,
+properties, and links:
+
+~~~~ typescript twoslash
+import { lookupWebFinger } from "@fedify/webfinger";
+
+const result = await lookupWebFinger("acct:alice@example.com");
+if (result != null) {
+  console.log("Subject:", result.subject);
+  console.log("Aliases:", result.aliases);
+
+  // Find the ActivityPub actor URL
+  const actorLink = result.links?.find(
+    (link) => link.rel === "self" && link.type === "application/activity+json",
+  );
+  if (actorLink?.href != null) {
+    console.log("Actor URL:", actorLink.href);
+  }
+}
+~~~~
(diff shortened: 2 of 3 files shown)


packages/vocab-tools/src/class.test.ts
```diff
@@ -88,7 +88,7 @@ if ("Deno" in globalThis) {
 
 async function getEntireCode() {
   const packagesDir = dirname(dirname(import.meta.dirname!));
-  const schemaDir = join(packagesDir, "fedify", "src", "vocab");
+  const schemaDir = join(packagesDir, "vocab");
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in e66567996f00280ddc0ed60a9ed91e1714a1616e

packages/webfinger/src/lookup.test.ts
```diff
@@ -1,7 +1,7 @@
 import { test } from "@fedify/fixture";
-import { assertEquals } from "@std/assert";
 import { withTimeout } from "es-toolkit";
 import fetchMock from "fetch-mock";
+import { deepStrictEqual } from "node:assert";
```
dahlia: Use [strict mode](https://nodejs.org/api/assert.html#strict-assertion-mode) instead:

```suggestion
import { deepStrictEqual } from "node:assert/strict";
```

2chanhaeng: Fixed at a2ef822ea00c76dd844dbdd4730321ad52b65977!

dahlia:
@2chanhaeng I fixed a CI bug on the `next` branch today. Could you please rebase your work on top of the latest `next` branch?

2chanhaeng:
> The `platform` setting in the _tsdown.config.ts_ files for all projects was changed from `"node"` → `"neutral"`. Why was this change necessary?

I'm not sure of the exact reason, but there was a bug where some packages couldn't reference built packages. Claude said converting the platform to "neutral" would resolve the error, and it actually did get fixed. If you tell me to look for another way, I will look for it more.

packages/vocab/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+~~~~ bash [Yarn]
+yarn add @fedify/vocab
+~~~~
+
+~~~~ bash [Bun]
+bun add @fedify/vocab
+~~~~
+
+:::
```
dahlia: `::: code-group` directive is a vendor extension for VitePress. You can't use this syntax here!

By the way, this instruction should be included in the *docs/manual/vocab.md*.

dahlia (changes requested):

packages/vocab/README.md
2chanhaeng: Fixed at fe01ca695b1af42dcab9ee6dd5b97a77dea8ebb1!

CHANGES.md
```diff
@@ -250,6 +250,36 @@ To be released.
...
+### @fedify/fixture
+
+ -  Created Fedify test fixture utilities as the *@fedify/fixture* package.
+    This package provides reusable test fixtures for Fedify applications,
+    simplifying the setup of common test scenarios.
+    [[#508], [#516] by ChanHaeng Lee]
+
+[#508]: https://github.com/fedify-dev/fedify/issues/508
+[#516]: https://github.com/fedify-dev/fedify/pull/516
+
```
dahlia: Since *@fedify/fixture* is not published, users shouldn't be aware of it. Please remove it!

2chanhaeng: Fixed at 9f511e5f30105f43bf83b8cf576a752d7eab996d!

dahlia (approved):
Great job! Thanks!

docs/manual/webfinger.md
```diff
@@ -248,3 +248,130 @@ if (activityPubActorLink?.href) {
 > which automatically performs WebFinger lookups when given a handle.
 > Use `~Context.lookupWebFinger()` when you need the raw WebFinger data or
 > want more direct control over the lookup process.
+
+
+Using WebFinger client
+----------------------
+
+The `@fedify/webfinger` package provides a standalone WebFinger client that can
+be used independently of the main Fedify framework.  This is useful when you
```
sij411, line 257: I know this is very trivial and it may not the problem, but I think there are some double spaces between sentences. Making it one space could be prettier?

sij411: Not only this sentence, but also in general i mean.

dahlia: Actually, having double spaces between sentences is one of my habits. Apparently @2chanhaeng mimicked it.

By the way, it doesn't matter because two or more spaces are normalized when it's rendered in browsers.

2chanhaeng: Yeah, ~~Claude~~ I did it. 😂

sij411 (approved):

Issue #477, opened 2025-11-04
`SqliteMessageQueue` for single-node deployments


dahlia:
## Summary

Add a SQLite-based `MessageQueue` implementation to the `@fedify/sqlite` package to complement the existing `SqliteKvStore`. This would enable complete Fedify stack deployment without requiring Redis or PostgreSQL infrastructure, particularly beneficial for single-node instances.

## Motivation

Currently, `@fedify/sqlite` provides `SqliteKvStore` but no `MessageQueue` implementation, unlike `@fedify/redis` and `@fedify/postgres` which offer both. This creates an asymmetry where users choosing SQLite for key–value storage must still set up additional infrastructure for message queuing.

**Use cases that would benefit:**

- Personal or small community fediverse instances running on single servers
- Development and testing environments
- Self-hosted instances prioritizing simplicity and minimal infrastructure
- Docker/container deployments aiming for single-container operation

### Core features

- ACID-compliant message enqueue/dequeue operations
- Topic-based message routing
- Retry mechanism with configurable attempts
- Message expiration/cleanup policies
- WAL mode for improved concurrency

### Example schema

```sql
CREATE TABLE IF NOT EXISTS fedify_message_queue (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  payload TEXT NOT NULL,
  created INTEGER NOT NULL,
  scheduled INTEGER NOT NULL,
  processed INTEGER,
  retry_count INTEGER DEFAULT 0,
  max_retries INTEGER DEFAULT 3,
  error TEXT
);

CREATE INDEX IF NOT EXISTS idx_scheduled_processed 
  ON fedify_message_queue(scheduled, processed);
```

### API design

The implementation should follow the existing `MessageQueue` interface:

```typescript
export class SqliteMessageQueue implements MessageQueue {
  readonly nativeRetrial = false;
  
  constructor(db: Database, options?: SqliteMessageQueueOptions);
  
  enqueue(
    message: any,
    options?: MessageQueueEnqueueOptions,
  ): Promise<void>;
  
  enqueueMany?(
    messages: any[],
    options?: MessageQueueEnqueueOptions,
  ): Promise<void>;
  
  listen(
    handler: (message: any) => Promise<void> | void,
    options?: MessageQueueListenOptions,
  ): Promise<void>;
}
```

### Explicit limitations

This implementation would be explicitly designed for *single-node deployments only*:

- Cannot be used in multi-node/distributed setups
[...]

PR #526, opened 2026-01-16
Add SqliteMessageQueue and related tests

2chanhaeng:
Summary
-------

Introduce the `SqliteMessageQueue` class along with a testing framework for
message queues. Refactor existing tests to utilize the new `testMessageQueue`
utility for improved test structure and readability.

Related issue
-------------

 -  closes https://github.com/fedify-dev/fedify/issues/477

Changes
-------

 -  Added `SqliteMessageQueue` class to `@fedify/sqlite` package implementing
    the `MessageQueue` interface using SQLite as the backing store
 -  Added `testMessageQueue()` utility function to `@fedify/testing` package
    for standardized testing of `MessageQueue` implementations
 -  Added `waitFor()` and `getRandomKey()` helper functions to `@fedify/testing`
 -  Refactored message queue tests across multiple packages to use
    `testMessageQueue`:
     -  `@fedify/amqp`
     -  `@fedify/denokv`
     -  `@fedify/postgres`
     -  `@fedify/redis`
 -  Added new test file for `SqliteMessageQueue`
 -  Exported new classes and functions from their respective modules

Benefits
--------

 -  Provides a SQLite-based message queue implementation
 -  Reduces code duplication across message queue test files by providing
    a reusable test harness
 -  Ensures consistent test coverage across all `MessageQueue` implementations
 -  Makes it easier to add new `MessageQueue` implementations in the future
    with standardized testing

Checklist
---------

 -  [ ] Did you add a changelog entry to the *CHANGES.md*?
 -  [ ] Did you write some relevant docs about this change (if it's a new feature)?
 -  [ ] Did you write a regression test to reproduce the bug (if it's a bug fix)?
 -  [ ] Did you run `mise test` on your machine?

### DIFF
--- a/packages/sqlite/src/mq.ts
+++ b/packages/sqlite/src/mq.ts
@@ -350,9 +350,13 @@ export class SqliteMessageQueue implements MessageQueue, Disposable {
     while (signal == null || !signal.aborted) {
       let timeout: ReturnType<typeof setTimeout> | undefined;
       await new Promise<unknown>((resolve) => {
-        signal?.addEventListener("abort", resolve);
+        const onAbort = () => {
+          signal?.removeEventListener("abort", onAbort);
+          resolve(undefined);
+        };
+        signal?.addEventListener("abort", onAbort);
         timeout = setTimeout(() => {
-          signal?.removeEventListener("abort", resolve);
+          signal?.removeEventListener("abort", onAbort);
           resolve(0);
         }, this.#pollIntervalMs);
         timeouts.add(timeout);


packages/testing/src/mq-tester.ts
```diff
@@ -0,0 +1,189 @@
...
+        messages.push(message);
+      }, controller);
+      const listening2 = mq2.listen((message: string) => {
+        messages.push(message);
+      }, controller);
+
+      // Test: enqueue()
+      await mq1.enqueue("Hello, world!");
+      await waitFor(() => messages.length > 0, 15_000);
+      deepStrictEqual(messages, ["Hello, world!"]);
```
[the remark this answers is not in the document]
2chanhaeng: Never show me `any` anymore.

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,309 @@
...
+          throw error;
+        }
+
+        // Check for next message immediately
+        continue;
+      }
+
+      // No messages available, wait before polling again
+      await this.#wait(this.#pollIntervalMs, signal);
+    }
```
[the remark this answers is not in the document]
2chanhaeng: The first is applied, but not the second. Because error handling is for the app, not the mq driver.

packages/postgres/src/kv.test.ts
```diff
@@ -13,7 +13,7 @@ if ("Temporal" in globalThis) {
   Temporal = temporal.Temporal;
 }
 
-const dbUrl = process.env.DATABASE_URL;
+const dbUrl = process.env.POSTGRES_URL;
```
sij411, line 16: If this renaming is intentional, could you explain why this is needed?

packages/testing/src/mq-tester.ts
```diff
@@ -0,0 +1,181 @@
...
+  const test = options?.test ?? defaultTest;
+  test(name, async () => {
+    const mq1 = await getMessageQueue();
+    const mq2 = await getMessageQueue();
+    const controller = new AbortController();
+    try {
+      // Initialize if needed (e.g., create database tables)
+      if (options.initialize != null) {
+        await options.initialize(mq1);
+      }
```
sij411: I have some comments of `options.initialize()`:
- If we have to initialise the message queues, why don't we init both `mq1` and `mq2`?
- I think message queues such as `SqliteMessageQueue` and `PostgresMessageQueue` already initialise internally so I don't think it's necessary.
- Last, I couldn't find any tests use this (`options.initialize()`).

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,308 @@
...
+   */
+  constructor(
+    readonly db: PlatformDatabase,
+    readonly options: SqliteMessageQueueOptions = {},
+  ) {
+    this.#db = new SqliteDatabase(db);
+    this.#initialized = options.initialized ?? false;
+    this.#tableName = options.tableName ?? SqliteMessageQueue.#defaultTableName;
+    this.#pollIntervalMs = Temporal.Duration.from(
+      options.pollInterval ?? { milliseconds: 500 },
```
sij411: > /**
> 
> 
>    * The poll interval for the message queue.  5 seconds by default.
>    */
>   pollInterval?: Temporal.Duration | Temporal.DurationLike;
> }

There is a mismatch between in the document(line 31) and the actual default value.

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,308 @@
...
+    logger.debug(
+      "Starting to listen for messages on table {tableName}...",
+      { tableName: this.#tableName },
+    );
+
+    while (signal == null || !signal.aborted) {
+      const now = Temporal.Now.instant().epochMilliseconds;
+
+      // Atomically fetch and delete the oldest message that is ready to be
+      // processed using DELETE ... RETURNING (SQLite >= 3.35.0)
```
sij411: Maybe we could put this in the document as well?

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,308 @@
...
+
+        try {
+          await handler(message);
+          logger.debug("Processed message {id}.", { id: result.id });
+        } catch (error) {
+          logger.error(
+            "Failed to process message {id}: {error}",
+            { id: result.id, error },
+          );
+          throw error;
```
sij411: | Implementation | Handler Error      | Result                            |
  |----------------|--------------------|-----------------------------------|
  | Redis          | Not caught         | Breaks poll(), outer loop retries |
  | Postgres       | Not caught         | Breaks poll(), outer loop retries |
  | Sqlite         | Caught + re-thrown | Terminates entire listen()        |

I found SqliteMessageQueue is inconsistent in error handling. I think it should either:
  1. Not re-throw (like Redis/Postgres) - let outer loop continue
  2. Or document this "fail-fast" behavior explicitly
  
  What do you think?

packages/postgres/src/kv.test.ts
2chanhaeng: The other database URL variable names are matched with their respective database names.(e.g. REDIS_URL for redis) Therefore, to avoid confusion, I've also modified this one to match the naming convention.

packages/testing/src/mq-tester.ts
2chanhaeng: I was going to implement initialize but then decided to remove it and forgot to do so. I'll fix it. Thanks!

packages/sqlite/src/mq.ts
2chanhaeng: I forgot the comment. Thanks!

packages/sqlite/src/mq.ts
2chanhaeng: Oh, that's a good point! I'll fix it.

packages/sqlite/src/mq.ts
2chanhaeng: OK!

packages/testing/src/mq-tester.ts
2chanhaeng: Fixed at [2dc776b](https://github.com/fedify-dev/fedify/pull/526/commits/2dc776b3f1aad0add17bc66d1278930f08e521bb)!

packages/sqlite/src/mq.ts
2chanhaeng: Fixed at https://github.com/fedify-dev/fedify/pull/526/commits/bdb43281d5227169a0ee3ed236ec48926e56ab95!

packages/sqlite/src/mq.ts
2chanhaeng: Fixed at https://github.com/fedify-dev/fedify/pull/526/commits/3110664afe41c67551c00d7881f9d81f08cc236f!

packages/sqlite/src/mq.ts
2chanhaeng: Fixed at https://github.com/fedify-dev/fedify/pull/526/commits/13908193155fe8edef32a49a4304443aba1aacc4!

sij411:
> > How about adding `[Symbol.dispose]()` for consistency with other MessageQueue implementations?
> 
> That's a good idea, but it seems to go beyond #477. How about you create a separate issue for it?

Well, the cleanup method belongs to `sqlite/src/ mq.ts` so i don't think that's out of scope if you implement it for only SqliteMessageQueue. It seems quite simple but caring other message queues should be handled in other PR like you said. (Neither amqp has that)

sij411 (approved):

packages/postgres/src/kv.test.ts
sij411: Aha, and that won't have side-effects, right? It's an environment variable so I just worried about that. If that's fine then I got no worries.

2chanhaeng: Well, I checked again, and then I found more `DATABASE_URL`! Because of you, I could check one more time. Thanks!

Fixed at df0e05a6e65ea0af5226f5071738b19a1fc24a17!

2chanhaeng (commented):
> > > How about adding `[Symbol.dispose]()` for consistency with other MessageQueue implementations?
> > 
> > That's a good idea, but it seems to go beyond #477. How about you create a separate issue for it?
> 
> Well, the cleanup method belongs to `sqlite/src/ mq.ts` so i don't think that's out of scope if you implement it for only SqliteMessageQueue. It seems quite simple but caring other message queues should be handled in other PR like you said. (Neither amqp has that) 

Oh, OK. I added it at 80e11c683a01497d060528a35b3eb2f4c65c685b!

packages/testing/src/mq-tester.ts
```diff
@@ -0,0 +1,169 @@
+import type { MessageQueue } from "@fedify/fedify";
+import { test as defaultTest } from "@fedify/fixture";
```
dahlia: Hmm, it makes `@fedify/testing` (which is public) depend on `@fedify/fixture`, which is not public. I guess we should either:

- Extract `@fedify/testing/mq-tester` into `@fedify/fixture/mq-tester`, or
- Make it depend on `node:test` instead of `@fedify/fixture`.

2chanhaeng: How about removing the default value of `TestMessageQueueOptions.test` and making it non-optional? Then the dependency can be removed.

dahlia: Turns out that's… quite unnecessarily too abstract! How about letting it be used like:

```ts
test("...", () => testMessageQueue());
```

… instead of:

```ts
testMessageQueue("...", { test });
```

which seem an unneeded [IoC](https://en.wikipedia.org/wiki/Inversion_of_control)?

2chanhaeng: Yeah, that will be more useful!

2chanhaeng: Yeah, that will be more useful!

2chanhaeng: Fixed at c180911b8e88babbc6e13c922350bbebaec434cd!

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,361 @@
...
+        const result = this.#db
+          .prepare(
+            `DELETE FROM "${this.#tableName}"
+            WHERE id = (
+              SELECT id FROM "${this.#tableName}"
+              WHERE scheduled <= ?
+              ORDER BY scheduled
+              LIMIT 1
+            )
+            RETURNING id, message`,
```
dahlia: While `DELETE … RETURNING` is atomic, when multiple processes execute the same query simultaneously, there's a potential race condition. Using `BEGIN IMMEDIATE` ensures proper locking and prevents duplicate message processing.

Also, don't forget you need to handle `SQLITE_BUSY` error as well!

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,361 @@
...
+ *   queue: new SqliteMessageQueue(db),
+ * });
+ * ```
+ */
+export class SqliteMessageQueue implements MessageQueue, Disposable {
+  static readonly #defaultTableName = "fedify_message";
+  static readonly #tableNameRegex = /^[A-Za-z_][A-Za-z0-9_]{0,63}$/;
+  // In-memory event emitter for notifying listeners when messages are enqueued.
+  // Scoped per table name to allow multiple queues to coexist.
+  static readonly #notifyChannels = new Map<string, EventTarget>();
```
dahlia, line 92: The static `#notifyChannels` map can accumulate `EventTarget` instances indefinitely. Consider implementing cleanup logic when the last listener for a channel is removed.

This is a minor issue for most use cases, but worth noting for long-running applications with many different table names.

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,361 @@
...
+      this.#db.exec("COMMIT");
+      logger.debug("Enqueued messages.", { messages });
+
+      // Notify listeners that messages have been enqueued
+      const delayMs = delay.total("millisecond");
+      SqliteMessageQueue.#getNotifyChannel(this.#tableName).dispatchEvent(
+        new EnqueueEvent(delayMs),
+      );
+    } catch (error) {
+      this.#db.exec("ROLLBACK");
```
dahlia: Currently, errors are caught and re-thrown without logging. Adding error logs would help with debugging:


```suggestion
      this.#db.exec("ROLLBACK");
      logger.error("Failed to enqueue messages: {error}", { error, messages });
```

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,361 @@
...
+    } else {
+      logger.debug("Enqueuing a message...", { message });
+    }
+
+    this.#db
+      .prepare(
+        `INSERT INTO "${this.#tableName}" (id, message, created, scheduled)
+        VALUES (?, ?, ?, ?)`,
+      )
+      .run(id, encodedMessage, now, scheduled);
```
dahlia: Even with `busy_timeout` set, `SQLITE_BUSY` errors can still occur when the timeout expires. In a multi-process message queue, this is a normal condition that requires retry logic.

Consider also adding retry logic to `enqueueMany()` and `listen()` methods!

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,361 @@
...
+
+  /**
+   * Creates the message queue table if it does not already exist.
+   * Does nothing if the table already exists.
+   */
+  initialize(): void {
+    if (this.#initialized) {
+      return;
+    }
+
```
dahlia, line 385: Without [`journal_mode=WAL`](https://sqlite.org/wal.html), SQLite uses the default rollback journal mode which severely limits concurrent access. This will cause reader/writer conflicts in multi-process environments.

dahlia (changes requested):

packages/sqlite/src/mq.ts
2chanhaeng: Fixed at 7e13eec70359624a461a37c71ae98f85c0fa9679 and 06301ba0785225f5a0179b5cce75872dddd1229e!

packages/sqlite/src/mq.ts
2chanhaeng: Fixed at a2851bfb0bea4ccdddef635ca45defc97eb8cd77!

packages/sqlite/src/mq.ts
2chanhaeng: Fixed at c1697fce8a0e3dfc3f59c453e501d7d06043e868!

packages/sqlite/src/mq.ts
2chanhaeng: Fixed at 06301ba0785225f5a0179b5cce75872dddd1229e!

packages/sqlite/src/mq.ts
2chanhaeng: Fixed at c128708c4bf6967f84950ba421a40d35e68cf77a!

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,575 @@
...
+              WHERE id = (
+                SELECT id FROM "${this.#tableName}"
+                WHERE scheduled <= ?
+                ORDER BY scheduled
+                LIMIT 1
+              )
+              RETURNING id, message`,
+            )
+            .get(now) as { id: string; message: string } | undefined;
+        });
```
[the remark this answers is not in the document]
2chanhaeng: No, the message is NOT re-enqueued.

packages/sqlite/src/mq.ts
```diff
@@ -0,0 +1,575 @@
...
+    try {
+      this.#db.close();
+      this.#unregisterInstance();
+    } catch (error) {
+      logger.error(
+        "Failed to close the database connection for table {tableName}: {error}",
+        { tableName: this.#tableName, error },
+      );
+    }
+  }
```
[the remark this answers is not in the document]
2chanhaeng: Wrong. `unregisterInstance` is inside the try block, so it only runs if `close` succeeds. If `close` throws, execution jumps to `catch` and the instance remains registered.

PR #529, opened 2026-01-21
Add --tunnel-service option to lookup, inbox, and relay commands

### LINKED ISSUE #525: Add `--tunnel-service` option to `fedify lookup` and `fedify inbox` commands
Currently, the `fedify tunnel` command provides a `-s`/`--service` option to choose the tunneling service (localhost.run, serveo.net, or pinggy.io). However, two other commands that use tunneling internally don't have this option.

The `fedify lookup` command uses tunneling when `--authorized-fetch` is specified, and `fedify inbox` uses tunneling by default unless `-T`/`--no-tunnel` is provided. In both cases, users cannot select which tunneling service to use.

## Rationale

This inconsistency is problematic when the default tunneling service is unavailable or has poor connectivity. Users should be able to switch to alternative services just like they can with `fedify tunnel`.

## Proposed change

Add a `--tunnel-service` option to `fedify lookup` and `fedify inbox` commands. For consistency, also add `--tunnel-service` as an alias to the existing `-s`/`--service` option in `fedify tunnel` (keeping both for backward compatibility).

For `fedify lookup`, the option would only apply when `--authorized-fetch` is used. For `fedify inbox`, it would apply when tunneling is enabled (no `-T`/`--no-tunnel` flag).

The new option should only have a long-form (no short abbreviation) since suitable short-form options are already taken (`-t` for `--traverse`, `-S` for `--suppress-errors`). Since tunneling service selection isn't frequently changed, requiring the full option name is acceptable.

sij411:
## Summary

- Adds `--tunnel-service` option to `fedify lookup`, `fedify inbox`, and `fedify relay` commands to select the tunneling service (localhost.run, serveo.net, or pinggy.io)
- Adds `--tunnel-service` as an alias to the existing `-s/--service` option in the `fedify tunnel` command for consistency
- Uses Optique's `or()` combinator to make `--tunnel-service` only available when tunneling is enabled (mutually exclusive with `--no-tunnel`)

Closes #525 

## Test plan

- [x] Run `fedify lookup --help` and verify `--tunnel-service` option appears
- [x] Run `fedify inbox --help` and verify `--tunnel-service` option appears  
- [x] Run `fedify relay --help` and verify `--tunnel-service` option appears
- [x] Run `fedify tunnel --help` and verify `--tunnel-service` alias works
- [x] Test `fedify lookup -a --tunnel-service serveo.net acct:hongminhee@hollo.social` to verify the option works
- [x] Verify that `--tunnel-service` cannot be used together with `--no-tunnel`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -267,6 +267,21 @@ To be released.
 [#437]: https://github.com/fedify-dev/fedify/issues/437
 
 
+Version 1.10.1
+--------------
+
+Released on January 22, 2026.
+
+### @fedify/testing
+
+ -  Fixed `TestContext.getActor()` and `TestContext.getObject()` returning
+    `null` instead of calling registered dispatchers.  The methods now properly
+    invoke actor and object dispatchers when they are registered via
+    `setActorDispatcher()` and `setObjectDispatcher()`.  [[#530]]
+
+[#530]: https://github.com/fedify-dev/fedify/issues/530
+
+
 Version 1.10.0
 --------------
 
@@ -362,6 +377,19 @@ Released on December 24, 2025.
 ### @fedify/cfworkers
 
  -  Implemented `list()` method in `WorkersKvStore`.  [[#498], [#500]]
+
+
+Version 1.9.3
+-------------
+
+Released on January 22, 2026.
+
+### @fedify/testing
+
+ -  Fixed `TestContext.getActor()` and `TestContext.getObject()` returning
+    `null` instead of calling registered dispatchers.  The methods now properly
+    invoke actor and object dispatchers when they are registered via
+    `setActorDispatcher()` and `setObjectDispatcher()`.  [[#530]]
 
 
 Version 1.9.2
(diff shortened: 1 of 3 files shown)


packages/cli/src/inbox.tsx
```diff
@@ -86,14 +91,33 @@ export const inboxCommand = command(
...
+          }),
+          () => false as const,
+        ),
+      }),
+      object({
+        tunnel: constant(true),
+        tunnelService: optional(
+          option(
+            "--tunnel-service",
+            choice(["localhost.run", "serveo.net", "pinggy.io"]),
```
dahlia: You could extract this list into a constant.

packages/cli/src/inbox.tsx
```diff
@@ -86,14 +91,33 @@ export const inboxCommand = command(
...
+            {
+              description: message`The tunneling service to use: ${
+                value("localhost.run")
+              }, ${value("serveo.net")}, or ${value("pinggy.io")}.`,
```
dahlia: Try using [`valueSet()`](https://optique.dev/concepts/messages#value-sets) instead.

dahlia (changes requested):

dahlia (changes requested):
We need to document these new options in *docs/cli.md*.

PR #550, opened 2026-01-31
Fix traverseCollection() for inline CollectionPage without id

moreal:
Summary
-------

`traverseCollection()` used `collection.firstId` to determine whether a collection is paginated.  When a server (e.g. Mastodon) embeds a CollectionPage inline in the first property without an explicit id, `firstId` returns null, causing the function to enter the non-paginated branch and call `collection.getItems()` — which yields nothing because items live on the page, not the collection itself.

Replace the `firstId` check with an actual `getFirst()` call so that inline `CollectionPage` objects are correctly detected and traversed.

Related issue
-------------

I couldn't find related issue when searching with following queries:

- `label:type/bug`
- `traverse`
- `collection`

Changes
-------

 -  Added regression test for inline paged collection.
 -  Let `traverseCollection` call `Collection.getFirst()` instead to check only `Collection.firstId`.

Benefits
--------

`traverseCollection` can handle collections with a `first` property without `id` property. (e.g., Mastodon `Note.replies`)

Checklist
---------

 -  [ ] Did you write some relevant docs about this change (if it's a new feature)?
 -  [x] Did you write a regression test to reproduce the bug (if it's a bug fix)?
 -  [ ] Did you write some tests for this change (if it's a new feature)?
 -  [ ] Did you run `mise test` on your machine?
     - In `1.9-maintenance` branch, there is no `test` mise task. So I run `deno task test`, `pnpm run test`, `pnpm run test:bun`

Additional notes
----------------

Mastodon `Note.replies` example:

```
fedify lookup --raw 'https://social.silicon.moe/@moreal/115990520528477317' | jq '.replies'
✔ Fetched object: https://social.silicon.moe/@moreal/115990520528477317.
{
  "id": "https://social.silicon.moe/users/moreal/statuses/115990520528477317/replies",
  "type": "Collection",
  "first": {
    "type": "CollectionPage",
    "next": "https://social.silicon.moe/users/moreal/statuses/115990520528477317/replies?min_id=115990521217861925&page=true",
    "partOf": "https://social.silicon.moe/users/moreal/statuses/115990520528477317/replies",
    "items": [
      "https://social.silicon.moe/users/moreal/statuses/115990521217861925"
    ]
  }
}
✔ Successfully fetched the object.
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,15 @@ Version 1.9.5
 
 To be released.
 
+### @fedify/fedify
+
+ -  Fixed `traverseCollection()` yielding no items when a `Collection` has
+    an inline `CollectionPage` in its `first` property without an explicit
+    `id`.  This is common in Mastodon's `replies` collections.  The function
+    previously used `collection.firstId` to determine pagination, which
+    returned `null` for inline pages without an `id`, causing it to
+    incorrectly fall into the non-paginated branch.
+
 
 Version 1.9.4
 -------------
--- a/packages/fedify/src/vocab/lookup.ts
+++ b/packages/fedify/src/vocab/lookup.ts
@@ -290,14 +290,14 @@ export async function* traverseCollection(
   collection: Collection,
   options: TraverseCollectionOptions = {},
 ): AsyncIterable<Object | Link> {
-  if (collection.firstId == null) {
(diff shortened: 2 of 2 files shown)


dahlia:
Released with 1.9.5 and 1.10.3.

PR #563, opened 2026-02-10
Resolve issues related with `fedify init`

### LINKED ISSUE #351: [Proposal] `npx create-fedify-app`
### Summary

Implement `npx create-fedify-app`.

### Problem

- Install [`@fedify/cli`](https://github.com/fedify-dev/fedify/blob/main/packages/cli) globally.
- Execute `fedify init`.
- Install [`@fedify/fedify`](https://github.com/fedify-dev/fedify/blob/main/packages/fedify).
- etc...

### Proposed Solution

---

### Alternatives Considered

_No response_

### Scope / Dependencies

_No response_

### LINKED ISSUE #461: Add tests for `init` CLI  and fix the command

### LINKED ISSUE #482: Separate `@fedify/init` from `@fedify/cli`
### Summary

Separate `fedify init` from `@fedify/cli` and create `@fedify/init`. `@fedify/cli` must then re-import this to ensure `fedify init` and `fedify test-init` function correctly. Additionally, ensure it operates independently to simultaneously resolve #351.

### Target Area

- `packages/cli/src/init`
- `packages/init`

### Motivation

`fedify init` already has enough code to warrant being separated into its own package. With the addition of `fedify test-init` in #479, the code has grown even longer. Furthermore, the internal addition of the `--test-mode` argument has raised concerns that it could confuse users. Additionally, considering existing requests like #351, the idea arose to separate `fedify init` from `@fedify/cli` and create a new package called `@fedify/init`.

### Risk Assessment

- Increament of the time for CI/CD by sub-package addition.

### Benefits

- Separation of the concern and reduction the amount of code in `@fedify/cli`.

### LINKED ISSUE #485: Advice about server to users who use `fedify test-init`
### Summary

Advice about server to users who use `fedify test-init`

### Problem

`fedify test-init` is a CLI to be added in #479, enabling contributors creating Fedify integrations to easily test their creations with other combinations. However, certain combinations such as `postgres`, `amqp`, and `redis` require the server to be running to test properly. Should the server not be running, functionality must be added to inform the user that the server is not running and provide instructions on how to install and run it.
While I could add this myself, #479 is currently too delayed and there are numerous other related issues. Therefore, I would prefer to create a sub-issue for this, enabling either myself or another contributor to address it in the future.

### Proposed Solution

Show log about servers. [Redis](https://redis.io/docs/latest/operate/oss_and_stack/install/archive/install-redis/), [Postgres](https://www.postgresql.org/download/), [RabbitMQ](https://www.rabbitmq.com/docs/download) for `amqp`, ...

### Alternatives Considered

_No response_

### Scope / Dependencies

#479

### LINKED ISSUE #513: `fedify init`: Add `@fedify/lint` to generated projects by default
### Summary

Integrate `@fedify/lint` into projects generated by `fedify init` command, so developers can catch common Fedify mistakes at development time rather than runtime.

### Background

The `@fedify/lint` package was recently added in #494 (closes #287). It provides lint rules that detect:

 -  Missing actor properties (id, inbox, outbox, followers, etc.)
 -  Incorrect Context method usage (e.g., using `new URL()` instead of `ctx.getActorUri()`)
 -  Missing collection filtering implementation

Currently, `fedify init` does not include this package, so new projects miss out on these helpful validations.

#### Deno Projects

Add `@fedify/lint` as a Deno Lint plugin in *deno.json*:

~~~~ json
{
  "lint": {
    "plugins": ["jsr:@fedify/lint"]
  }
}
~~~~

#### Node.js/Bun Projects

1. **Add ESLint with `@fedify/lint`**:

   Create *eslint.config.js*:

   ~~~~ javascript
   import fedifyLint from "@fedify/lint";

   export default fedifyLint;
   ~~~~

2. **Add devDependencies**:

   ~~~~ json
   {
     "devDependencies": {
       "eslint": "^9.0.0",
       "@fedify/lint": "^2.0.0"
     }
   }
   ~~~~

3. **Keep Biome as formatter only**:

   Update *biome.json*:

   ~~~~ json
   {
     "linter": { "enabled": false },
     "formatter": { "enabled": true }
   }
   ~~~~

4. **Add lint script to *package.json***:

   ~~~~ json
   {
     "scripts": {
       "lint": "eslint ."
     }
   }
   ~~~~

5. **Update VS Code extensions**:

   Add ESLint extension recommendation in *.vscode/extensions.json*:

   ~~~~ json
   {
     "recommendations": [
       "dbaeumer.vscode-eslint"
     ]
   }
   ~~~~

### Implementation Checklist

 -  [ ] Update *packages/cli/src/init/json/* with new config templates
 -  [ ] Modify `getDevDependencies()` in *packages/cli/src/init/action/deps.ts*
 -  [ ] Update `getJsons()` in *packages/cli/src/init/action/patch.ts*
 -  [ ] Add *eslint.config.js* template
 -  [ ] Update VS Code extension recommendations
 -  [ ] Update *biome.json* to disable linter
 -  [ ] Add `lint` script to package.json template
 -  [ ] Test with all framework combinations (Hono, Express, Nitro, Next.js, Elysia)
 -  [ ] Update documentation if needed

### Related

 -  #287 (original lint feature request)
 -  #494 (PR that added `@fedify/lint`)

### LINKED ISSUE #533: Consolidate issues related to `fedify init` and `test-init`
### Summary

Consolidate issues related to `fedify init` and `test-init` like followings : 
- #351 
- #461 
- #482 
- #485
- #513

### Problem

Handling issues related to `fedify init` and `test-init` separately would distribute responsibility and increase the likelihood of merge conflicts.

### Proposed Solution

Consolidating them as a single issue and PR.

### Alternatives Considered

_No response_

### Scope / Dependencies

- `packages/cli/src/init`

2chanhaeng:
## Summary

Resolve issues related with `fedify init`. This includes:

- #351
- #461
- #482
- #485
- #513

## Related Issue

- closes #351
- closes #461
- closes #482
- closes #485
- closes #513
- closes #533

## Changes

There have been several accumulated issues related to `fedify init`. To avoid potential conflicts from separate pull requests addressing more than one issue, we have consolidated and resolved them in a single issue, #533.

- Separated `@fedify/init` from `@fedify/cli`.
- Created `create-fedify-app` using `@fedify/init`.
- Added `@fedify/lint` and its related configurations and files when creating a project with `@fedify/init`.
- When testing `@fedify/lint` with `test-init`, it now notifies the user if related local database servers like `Redis` are not running.

## Benefits

- Separating `@fedify/init` allows for the creation of additional packages or apps, such as `create-fedify-app`.
- Users can now create a Fedify app directly via `create-fedify-app` without needing to install `@fedify/cli`.
- Fedify apps created with `fedify init` and `create-fedify-app` are automatically linted, improving user convenience.
- `test-init` now helps with error resolution by notifying the user if local database servers are not running.

## Checklist

## Additional Notes

### DIFF
--- a/packages/init/src/command.ts
+++ b/packages/init/src/command.ts
@@ -122,7 +122,7 @@ export const testInitCommand = command(
     optional(or(noHydRun, noDryRun)),
   ),
   {
-    brief: message`Test an initializing command .`,
+    brief: message`Test an initializing command.`,
     description: message`Test an initializing command on temporary directories.
 
 Unless you specify all options (${optionNames(["-w", "--web-framework"])}, ${


packages/init/src/test/db.ts
```diff
@@ -0,0 +1,67 @@
...
+    });
+    return true;
+  } catch (error: unknown) {
+    if (error instanceof TypeError) {
+      const msg = error.message.toLowerCase();
+      if (msg.includes("refused") || msg.includes("econnrefused")) {
+        return false;
+      }
+      return true;
+    }
```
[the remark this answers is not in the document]
2chanhaeng: It's local-only test.

packages/init/src/globals.ts
sij411, line 1: Just wondering, there is a debug option internally used in `@fedify/cli`. Should we keep it separately? since it's in different package, that would work too.

packages/create-fedify-app/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "create-fedify-app",
```
dahlia: How about renaming it to `@fedify/create` so that users can use `npm init @fedify`?

packages/init/src/utils.ts
```diff
@@ -0,0 +1,189 @@
+import { isObject } from "@fxts/core";
+import { message } from "@optique/core";
+import { print, printError } from "@optique/run";
+import { Chalk } from "chalk";
+import { flow, toMerged } from "es-toolkit";
+import { spawn } from "node:child_process";
+import process from "node:process";
+
+export const colorEnabled: boolean = process.stdout.isTTY &&
+  !("NO_COLOR" in process.env && process.env.NO_COLOR !== "");
```
sij411, line 10:
```suggestion
  !("NO_COLOR" in process.env);
```
With original line, `NO_COLOR = ""` might disable the color.

packages/init/src/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+  (source: Parameters<typeof toMerged>[1] = {}) =>
+  (target: Parameters<typeof toMerged>[0] = {}) => toMerged(target, source);
+
+export const replace = (
+  pattern: string | RegExp,
+  replacement: string | ((substring: string, ...args: unknown[]) => string),
+) =>
+(text: string): string => text.replace(pattern, replacement as string);
+
+export const replaceAll = (
```
sij411, line 52: This isn't being used in codes

packages/create-fedify-app/package.json
2chanhaeng: Fixed in 0394a9549c68de3611d0a90a0b2db086f7b1d16c!

packages/init/src/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+  if (!head) yield [] as ItersItems<T>;
+  else {
+    for (const x of head) {
+      for (const xs of product(...tail)) yield [x, ...xs] as ItersItems<T>;
+    }
+  }
+}
+
+export type GeneratedType<T extends Generator> = T extends
+  Generator<unknown, infer R, unknown> ? R : never;
```
sij411: This looks like it extracts the yield type, but it actually extracts any due to Generator's default type parameters. Might want to double-check if that's intentional currently all the type safety here is coming from manual as casts, not the utility. Also It won't be the problem in runtime since it doesn't use stricter compiler. Just a bit worrying.

packages/init/src/utils.ts
sij411: line 12 seems not used in code too?

packages/init/src/test/db.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (error instanceof DOMException && error.name === "TimeoutError") {
+      return false;
+    }
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
dahlia: I don't think we can use any Deno API here. Instead, we should use the `node:net` module.

packages/init/src/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+  pattern: string | RegExp,
+  replacement: string | ((substring: string, ...args: unknown[]) => string),
+) =>
+(text: string): string => text.replace(pattern, replacement as string);
+
+export const replaceAll = (
+  pattern: string | RegExp,
+  replacement: string | ((substring: string, ...args: unknown[]) => string),
+) =>
+(text: string): string => text.replaceAll(pattern, replacement as string);
```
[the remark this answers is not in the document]
2chanhaeng: It's not my fault. Blame `tsc`.

packages/create/src/mod.ts
dahlia, line 1: I think we need a shebang line here to make this script executable?

packages/cli/package.json
dahlia: Shouldn't this line be removed now?

dahlia (changes requested):

packages/init/src/test/db.ts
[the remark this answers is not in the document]
sij411: The `isPortOpen` with Deno fetch doesn't work as expected with node, but Node API does.

packages/create/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "https://github.com/sponsors/dahlia"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/fedify-dev/fedify.git",
+    "directory": "packages/create"
+  },
+  "type": "module",
+  "bin": {
+    "@fedify/create": "./dist/mod.js"
```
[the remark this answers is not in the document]
2chanhaeng: [The NPM docs](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#bin) said `bin` attr is **path** so it doesn't matter.

packages/cli/deno.json
```diff
@@ -58,27 +62,8 @@
       "dependencies": [
         "codegen"
       ]
-    },
-    "test-init": {
-      "command": "FEDIFY_TEST_MODE=true deno run --allow-all src/init/test/mod.ts test-init",
-      "dependencies": [
-        "codegen"
-      ]
     }
```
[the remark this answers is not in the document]
2chanhaeng: This moved to `@fedify/init` so no needed.

packages/create/src/mod.ts
2chanhaeng: Fixed at accba8ee538fa00c4dab8ac62bc0a4bd2a9523a1!

packages/cli/package.json
2chanhaeng: Fixed at 0f81f63f0b57788b4d696ba46cf0ba70a8923413!

packages/init/src/test/db.ts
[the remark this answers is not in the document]
2chanhaeng: Fixed at 19f29825714ae5a1c4013d86f19bb4c81e83304f!

packages/init/src/utils.ts
2chanhaeng: `replaceAll` is being used in some files, such as `actions/patch.ts`. `colors` is being used in `actions/notice.ts`.

packages/init/src/globals.ts
2chanhaeng: Well, I've added that option since other CLI tools have it, though it isn't actually being used internally at the moment. Maybe it'll come in handy someday?
Btw, I moved it to `src/command.ts`.

packages/init/src/utils.ts
2chanhaeng: Well, empty string is falsy. So I think it's OK?

packages/init/src/utils.ts
sij411: ohhh okay, vsc's reference couldn't find them

packages/init/src/utils.ts
sij411: OH sorry I misread it nvm.

packages/init/src/utils.ts
2chanhaeng: Fixed at abfdbf0f42723de440c9c51e59a1ac3422a648a7!
It was just a wrong code. How did I not notice this problem until now? 😂

packages/create/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "homepage": "https://fedify.dev/",
+  "bugs": {
+    "url": "https://github.com/fedify-dev/fedify/issues"
+  },
+  "license": "MIT",
+  "author": {
+    "name": "Hong Minhee",
+    "email": "hong@minhee.org",
+    "url": "https://hongminhee.org/"
+  },
```
dahlia, line 21: Since `@fedify/create` is a new package created by you in this PR, the `author` field should credit you rather than me. Could you update it to your information?

The same applies to *packages/init/package.json* as well.

packages/init/deno.json
```diff
@@ -0,0 +1,51 @@
...
+    ],
+    "rules": {
+      "exclude": [
+        "no-slow-types"
+      ]
+    }
+  },
+  "test": {
+    "exclude": [
+      "src/init/test/**"
```
dahlia: I think this `exclude` pattern should be updated?

packages/create/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+The tool supports the same project configurations as `fedify init`:
+
+ -  **Web frameworks**: [Hono], [Nitro], [Next.js], [Elysia], [Express]
+ -  **Package managers**: Deno, pnpm, Bun, Yarn, npm
+ -  **Key-value stores**: Deno KV, Redis, PostgreSQL
+ -  **Message queues**: Deno KV, Redis, PostgreSQL, AMQP
+
+See the [`@fedify/init`] package or the [Fedify CLI docs] for details on
+available options (`-r`, `-p`, `-w`, `-k`, `-q`, `--dry-run`).
```
dahlia: What are `-r` and `-q`?

packages/init/src/command.ts
dahlia: There's an unnecessary space right after the period.

packages/init/package.json
dahlia, line 1: It would be great if it has its `engines` field.

packages/init/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "dist/",
+    "package.json",
+    "README.md"
+  ],
+  "dependencies": {
+    "@fxts/core": "catalog:",
+    "@inquirer/prompts": "^7.8.4",
+    "@logtape/logtape": "catalog:",
+    "@optique/core": "^0.9.0",
+    "@optique/run": "^0.9.0",
```
dahlia: Since `@optique/core` and `@optique/run` are now used across three packages (`@fedify/cli`, `@fedify/init`, and `@fedify/create`), it would be nice to add them to the pnpm catalog in *pnpm-workspace.yaml* and use `"catalog:"` in each *package.json*—consistent with how `es-toolkit`, `@logtape/logtape`, etc. are already managed.

packages/init/deno.json
```diff
@@ -0,0 +1,51 @@
...
+  "version": "2.0.0",
+  "license": "MIT",
+  "exports": "./src/mod.ts",
+  "imports": {
+    "@fedify/init/": "./src/",
+    "@fxts/core": "npm:@fxts/core@^1.21.1",
+    "@inquirer/prompts": "npm:@inquirer/prompts@^7.8.4",
+    "@logtape/logtape": "jsr:@logtape/logtape@^2.0.0",
+    "@optique/core": "jsr:@optique/core@^0.9.0",
+    "@optique/run": "jsr:@optique/run@^0.9.0",
```
dahlia: Similarly on the Deno side, `@optique/core` and `@optique/run` in *packages/init/deno.json* `imports` could be moved to the root *deno.json* `imports` so they are managed at the workspace level—same as `@fxts/core`, `@logtape/logtape`, `es-toolkit`, etc. are already.

packages/create/README.md
2chanhaeng: Ah, I I got confused for it. `-q` was message Queue(so `-m` is correct) and `-r` was ... I can't even remember what I was mixing it up with. 😂 Anyway, I fixed in 401dc93a2332d48a907298aa1d6e89ae27d2a9cf!

packages/create/package.json
2chanhaeng: Fixed in 4dcefaf54ba9f3f1ecf8e3c157a54a1a92610c7d!

packages/init/package.json
2chanhaeng: Fixed in 1a27f182c5f88b894cdf8c6929287713519337bd!

packages/init/deno.json
2chanhaeng: Fixed in 1a27f182c5f88b894cdf8c6929287713519337bd!
Also, since `@fedify/cli` and `@fedify/init` use `chalk`, so I moved it to the workspace.

packages/init/deno.json
2chanhaeng: Fixed in 384401bdbf691cc782b1dbceb1e44a0381beed03!

packages/init/package.json
2chanhaeng: Fixed in 5987ab14da272d6c5edff640951a8662ff135097!

packages/init/src/command.ts
2chanhaeng: Fixed at b0ec83ddfb66034809325fb251c38cc86276dfed!

packages/init/src/utils.ts
sij411: One guess is type casting hides the problems.

PR #566, opened 2026-02-12
Add configuration file support for CLI

### LINKED ISSUE #265: Add configuration file support to Fedify CLI

sij411:
## Summary

 - Add unified configuration file support for Fedify CLI using Optique and Valibot
  - Support hierarchical config loading with TOML format from multiple levels:
    - System-wide: `/etc/fedify/config.toml`
    - User-level: `~/.config/fedify/config.toml` (or $XDG_CONFIG_HOME/fedify/config.toml)
    - Project-level: `.fedify.toml` in current directory
    - Custom path via `--config` (highest precedence)
  - Configuration files are deep-merged (later files override earlier ones)
  - Add `--ignore-config` option to skip all config file loading
  - Bind all command options (`inbox`, `lookup`, `webfinger`, `nodeinfo`, `tunnel`, `relay`) with config
  values
  - Remove `globals.ts` and relocate its contents to appropriate modules
  
Closes #265

## Test plan

- [x] Verify config file loading from default paths
- [x] Verify `--config` option loads custom config file
- [x] Verify `--ignore-config` skips config loading
- [x] Verify command options are properly bound with config values
- [x] Run existing CLI tests


packages/cli/src/relay.ts
```diff
@@ -36,25 +37,38 @@ export const relayCommand = command(
...
+        optional(
+          option("--persistent", string({ metavar: "PATH" }), {
+            description:
+              message`Path to SQLite database file for persistent storage. If not specified, uses in-memory storage which is lost when the server stops.`,
+          }),
+        ),
+        {
+          context: configContext,
+          key: (config) => config.relay?.persistent,
+        },
```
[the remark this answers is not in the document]
sij411: It will return `string | undefined`.

packages/cli/src/mod.ts
```diff
@@ -9,25 +16,51 @@ import { inboxCommand, runInbox } from "./inbox.tsx";
...
+
+      const userConfigDir = envPaths("fedify", { suffix: "" }).config;
+      const system = tryLoadToml("/etc/fedify/config.toml");
+      const user = tryLoadToml(join(userConfigDir, "config.toml"));
+      const project = tryLoadToml(join(process.cwd(), ".fedify.toml"));
+
+      // Custom config via --config throws on error (required file)
+      const custom = parsed.configPath
+        ? parseToml(readFileSync(parsed.configPath, "utf-8"))
+        : {};
```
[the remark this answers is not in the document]
sij411: Done

packages/cli/src/mod.ts
```diff
@@ -9,25 +16,51 @@ import { inboxCommand, runInbox } from "./inbox.tsx";
...
+  const result = await runWithConfig(command, configContext, {
     programName: "fedify",
-    help: "both",
+    load: (parsed) => {
+      if (parsed.ignoreConfig) return {};
+
+      const userConfigDir = envPaths("fedify", { suffix: "" }).config;
+      const system = tryLoadToml("/etc/fedify/config.toml");
+      const user = tryLoadToml(join(userConfigDir, "config.toml"));
+      const project = tryLoadToml(join(process.cwd(), ".fedify.toml"));
```
dahlia: Two concerns about the system-wide config path hard-coded here:

### */etc/fedify/config.toml* is not XDG-compliant and not portable across installations

The user-level config on line 44 already follows platform conventions via `env-paths`, but the system-wide config path is hard-coded to */etc/fedify/config.toml*. According to the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/), system-wide config should be looked up from `XDG_CONFIG_DIRS` (defaulting to */etc/xdg*). This would also handle non-standard prefix installations more naturally (e.g., Homebrew on macOS, where a user might set `XDG_CONFIG_DIRS` to include */opt/homebrew/etc/xdg*).

Consider replacing the hard-coded path with something like:

```typescript
const xdgConfigDirs = (process.env.XDG_CONFIG_DIRS || "/etc/xdg")
  .split(":")
  .map((dir) => join(dir, "fedify", "config.toml"));
const system = xdgConfigDirs.reduce(
  (acc, path) => deepMerge(acc, tryLoadToml(path)),
  {},
);
```

Note that `XDG_CONFIG_DIRS` can contain multiple colon-separated paths, each of which should be searched. Earlier entries have higher precedence.

### This will fail on Windows

*/etc/fedify/config.toml* doesn't exist on Windows. The conventional location for system-wide application configuration on Windows is `ProgramData` (typically *C:\ProgramData*). Since `env-paths` doesn't provide a system-wide path, you'll need platform-specific logic here. For example:

```typescript
function getSystemConfigPaths(): string[] {
  if (process.platform === "win32") {
    const programData = process.env.ProgramData || "C:\\ProgramData";
    return [join(programData, "fedify", "config.toml")];
  }
  return (process.env.XDG_CONFIG_DIRS || "/etc/xdg")
    .split(":")
    .map((dir) => join(dir, "fedify", "config.toml"));
}
```

(On macOS, `XDG_CONFIG_DIRS` is not a native convention, but since we're already using `env-paths` for the user-level path, and `env-paths` handles macOS via *~/Library/Preferences*, following XDG for the system-level path on macOS seems like a reasonable simplification—the system-wide config is rarely used on macOS anyway.)

packages/cli/deno.json
```diff
@@ -8,19 +8,23 @@
...
+    "es-toolkit": "npm:es-toolkit@^1.39.4",
     "fetch-mock": "npm:fetch-mock@^12.5.4",
     "hono": "jsr:@hono/hono@^4.8.3",
     "icojs": "npm:icojs@^0.19.5",
     "inquirer-toggle": "npm:inquirer-toggle@^1.0.1",
     "ora": "npm:ora@^8.2.0",
     "shiki": "npm:shiki@^1.6.4",
+    "smol-toml": "npm:smol-toml@^1.6.0",
     "srvx": "npm:srvx@^0.8.7",
+    "valibot": "npm:valibot@^1.2.0",
```
dahlia: Valibot is [published on JSR](https://jsr.io/@valibot/valibot) as `@valibot/valibot`. Per the project's dependency convention, *deno.json* should use the JSR specifier rather than the npm one. Since the JSR name (`@valibot/valibot`) differs from the npm name (`valibot`), an alias is needed to keep the import path consistent—same pattern used for Hono (`"hono": "jsr:@hono/hono@^4.8.3"`):

```suggestion
    "valibot": "jsr:@valibot/valibot@^1.2.0",
```

The *package.json* entry (`"valibot": "^1.2.0"`) is fine as-is.

packages/cli/deno.json
```diff
@@ -8,19 +8,23 @@
...
-    "@optique/core": "jsr:@optique/core@^0.9.0",
-    "@optique/run": "jsr:@optique/run@^0.9.0",
+    "@optique/config": "jsr:@optique/config@^0.10.0-dev.355",
+    "@optique/core": "jsr:@optique/core@^0.10.0-dev.355",
+    "@optique/run": "jsr:@optique/run@^0.10.0-dev.355",
     "@poppanator/http-constants": "npm:@poppanator/http-constants@^1.1.1",
     "chalk": "npm:chalk@^5.6.2",
     "cli-table3": "npm:cli-table3@^0.6.5",
     "env-paths": "npm:env-paths@^3.0.0",
+    "es-toolkit": "npm:es-toolkit@^1.39.4",
```
dahlia: `es-toolkit` is already declared in the workspace root *deno.json* (`^1.43.0`), so this package-level entry (`^1.39.4`) is redundant and can be removed. Deno workspaces inherit `imports` from the root, so this line just shadows the root mapping with an older version range.

packages/cli/src/options.ts
```diff
@@ -39,27 +43,105 @@ export const tunnelServiceOption = optional(
...
+    }),
+    object({
+      ignoreConfig: constant(false),
+      configPath: option("--config", string({ metavar: "PATH" }), {
+        description: message`Load an additional configuration file.`,
+      }),
+    }),
+  ),
+  { ignoreConfig: false, configPath: undefined } as const,
 );
```
dahlia: `debugOption` is grouped under `"Global options"` (line 88), but `configOption` (lines 129–147) has no group label, so they'll appear as separate sections in `--help` output.

Since both are global options that apply to every command, consider merging them into a single `globalOptions` using Optique's `merge()`:

```typescript
export const globalOptions = merge(
  "Global options",
  object({
    debug: bindConfig(
      option("-d", "--debug", {
        description: message`Enable debug mode.`,
      }),
      {
        context: configContext,
        key: (config) => config.debug ?? false,
        default: false,
      },
    ),
  }),
  withDefault(
    or(
      object({
        ignoreConfig: map(
          flag("--ignore-config", {
            description: message`Ignore all configuration files.`,
          }),
          () => true as const,
        ),
      }),
      object({
        ignoreConfig: constant(false),
        configPath: option("--config", string({ metavar: "PATH" }), {
          description: message`Load an additional configuration file.`,
        }),
      }),
    ),
    { ignoreConfig: false, configPath: undefined } as const,
  ),
);
```

This way all global options appear under one `"Global options"` section in `--help`, and you only need a single export instead of two.

docs/cli.md
```diff
@@ -102,6 +102,104 @@ the appropriate executable for your platform and put it in your `PATH`.
...
+1.  `/etc/fedify/config.toml` (system-wide)
+2.  *&lt;user-config-dir&gt;/fedify/config.toml* (user-level)
+3.  `.fedify.toml` in the current working directory (project-level)
+4.  Custom path via `--config` option (explicit)
+
+The user config directory varies by operating system:
+
+ -  **Linux**: *~/.config/fedify/config.toml*
+ -  **macOS**: *~/Library/Application Support/fedify/config.toml*
+ -  **Windows**: *%APPDATA%\\fedify\\config.toml*
```
dahlia: Several issues with this section:

### Formatting inconsistency

The four list items use inconsistent formatting:

- Item 1 uses backticks (`` `/etc/fedify/config.toml` ``)
- Item 2 uses asterisks/italics (`*<user-config-dir>/fedify/config.toml*`)
- Item 3 uses backticks for just the filename (`` `.fedify.toml` in the current working directory``)
- Item 4 has no path at all, just a description

Per the project's Markdown conventions, file paths should be wrapped in asterisks. The level of detail also varies—items 1–2 show full paths, item 3 shows only the filename, and item 4 shows no path.

### Incorrect platform-specific paths

The OS-specific paths listed below don't match what `env-paths("fedify", { suffix: "" }).config` actually returns:

| OS | Documented path | Actual `env-paths` result |
|----|----------------|--------------------------|
| macOS | *~/Library/Application Support/fedify/config.toml* | *~/Library/Preferences/fedify/config.toml* |
| Windows | *%APPDATA%\\fedify\\config.toml* | *%APPDATA%\\fedify\\Config\\config.toml* |

### Missing `XDG_CONFIG_HOME` mention

The issue description (#265) correctly states `$XDG_CONFIG_HOME/fedify/config.toml`, and `env-paths` does respect `XDG_CONFIG_HOME` on Linux, but this documentation makes no mention of the environment variable at all. Users who have customized `XDG_CONFIG_HOME` wouldn't know where to put their config file.

### Consider using XDG directly on macOS instead of `env-paths`

`env-paths` returns `~/Library/Preferences` on macOS, but this is a poor fit for CLI tools. Apple's own [macOS Standard Directories documentation](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html) describes `~/Library/Preferences` and `~/Library/Application Support` as being for *apps* (GUI applications with bundle identifiers), not command-line tools. In practice, CLI tools on macOS—including Apple's own `bash`, `git`, `vim`, and `zsh`—use `~/.config` following the XDG convention. Dotfile managers (chezmoi, dotbot, yadm, etc.) also assume `~/.config` on macOS. See [this blog post](https://becca.ooo/blog/macos-dotfiles/) for a thorough analysis.

I'd suggest dropping the `env-paths` dependency and implementing XDG directly:

- **Linux/macOS**: `$XDG_CONFIG_HOME/fedify/config.toml` (default: *~/.config/fedify/config.toml*)
- **Windows**: *%APPDATA%\\fedify\\config.toml*

This gives macOS users a predictable, conventional path (`~/.config/fedify/config.toml`) and removes a dependency. It also makes the documentation simpler—the path is the same on Linux and macOS.

dahlia (changes requested):

packages/cli/src/mod.ts
```diff
@@ -9,25 +17,96 @@ import { inboxCommand, runInbox } from "./inbox.tsx";
...
+      return [system, user, project, custom].reduce(
+        (acc, config) => deepMerge(acc, config),
+        {},
+      );
+    },
+    args: process.argv.slice(2),
+    help: {
+      mode: "both",
+      onShow: () => process.exit(0),
+    },
```
dahlia, line 109: You need to configure the `onError` option as well:

```suggestion
    },
    onError: process.exit.bind(process),
```

packages/cli/package.json
```diff
@@ -45,7 +73,6 @@
   },
   "dependencies": {
     "@fedify/fedify": "workspace:*",
```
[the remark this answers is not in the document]
sij411: `@fedify/init` is now an indenpendent package.

packages/cli/src/mod.ts
```diff
@@ -9,25 +17,96 @@ import { inboxCommand, runInbox } from "./inbox.tsx";
...
+            }`,
+          );
+          process.exit(1);
+        }
+      }
+
+      return [system, user, project, custom].reduce(
+        (acc, config) => deepMerge(acc, config),
+        {},
+      );
```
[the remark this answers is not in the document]
sij411: Configuration validations will be handled by Optique.

packages/cli/src/options.ts
```diff
@@ -39,27 +45,123 @@ export const tunnelServiceOption = optional(
...
+      ),
+    }),
+    object({
+      ignoreConfig: constant(false),
+      configPath: option("--config", string({ metavar: "PATH" }), {
+        description: message`Load an additional configuration file.`,
+      }),
+    }),
+  ),
+  { ignoreConfig: false, configPath: undefined } as const,
```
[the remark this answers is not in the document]
sij411: `or` already does.

packages/cli/src/mod.ts
sij411: Done with `53a2390`

dahlia (approved):

Issue #578, opened 2026-02-20
Add vocabulary for proposal, intent and measure


scammo:
### Summary

I would like to develop federated solutions that allow individuals and organisations to publicly list their offers, which marketplaces can then aggregate. This is inspired by [FEP-0837: Federated Marketplace; Status: Draft](https://codeberg.org/fediverse/fep/src/branch/main/fep/0837/fep-0837.md) and the needs of different organisations and individuals. This is not the full implementation of FEP-0837, just the first step. 

### Problem

Currently, there is no vocabulary included for creating offers in Fedify, and no way to add custom vocabulary. 

### Proposed Solution

Adding the vocabulary for proposal, intent and measure based on the draft Version of FEP-0837. 

### Alternatives Considered

It would also suit me if it were possible to add custom vocabulary to Fedify projects like outlined in #207 

### Scope / Dependencies

Adding the YAML files for proposal, intent and measure, and adding the according vocab-tools snapshots.

PR #579, opened 2026-02-21
Add vocabulary relating to proposals, intentions and measures. 

scammo:
Summary
-------

Add vocabulary relating to proposals, intentions and measures [according to the docs](https://fedify.dev/manual/vocab). 

Related issue
-------------

 -  enhances #578 

Changes
-------

 -  Add vocabulary: "proposals", "intentions" and "measures". 

Benefits
--------

Users can now build a federated system based on the first part of FEP-0837: Federated Marketplace (https://codeberg.org/fediverse/fep/src/branch/main/fep/0837/fep-0837.md). This is really interesting for making offered goods or services available for discovery in the Fediverse. 

Checklist
---------

 -  [x] Did you write some relevant docs about this change (if it's a new feature)?

Additional notes
----------------

Please note that this is both my first pull request and my first enhancement issue for this project. While I have extensive Mastodon/event experience, I have no prior experience of enhancing federation tooling directly. I am happy to make any necessary changes or additions. Thank you for your work!

### DIFF
--- a/packages/vocab-tools/src/__snapshots__/class.test.ts.node.snap
+++ b/packages/vocab-tools/src/__snapshots__/class.test.ts.node.snap
@@ -10188,6 +10188,10 @@ get urls(): ((URL | Link))[] {
       return await ReplyAuthorization.fromJsonLd(json, options);
     }
     
+    if (values[\\"@type\\"].includes(\\"https://w3id.org/valueflows/ont/vf#Proposal\\")) {
+      return await Proposal.fromJsonLd(json, options);
+    }
+    
     if (values[\\"@type\\"].includes(\\"https://www.w3.org/ns/activitystreams#Activity\\")) {
       return await Activity.fromJsonLd(json, options);
     }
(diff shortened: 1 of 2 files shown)


packages/vocab/src/measure.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+- singularName: hasNumericalValue
+  functional: true
+  compactName: hasNumericalValue
+  uri: "http://www.ontology-of-units-of-measure.org/resource/om-2/hasNumericalValue"
+  description: |
+    The amount of the resource.  If not specified, arbitrary amounts can be
+    used when responding to the proposal.
+  range:
+  - "http://www.w3.org/2001/XMLSchema#string"
```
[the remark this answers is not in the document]
scammo: The Gemini comment is too short-sighted; xsd:float would indeed cause problems in a marketplace context due to the rounding errors documented in W3C/RDF-CONCEPTS#43.
The cleanest solution would be to add support for xsd:decimal in type.ts and then set hasNumericalValue to xsd:decimal. In JavaScript/TypeScript, however, you would need to decide whether to map it as a number (which would incur the same precision issues), a string (as many JSON APIs handle it) or a BigInt or Decimal library.
As a pragmatic workaround until xsd:decimal is properly supported, xsd:string is reasonable, even if it is not semantically ideal.

scammo:
FYI: I updated the comments, CHANGES.md and the snapshots to the current version :)

packages/vocab/src/measure.yaml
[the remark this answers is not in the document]
dahlia: Great analysis, @scammo. I agree that `xsd:float` would be the wrong direction here—rounding errors in a monetary/marketplace context are a real concern.

Using `xsd:string` as a pragmatic stopgap until `xsd:decimal` is properly supported makes sense for this PR's scope. However, `xsd:decimal` support is definitely worth pursuing properly. Would you be willing to open a separate issue for adding `xsd:decimal` support to the vocabulary type system? That would let us track it and discuss the TypeScript representation (number vs. string vs. a Decimal library) separately without blocking this PR.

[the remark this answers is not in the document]
scammo: Thanks for the feedback, @dahlia ! I agree that tracking xsd:decimal support separately is the right approach.

I've opened #617  to track adding xsd:decimal support to the vocabulary type system. That issue covers the TypeScript representation question (number vs. string vs. a Decimal library), so we can discuss it properly there.

For the work in this PR, xsd:string, this is a pragmatic solution to move this forward.

packages/vocab/src/proposal.yaml
```diff
@@ -0,0 +1,72 @@
+$schema: ../../vocab-tools/schema.yaml
+name: Proposal
+compactName: Proposal
+uri: "https://w3id.org/valueflows/ont/vf#Proposal"
+extends: "https://www.w3.org/ns/activitystreams#Object"
```
dahlia, line 5: According to FEP-0837, `Proposal` is a pure ValueFlows type, not an ActivityStreams `Object`. However, I see you've set `extends: "https://www.w3.org/ns/activitystreams#Object"` here.

This does have a practical benefit: inheriting properties like `name`, `content`, `attributedTo`, `published`, `to`, and `location` that the spec recommends for Proposal. So it may be an intentional design choice. Could you clarify the reasoning?

If the intent is to reuse AS properties for interoperability, that's reasonable, but it should be documented (e.g., in the description) so future maintainers understand this is a deliberate deviation from the spec rather than an oversight.

packages/vocab/src/intent.yaml
```diff
@@ -0,0 +1,72 @@
+$schema: ../../vocab-tools/schema.yaml
+name: Intent
+compactName: Intent
+uri: "https://w3id.org/valueflows/ont/vf#Intent"
```
dahlia, line 4: `Intent` is marked as `entity: true`, which means it can be independently fetched by URI. In FEP-0837, Intents are typically embedded within a Proposal and use fragment URIs (e.g., `#primary`, `#reciprocal`), which suggests they function more as embedded value objects.

Is there a use case where an `Intent` would be fetched independently outside of its parent `Proposal`? If not, `entity: false` might be more appropriate, similar to how `Measure` is defined.

Also, unlike `Proposal`, `Intent` has no `extends` field. If `Proposal` extends `Object` for practical reasons (reusing `name`, `content`, etc.), should `Intent` also extend `Object` for consistency, or is the asymmetry intentional?

packages/vocab/src/proposal.yaml
```diff
@@ -0,0 +1,72 @@
...
+  reciprocal: "vf:reciprocal"
+  action: "vf:action"
+  resourceConformsTo:
+    "@id": "vf:resourceConformsTo"
+    "@type": "@id"
+  resourceQuantity: "vf:resourceQuantity"
+  availableQuantity: "vf:availableQuantity"
+  minimumQuantity: "vf:minimumQuantity"
+  hasUnit: "om2:hasUnit"
+  hasNumericalValue: "om2:hasNumericalValue"
```
dahlia, line 50: The `defaultContext` here includes mappings for `Intent`, `action`, `resourceConformsTo`, `hasUnit`, `hasNumericalValue`, etc. These are properties that belong to `Intent` and `Measure`, not to `Proposal` itself.

I understand this is likely to ensure correct JSON-LD serialization when `Intent`/`Measure` objects are embedded. However, each of those types already defines its own `defaultContext`. Could you verify that the code generator handles nested contexts correctly without this duplication? If the duplication is necessary, a comment explaining why would be helpful.

packages/vocab/src/measure.yaml
dahlia: I see this was already discussed and `xsd:string` was agreed upon as a pragmatic stopgap until `xsd:decimal` support is added (#617). That's fine for now.

One small suggestion: it would be helpful to add a note in the `description` field mentioning that `xsd:string` is a temporary choice and that the value should represent a decimal number. This helps consumers of the API understand the expected format. For example:

```yaml
description: |
  The amount of the resource, represented as a decimal string.  If not
  specified, arbitrary amounts can be used when responding to the proposal.

  NOTE: This property uses `xsd:string` as a temporary measure until
  `xsd:decimal` is supported.  See #617.
```

CHANGES.md
```diff
@@ -610,6 +620,21 @@ Released on February 22, 2026.
...
+
+     -  Added `Proposal` class for publishing offers or requests.
+     -  Added `Intent` class for describing economic transactions within
+        a proposal, with `action`, `resourceConformsTo`, `resourceQuantity`,
+        `availableQuantity`, and `minimumQuantity` properties.
+     -  Added `Measure` class for representing quantities with units of
+        measure, with `hasUnit` and `hasNumericalValue` properties.
+
+[#578]: https://github.com/fedify-dev/fedify/issues/578
+[ValueFlows]: https://www.valueflo.ws/
```
dahlia: These entries appear under the **Version 1.5.0** section, which has already been released. Since this PR hasn't been merged yet, the changelog entry should only appear in the unreleased section at the top. Please remove this duplicate block.

Also, the `[#578]` reference link is defined twice (once in each section), and the `[ValueFlows]` reference link is defined but never used. Please clean these up.

CHANGES.md
```diff
@@ -81,6 +81,15 @@ To be released.
 
 ### @fedify/vocab
 
+ -  Added vocabulary types for economic resource coordination
+    in federated networks.  [[#578] by scammo]
+ -  Added `Proposal` class for publishing offers or requests.
+ -  Added `Intent` class for describing economic transactions within
+    a proposal, with `action`, `resourceConformsTo`, `resourceQuantity`,
+    `availableQuantity`, and `minimumQuantity` properties.
```
dahlia: The sub-items (`Proposal`, `Intent`, `Measure`) should be nested under the summary line rather than being flat siblings. Also, please run `mise run fmt` to fix Hongdown formatting issues.

Suggested structure:

```markdown
 -  Added vocabulary types for economic resource coordination
    in federated networks.  [[#578] by Samuel Brinkmann]

     -  Added `Proposal` class for publishing offers or requests.
     -  Added `Intent` class for describing economic transactions within
        a proposal, with `action`, `resourceConformsTo`, `resourceQuantity`,
        `availableQuantity`, and `minimumQuantity` properties.
     -  Added `Measure` class for representing quantities with units of
        measure, with `hasUnit` and `hasNumericalValue` properties.
```

packages/vocab/src/intent.yaml
scammo: FEP-0837 specifies that implementations SHOULD use fragment URIs (#primary, #reciprocal) to identify intents within a proposal, so they don't need independent dereferencing. I've changed Intent to `entity: false` to match Measure and reflect that intents are embedded value objects.

Regarding the asymmetry with Proposal: since Intent is now `entity: false`, it doesn't need to extend Object,  it's a pure value type embedded within its parent, so inheriting ActivityStreams properties like name, attributedTo, etc. wouldn't be useful here.

packages/vocab/src/measure.yaml
scammo: Thankyou for the suggestions, I added the Information.

packages/vocab/src/proposal.yaml
scammo: ou're right that FEP-0837 defines Proposal as a pure ValueFlows type. I chose to extend AS Object intentionally for practical interoperabilit. It gives Proposal useful inherited properties like name, content, attributedTo, published, to, and location that are essential for ActivityPub federation without needing to redefine them. As far as I understand (please correct my thinking is wrong), without this, other fediverse software wouldn't know how to display a Proposal.

I've added a note in the description documenting this as a deliberate deviation from the spec and explaining the reasoning.

packages/vocab/src/proposal.yaml
scammo: I verified this by temporarily removing the Intent/Measure mappings from Proposal's defaultContext and running codegen. Without them, the embedded Intent/Measure properties fall back to prefixed URIs instead of compact names:


// With duplication (correct):
```
"action": "transfer",
"hasUnit": "one"

```
// Without duplication (broken):
```
"vf:action": "transfer",
"om2:hasUnit": "one"
```
This happens because the code generator (codec.ts:134) embeds each type's own defaultContext directly — there's no recursive merging from referenced types. When Intent is serialized inside Proposal, it's called with context: undefined (type.ts:530-532), so it relies entirely on the parent's context for compaction.

I've added a YAML comment above defaultContext explaining why the duplication is necessary.

Please correct me, if this approach is not the usal way to handle such situations.

CHANGES.md
scammo: Thank you for your suggestion. I did make this changes!

CHANGES.md
scammo: I fixed this. The duplicate block has been removed from the released section. The entry now only appears in the unreleased section. I've also cleaned up the duplicate [#578] reference link and removed the unused [ValueFlows] reference.

packages/vocab/src/proposal.yaml
```diff
@@ -0,0 +1,84 @@
...
+  redefine them.
+# NOTE: The context includes mappings for Intent and Measure properties
+# (action, resourceConformsTo, hasUnit, etc.) because the code generator
+# does not automatically merge contexts from referenced types.  Proposal
+# must declare all mappings needed for correct JSON-LD serialization of
+# its embedded Intent and Measure objects.
+defaultContext:
+- "https://www.w3.org/ns/activitystreams"
+- "https://w3id.org/security/data-integrity/v1"
+- vf: "https://w3id.org/valueflows/ont/vf#"
```
dahlia, line 29: Because `Proposal` extends `Object`, `Proposal.toJsonLd()` will still emit inherited compact keys such as `sensitive`, `emojiReactions`, `interactionPolicy`, and `approvedBy`. This new `defaultContext` replaces the parent context instead of extending it, but it omits `https://gotosocial.org/ns` and the `sensitive`/`emojiReactions` aliases from `object.yaml`, so any `Proposal` using those inherited fields serializes to compact JSON-LD whose terms are no longer defined in the emitted `@context` and will not round-trip correctly for consumers.

packages/vocab-tools/src/__snapshots__/class.test.ts.deno.snap
```diff
@@ -10178,6 +10178,10 @@ get urls(): ((URL | Link))[] {
       return await ChatMessage.fromJsonLd(json, options);
     }
     
+<<<<<<< HEAD
+    if (values[\\"@type\\"].includes(\\"https://w3id.org/valueflows/ont/vf#Proposal\\")) {
+      return await Proposal.fromJsonLd(json, options);
+=======
```
dahlia: These snapshot updates now contain raw `<<<<<<<` / `=======` / `>>>>>>>` merge markers, and the same issue appears in the Node and Bun variants as well. *packages/vocab-tools/src/class.test.ts* loads these files verbatim for the `generateClasses()` snapshot assertions, so the snapshot suite will fail as soon as it compares the generator output against these unresolved merge artifacts.

packages/vocab/src/measure.yaml
```diff
@@ -0,0 +1,35 @@
...
+entity: false
+description: A quantity with a unit of measure.
+defaultContext:
+- "https://www.w3.org/ns/activitystreams"
+- om2: "http://www.ontology-of-units-of-measure.org/resource/om-2/"
+  hasUnit: "om2:hasUnit"
+  hasNumericalValue: "om2:hasNumericalValue"
+
+properties:
+- singularName: hasUnit
```
dahlia: At the Fedify API layer, `hasUnit` and `hasNumericalValue` read like Boolean predicates rather than value-bearing properties. Since Fedify already renames several vocabulary properties for a more natural TypeScript surface, it would be more ergonomic to keep the wire names in `compactName`/`uri` but expose these as `unit` and `numericalValue` via `singularName`. If you make that change, please also extend the renaming table in *docs/manual/vocab.md* so the API and manual stay aligned.

packages/vocab/src/measure.yaml
```diff
@@ -0,0 +1,35 @@
...
+  compactName: hasNumericalValue
+  uri: "http://www.ontology-of-units-of-measure.org/resource/om-2/hasNumericalValue"
+  description: |
+    The amount of the resource, represented as a decimal string.  If not
+    specified, arbitrary amounts can be used when responding to the proposal.
+
+    NOTE: This property uses `xsd:string` as a temporary measure until
+    `xsd:decimal` is supported.  See #617.
+  range:
+  - "http://www.w3.org/2001/XMLSchema#string"
```
dahlia: The earlier rationale for using `xsd:string` here was that Fedify could not model `xsd:decimal` yet, but #617 has since been completed by #640 and the runtime/codegen now support `xsd:decimal` as a dedicated `Decimal` type. Keeping this property as `xsd:string` still accepts arbitrary strings and loses the semantic validation this vocabulary now has available, so this range should be updated to `xsd:decimal` and the temporary note removed.

dahlia (changes requested):
Thanks for the updates. I'm still blocking on a few remaining issues: `Proposal` needs to preserve inherited `Object` term mappings in its context, the generated snapshot files currently contain unresolved merge markers and appear to be breaking the snapshot CI, and `Measure.hasNumericalValue` should now move from `xsd:string` to `xsd:decimal` since decimal support landed in #640. Please address the inline comments and then I'll take another look.

CHANGES.md
```diff
@@ -20,6 +20,20 @@ To be released.
  -  Added `DocumentLoaderFactoryOptions.maxRedirection` to configure the
     maximum number of redirects followed by `getDocumentLoader()`.
 
+### @fedify/vocab
```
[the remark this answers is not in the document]
scammo: As far as I can see, the heading style is consistent with all the other heading styles and the 'Mise Run Format'. If I have missed something, please clarify.

packages/vocab-tools/src/__snapshots__/class.test.ts.deno.snap
scammo: Sorry for that oversight. I have fixed it.

packages/vocab/src/measure.yaml
scammo: Thankyou fir the advice. Since the introduction of xsd:decimal support in #640, I have updated hasNumericalValue to use xsd:decimal and removed the temporary workaround note. The property now uses singularName: numericalValue to provide a cleaner TypeScript API, while retaining compactName: hasNumericalValue for JSON-LD serialisation.

packages/vocab/src/measure.yaml
scammo: I've renamed singularName to unit and numericalValue while keeping compactName as hasUnit and hasNumericalValue for the JSON-LD wire format. I've also updated the renaming table in docs/manual/vocab.md.

packages/vocab/src/proposal.yaml
scammo: I've added the missing https://gotosocial.org/ns/ context URL and the fedibird/sensitive/emojiReactions aliases to ensure that inherited compact keys are processed correctly. I've also added a comment to explain why the default context of the proposal needs to include mappings for the Intent and Measure properties (the code generator doesn't automatically merge contexts from referenced types).

From my understanding, in the long term it could be interesting to have other ways to extend the existing context for new vocabulary. However, I'm not sure what the best way forward is or whether this is an issue for other use cases.

packages/vocab/src/proposal.yaml
```diff
@@ -0,0 +1,90 @@
...
+  interoperability within ActivityPub federation, even though FEP-0837
+  defines `Proposal` as a pure ValueFlows type.  Extending `Object`
+  provides useful inherited properties such as `name`, `content`,
+  `attributedTo`, `published`, `to`, and `location` without needing to
+  redefine them.
+# NOTE: The context includes mappings for Intent and Measure properties
+# (action, resourceConformsTo, hasUnit, etc.) because the code generator
+# does not automatically merge contexts from referenced types.  Proposal
+# must declare all mappings needed for correct JSON-LD serialization of
+# its embedded Intent and Measure objects.
```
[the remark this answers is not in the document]
scammo: I'd rather not create a dedicated regression test for this. As far as I understand. The existing snapshot tests already cover proposal serialisation with embedded intent/measure objects, so any context drift would be revealed there. Adding another test specifically for context compaction would be redundant, adding unnecessary maintenance without providing any meaningful additional coverage.

Issue #481, opened 2025-11-07
Interoperability smoke test suite


dahlia:
### Summary

As a server framework, Fedify's core value lies in its ability to correctly interoperate with other ActivityPub implementations in the Fediverse. Currently, we rely on unit tests and manual testing, but we lack an automated, systematic way to verify E2E interoperability, similar to Node.js's CITGM (canary in the gold mine).

This issue proposes creating a new CI workflow dedicated to running smoke tests against live instances of major ActivityPub servers (e.g., Mastodon, Misskey) to ensure our federation logic is robust and compatible.

### Proposed solution

The plan involves three main components:

1.  **CI workflow:** A GitHub Actions workflow that uses `docker-compose` to spin up services for:
    * One or more target ActivityPub servers (e.g., a Mastodon instance).
    * Our “Fedify test harness” application.
2.  **Fedify test harness:** Since Fedify is a library, we will create a minimal, lightweight Fedify application *within this repository* (e.g., under `test/smoke-harness/`). Its sole purpose is to serve as an endpoint for these tests.
3.  **CI orchestrator:** The main test script (e.g., a Deno script) that orchestrates the E2E test.

### Implementation details

This E2E test cannot *only* test Fedify. It must verify that actions are correctly sent, received, and interpreted by both sides.

#### 1. Fedify test harness app

* It will be a minimal `fedify` app with basic handlers (e.g., for `Actor`, `Inbox`, `Outbox`).
* It will use a simple data store (e.g., in-memory or Deno KV).
* *Crucially*, it will expose internal “backdoor” APIs for the test orchestrator (e.g., `POST /_test/follow`, `POST /_test/create-note`, `GET /_test/get-latest-inbox-item`).

#### 2. CI orchestrator and verification

The orchestrator script will manage the entire test flow by communicating with *both* our test harness and the target server's API.

##### Example scenario: Fedify → Mastodon (`Create(Note)`)

1.  *Setup*: The orchestrator uses Mastodon's API (`tootctl` or REST) to create a test user (`@mastodon-user@mastodon.test`) and get an API token.
2.  *Action*: The orchestrator calls our harness's backdoor API: `POST /_test/create-note?content=...`
3.  *Federation*: Our harness app, using `fedify`, sends a `Create` activity to `@mastodon-user`'s inbox.
4.  *Verification*: The orchestrator uses the Mastodon API token to poll `@mastodon-user`'s home timeline (`GET /api/v1/timelines/home`).
[...]

PR #598, opened 2026-03-04
Add interoperability smoke tests for Mastodon (non-strict)

sij411:
## Summary

Implements the Phase 1 (Mastodon) interoperability smoke test suite described
in https://github.com/fedify-dev/fedify/issues/615.

### CI workflow (`.github/workflows/smoke-mastodon.yml`)

A GitHub Actions workflow that spins up a full Mastodon stack (web, sidekiq,
PostgreSQL, Redis) alongside a Fedify test harness via Docker Compose, then
runs E2E federation scenarios.

**When it runs:**

 -  On push to `main`, `next`, and `*.*-maintenance` branches
 -  On `workflow_dispatch` (manual trigger for testing specific branches)

It does **not** run on PRs, as the issue notes these tests are too
long-running and resource-intensive for every PR.

### Fedify test harness (`test/smoke/harness/`)

A minimal Fedify application that runs inside the Docker network.  It
exposes:

 -  Standard federation endpoints (actor, inbox, outbox, followers)
 -  Backdoor test-control APIs (`/_test/reset`, `/_test/inbox`,
    `/_test/follow`, `/_test/unfollow`, `/_test/create-note`,
    `/_test/health`)

### Orchestrator (`test/smoke/orchestrator.ts`)

A Deno script that drives 6 E2E scenarios through both the Mastodon API
and harness backdoor endpoints:

 -  Mastodon → Fedify (Follow)
 -  Fedify → Mastodon (Follow)
 -  Fedify → Mastodon (Create Note)
 -  Mastodon → Fedify (Reply)
 -  Mastodon → Fedify (Unfollow)
 -  Fedify → Mastodon (Unfollow)

Each follow scenario includes precondition checks (`ensureNotFollowing` /
`assertNotFollowing`) that verify the relationship starts clean before
exercising the follow path.

### Directory structure

```
test/smoke/
  orchestrator.ts              # shared test driver
  harness/                     # shared Fedify test harness
    main.ts
    federation.ts
    backdoor.ts
    store.ts
    deno.json
  mastodon/                    # Mastodon-specific
    docker-compose.yml
    mastodon.env
    provision.sh
    disable_force_ssl.rb
```

Mastodon-specific files live in `test/smoke/mastodon/` so that future
server targets (e.g., Misskey for Phase 1, Akkoma for Phase 2) can be
added as sibling directories sharing the same harness and orchestrator
infrastructure.

## Acceptance criteria (from the issue)

 -  [x] CI workflow created
 -  [x] Minimal Fedify test harness app built within the repo
 -  [x] E2E tests (`Follow`, `Create(Note)`, reply) pass against Mastodon
 -  [x] Workflow runs on pushes to `main`, `next`, `*.*-maintenance`, and
    on `workflow_dispatch`

## Test plan

 -  [x] All 6 scenarios pass locally via `act`
 -  [x] CI workflow `smoke-mastodon` passes on push to `main`

Related to #615 and #481  


.github/workflows/smoke-mastodon.yml
```diff
@@ -0,0 +1,91 @@
...
+          SECRET2=$(docker run --rm "$IMAGE" bundle exec rails secret)
+
+          {
+            echo "SECRET_KEY_BASE=$SECRET1"
+            echo "OTP_SECRET=$SECRET2"
+            docker run --rm "$IMAGE" bundle exec rails mastodon:webpush:generate_vapid_key \
+              | grep -E '^[A-Z_]+=.+'
+            docker run --rm "$IMAGE" bundle exec rails db:encryption:init \
+              | grep -E '^[A-Z_]+=.+'
+          } >> test/smoke/mastodon/mastodon.env
```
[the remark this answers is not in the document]
sij411: It doesn't.

test/smoke/orchestrator.ts
```diff
@@ -0,0 +1,346 @@
...
+
+async function testReply(): Promise<void> {
+  await harnessPost("/_test/reset");
+
+  const handle = `@testuser@${HARNESS_HOST}`;
+  const replyContent = `Reply smoke test ${Date.now()} ${handle}`;
+
+  await serverPost("/api/v1/statuses", {
+    status: replyContent,
+  });
```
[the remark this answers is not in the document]
sij411: It tests mention delivery.

test/smoke/harness/federation.ts
```diff
@@ -0,0 +1,98 @@
+import { createFederation, MemoryKvStore } from "@fedify/fedify/federation";
+import { generateCryptoKeyPair } from "@fedify/fedify/sig";
+import { Accept, Activity, Create, Follow, Person } from "@fedify/vocab";
```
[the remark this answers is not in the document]
sij411: it is being used...?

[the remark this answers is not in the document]
dahlia: Apparently false alarm.

test/smoke/mastodon/docker-compose.yml
```diff
@@ -0,0 +1,91 @@
...
+    networks: [smoke]
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 5s
+      retries: 10
+
+  # Fedify test harness — runs inside the Docker network so Mastodon's
+  # Resolv::DNS can resolve "fedify-harness" natively via Docker DNS.
+  fedify-harness:
+    image: denoland/deno:2.7.1
```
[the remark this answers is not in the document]
sij411: It doesn't make sense.

dahlia:
@codex review

test/smoke/orchestrator.ts
```diff
@@ -0,0 +1,359 @@
...
+
+  const handle = `@testuser@${HARNESS_HOST}`;
+  const replyContent = `Reply smoke test ${Date.now()} ${handle}`;
+
+  await serverPost("/api/v1/statuses", {
+    status: replyContent,
+    in_reply_to_id: parent.id,
+  });
+
+  await pollHarnessInbox("Create");
```
dahlia: The reply scenario now correctly sets `in_reply_to_id`, which is great. However, the harness `store` only records `{ id, type, receivedAt }` for each received activity. It doesn't capture the `inReplyTo` property of the inner `Note` object.

This means the test can only verify that *a* `Create` activity arrived at the harness inbox, but it cannot distinguish between a reply and a plain mention. If Mastodon were to deliver the activity without the `inReplyTo` field (or with the wrong value), this test would still pass.

To make this scenario meaningful, consider storing a bit more detail when a `Create` activity is received (at minimum the `inReplyTo` URI of the object) and then asserting on it in the orchestrator. Something like:

```typescript
// store.ts
export interface ReceivedActivity {
  id: string;
  type: string;
  receivedAt: string;
  inReplyTo?: string;  // for Create(Note) activities
}

// federation.ts
.on(Create, async (_ctx, create) => {
  const object = await create.getObject();
  store.push({
    id: create.id?.href ?? crypto.randomUUID(),
    type: "Create",
    receivedAt: new Date().toISOString(),
    inReplyTo: object instanceof Note ? object.inReplyToId?.href : undefined,
  });
})

// orchestrator.ts — assert the received Create has inReplyTo
```

test/smoke/orchestrator.ts
```diff
@@ -0,0 +1,359 @@
...
+    ["Mastodon → Fedify (Follow)", testFollowMastodonToFedify],
+    ["Fedify → Mastodon (Follow)", testFollowFedifyToMastodon],
+    ["Fedify → Mastodon (Create Note)", testCreateNote],
+    ["Mastodon → Fedify (Reply)", testReply],
+    ["Mastodon → Fedify (Unfollow)", testUnfollowMastodonFromFedify],
+    ["Fedify → Mastodon (Unfollow)", testUnfollowFedifyFromMastodon],
+  ];
+
+  let failed = false;
+  for (const [name, fn] of scenarios) {
```
dahlia, line 394: Each scenario calls `/_test/reset` at the start to clear the harness inbox, but ActivityPub delivery is asynchronous. Mastodon processes outgoing activities through Sidekiq, which may deliver them after the orchestrator has already moved on to the next scenario and called reset.

For example, if the `testCreateNote` scenario completes (the note appeared on the Mastodon timeline) but Mastodon hasn't yet delivered a follow-up side effect, that activity could arrive during `testReply` and be mistaken for the expected result, or vice versa.

A couple of options to mitigate this:

1. **Add a short settling delay** after each scenario before resetting (e.g., 2–3 seconds), giving Sidekiq time to flush its queue.
2. **Tag activities with scenario-specific markers** (e.g., include a unique token in note content or activity IDs) and filter on that token in `pollHarnessInbox`, so stale deliveries from a previous scenario are ignored.
3. **Check the inbox is empty** after reset and before the scenario begins its action, polling briefly to confirm no stragglers arrived.

Option 2 would be the most robust, as it doesn't rely on timing assumptions.

test/smoke/orchestrator.ts
sij411: Great. got it.

test/smoke/orchestrator.ts
sij411: Good. I'll apply.

test/smoke/orchestrator.ts
```diff
@@ -0,0 +1,404 @@
...
+  await serverPost(`/api/v1/accounts/${accountId}/follow`);
+
+  await pollHarnessInbox("Follow", (a) => !knownIds.has(a.id));
+
+  await poll("follow accepted", async () => {
+    const rels = await serverGet(
+      `/api/v1/accounts/relationships?id[]=${accountId}`,
+    ) as Relationship[];
+    const rel = rels.find((r) => r.id === accountId);
+    return rel?.following ? rel : null;
```
dahlia: `relationships[].following` does not seem strong enough to prove that the follow handshake actually completed.

In Mastodon, that flag is backed by the local follow row, so it can become `true` as soon as `POST /api/v1/accounts/{id}/follow` succeeds, before the ActivityPub `Accept` from the harness has been received and processed. In other words, this test is currently labeled as “follow accepted”, but it may still pass even if the harness stops sending `Accept` or the `Accept` never gets delivered.

That makes this a false positive for the interoperability scenario we want to cover. Could we assert on the `Accept` more directly here, or at least wait on some state transition that only happens after Mastodon has actually processed the `Accept`?

dahlia (changes requested):

test/smoke/orchestrator.ts
sij411: Great point. I will work on that.

sij411: Done with [this](https://github.com/fedify-dev/fedify/pull/598/commits/48a838e4179aa0c5a26cca82240d8231c3192821)

dahlia (approved):
Great job!

PR #601, opened 2026-03-05
Add @fedify/solidstart package for SolidStart integration

### LINKED ISSUE #476: Create @fedify/solidstart package for SolidStart integration
### Summary

Add official SolidStart integration package (@fedify/solidstart) to enable seamless federation capabilities in SolidStart applications, similar to existing integrations for SvelteKit (@fedify/sveltekit) and Next.js (@fedify/next).

### Problem

  SolidStart is a popular meta-framework for building Solid.js applications with server-side rendering, and developers using SolidStart currently lack a streamlined way to integrate Fedify's ActivityPub federation capabilities. While Fedify can theoretically work with any framework through its framework-agnostic core, developers have to manually handle request/response conversions and middleware setup, leading to boilerplate code and potential compatibility issues. This creates an unnecessary barrier for SolidStart developers who want to build federated applications and participate in the Fediverse ecosystem.

### Proposed Solution

 Create a dedicated @fedify/solidstart package in the packages/ directory that provides native SolidStart integration following the established patterns from other framework integrations. The package should:

  - Provide middleware/handlers compatible with SolidStart's API routes and server functions
  - Handle request/response object conversion between SolidStart and Fedify's internal interfaces
  - Include TypeScript types specific to SolidStart's request/response patterns
  - Follow the same builder pattern and configuration approach as existing integrations
  - Include an example application demonstrating usage in the examples/ directory
  - Provide comprehensive documentation.

This would allow SolidStart developers to add federation capabilities with minimal configuration, matching the developer experience already available to users of other meta-frameworks.

### Alternatives Considered

_No response_

### Scope / Dependencies

_No response_

dodok8:
Summary
-------

Add SolidStart integration support for Fedify, including a new
`@fedify/solidstart` package, example application, documentation,
and `fedify init` scaffolding support.

Closes #476.

Related issue
-------------

 -  closes https://github.com/fedify-dev/fedify/issues/476

Changes
-------

 -  Added `@fedify/solidstart` package (*packages/solidstart/*) providing
    `fedifyMiddleware()` for integrating Fedify with SolidStart's
    middleware system
 -  Added SolidStart example application (*examples/solidstart/*)
    demonstrating actor dispatching, WebFinger, and NodeInfo
 -  Added SolidStart section to the integration documentation
    (*docs/manual/integration.md*)
 -  Added `"solidstart"` option to `fedify init` with template files
    for app config, middleware, routes, and entry points
 -  Added `BANNED_COMBOS` mechanism to the test-init lookup filter
    to support per-framework+PM granular test skipping
 -  Fixed missing `::: code-group` / `:::` wrapping in the Elysia
    integration documentation section
 -  Updated *CHANGES.md* with the new package entry under Version 2.1.0
 -  Updated *README.md* and *packages/fedify/README.md* with JSR badge
    for `@fedify/solidstart`

Benefits
--------

 -  Users can now build federated applications with SolidStart, a
    modern full-stack framework built on SolidJS
 -  `fedify init` supports scaffolding SolidStart projects with all
    package managers (Deno, pnpm, npm)
 -  The `BANNED_COMBOS` test mechanism allows finer-grained control
    over which framework+PM combinations are tested for server lookup,
    rather than banning entire frameworks

Checklist
---------

 -  [x] Did you write some relevant docs about this change (if it's a new feature)?
 -  [ ] Did you write a regression test to reproduce the bug (if it's a bug fix)?

Additional notes
----------------

 -  The `fedify init` lookup test skips the `solidstart` + `deno`
    combination because Deno's `links` feature does not populate
    `node_modules/`, which Vite (used internally by vinxi/SolidStart)
    requires for SSR module resolution.  The `pnpm` and `npm` lookup
    tests pass fully.
 -  The `@fedify/solidstart` middleware converts SolidStart's H3 events
    to standard `Request`/`Response` objects for Fedify's federation
    handler, similar to how `@fedify/h3` works.

### DIFF
--- a/packages/init/src/test/lookup.ts
+++ b/packages/init/src/test/lookup.ts
@@ -19,9 +19,16 @@ import webFrameworks from "../webframeworks/mod.ts";
 const HANDLE = "john";
 const STARTUP_TIMEOUT = 30000; // 30 seconds
 const CWD = process.cwd();
-const BANNED_WFS: WebFramework[] = ["next"];
-const BANNED_COMBOS: [WebFramework, PackageManager][] = [
-  ["solidstart", "deno"],
+type LookupCase = [WebFramework, PackageManager, KvStore, MessageQueue];
+type LookupCasePattern = [
+  WebFramework | "*",
+  PackageManager | "*",
+  KvStore | "*",
+  MessageQueue | "*",
+];
+const BANNED_LOOKUP_CASES: LookupCasePattern[] = [
+  ["next", "*", "*", "*"],
+  ["solidstart", "deno", "*", "*"],
 ];
 
 /**
@@ -37,7 +44,7 @@ export default async function runServerAndLookupUser(
     printErrorMessage`\nNo directories to lookup test.`;
     return;
   }
-  const filtered = filterWebFrameworks(valid);
+  const filtered = filterLookupDirs(valid);
 
   printMessage`\nLookup Test start for ${String(filtered.length)} app(s)!`;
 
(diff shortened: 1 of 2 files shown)


packages/init/src/test/lookup.ts
```diff
@@ -56,16 +59,23 @@ function filterWebFrameworks(
...
+    const [wf, pm] = dir.split(sep).slice(-4, -2) as [
+      WebFramework,
+      PackageManager,
+    ];
+    if (BANNED_WFS.includes(wf)) return false;
+    if (BANNED_COMBOS.some(([bwf, bpm]) => bwf === wf && bpm === pm)) {
+      return false;
+    }
+    return true;
+  });
```
2chanhaeng: How about make this like
```ts
const BANNED: [
  WebFramework | "*",
  PackageManager | "*",
  MessageQueue | "*",
  KvStore | "*"
][] = [
  ["next", "*", "*", "*"],
  ["solidstart", "deno", "*", "*"],
];

// ...
  return dirs.filter((dir) => {
    const option = dir.split(sep).slice(-4, 0) as [
      WebFramework,
      PackageManager,
      MessageQueue,
      KvStore
    ];
    return !BANNED.some((banned) => banned.every((b, i) => b === "*" || b === option[i])
  ))});
```

2chanhaeng: Type assertion might be confusing. Just using `string` would be fine (if there are no type errors), I think?

dodok8: I applied this refactoring in 4bc79c6a.

packages/solidstart/package.json
```diff
@@ -0,0 +1,65 @@
...
+  "main": "./dist/index.cjs",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": {
+        "import": "./dist/index.d.ts",
+        "require": "./dist/index.d.cts",
+        "default": "./dist/index.d.ts"
+      },
+      "import": "./dist/index.js",
```
dahlia, line 41: `tsdown`'s current output does not match the paths exported here. After building `@fedify/solidstart`, the package emits `dist/index.mjs` and `dist/index.d.mts`, but this manifest points ESM and type resolution at `./dist/index.js` and `./dist/index.d.ts`. That means npm consumers can fail with `ERR_MODULE_NOT_FOUND`, and TypeScript consumers may not pick up the declarations at all. Could we align the `exports` entries with the actual build output, following the same pattern used by the other integration packages?

packages/init/src/templates/solidstart/src/middleware/index.ts.tpl
```diff
@@ -0,0 +1,4 @@
+import { fedifyMiddleware } from "@fedify/solidstart";
+import federation from "../federation.ts";
```
dahlia, line 2: The generated import should probably be extensionless here. With the current `../federation.ts` specifier, freshly scaffolded npm/pnpm/yarn/bun SolidStart apps fail type-checking with TS5097 because the generated `tsconfig.json` does not enable `allowImportingTsExtensions`. Using `../federation` would avoid breaking the generated project out of the box.

packages/init/src/templates/solidstart/app.config.ts.tpl
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from "@solidjs/start/config";
+
+export default defineConfig({
+  middleware: "src/middleware/index.ts",
+  server: {
+    preset: "node-server",
```
dahlia, line 6: This template is shared across all SolidStart scaffolds, but the server preset is hard-coded to Nitro's Node target. That leaves the advertised Deno scaffold misconfigured from the moment it is generated. Could we make the preset runtime-specific during template generation so that `--package-manager deno` emits the Deno-compatible preset instead of the Node one?

packages/init/src/test/lookup.ts
```diff
@@ -49,31 +59,50 @@ export default async function runServerAndLookupUser(
...
-  const hasBanned = BANNED_WFS.filter((wf) => wfs.has(wf));
-  if (isEmpty(hasBanned)) {
-    return dirs;
+  const hasBanned = getBannedLookupFrameworks().filter((wf) => wfs.has(wf));
+  if (!isEmpty(hasBanned)) {
+    const bannedLabels = hasBanned.map((wf) => webFrameworks[wf]["label"]);
+    printErrorMessage`\n${
+      values(bannedLabels)
+    } is not supported in lookup test yet.`;
   }
```
2chanhaeng, line 93: This log only shows the banned web frameworks, but not for combinations. It would be even better if logs were provided to identify which specific combinations were filtered out.

2chanhaeng (changes requested):

PR #626, opened 2026-03-16
Implement RFC 9421 §5 `Accept-Signature` challenge-response negotiation

### LINKED ISSUE #583: RFC 9421 outbound negotiation: implement `Accept-Signature` parsing and challenge-driven re-signing in `doubleKnock`
## Background
Fedify already implements HTTP Message Signatures from [RFC 9421](https://www.rfc-editor.org/rfc/rfc9421), but it does not yet implement `Accept-Signature` negotiation from [RFC 9421 §5](https://www.rfc-editor.org/rfc/rfc9421#section-5), [RFC 9421 §5.1](https://www.rfc-editor.org/rfc/rfc9421#section-5.1), and [RFC 9421 §5.2](https://www.rfc-editor.org/rfc/rfc9421#section-5.2).

## Current implementation references
- `doubleKnock()` currently retries by status code and does not parse or apply `Accept-Signature` challenge metadata: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/sig/http.ts#L1346-L1443
- Retry condition and fallback branch are status-based today: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/sig/http.ts#L1389-L1403
- RFC 9421 signature label is effectively fixed (`sig1`) in current request-signing path: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/sig/http.ts#L291-L301
- RFC 9421 covered components are currently fixed in request signing: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/sig/http.ts#L430-L483
- Existing parsing helpers are focused on `Signature-Input` and are not a dedicated `Accept-Signature` parser per [RFC 9421 §5.1](https://www.rfc-editor.org/rfc/rfc9421#section-5.1): https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/sig/http.ts#L308-L359
- Documentation currently explains double-knocking as two-spec fallback and does not cover challenge-aware negotiation from [RFC 9421 §5.2](https://www.rfc-editor.org/rfc/rfc9421#section-5.2): https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/docs/manual/send.md#L973-L980

## Problem statement
When a remote server requests specific covered components or metadata parameters such as `nonce`, `tag`, `alg`, or `keyid` (see [RFC 9421 §2.3](https://www.rfc-editor.org/rfc/rfc9421#section-2.3) and [RFC 9421 §5.1](https://www.rfc-editor.org/rfc/rfc9421#section-5.1)), Fedify cannot currently adapt the retry signature shape and may fail even when an interoperable retry is possible.

## Goal
Add outbound `Accept-Signature`-aware negotiation so `doubleKnock()` can consume challenge metadata and retry with a compatible [RFC 9421](https://www.rfc-editor.org/rfc/rfc9421) signature before applying legacy fallback behavior.

## Scope
This issue covers stage 1-3 only and is limited to outbound behavior.

## Proposed work
1. Add an `Accept-Signature` parser and internal model for Dictionary Structured Field members from [RFC 9421 §5.1](https://www.rfc-editor.org/rfc/rfc9421#section-5.1), including label, covered components, and requested metadata parameters.
2. Add validation for target-message applicability rules from [RFC 9421 §5](https://www.rfc-editor.org/rfc/rfc9421#section-5), such as rejecting response-only identifiers like `@status` for request-target signatures.
3. Extend RFC 9421 request-signing options to allow challenge-driven label, covered components, and requested metadata parameter inclusion.
4. Enforce safety constraints so remote-requested `alg` and `keyid` are only honored when compatible with local key material and supported algorithms.
5. Integrate challenge-aware retry into `doubleKnock()` so relevant failure responses first attempt `Accept-Signature`-compatible re-signing.
6. Keep existing fallback behavior when challenge data is missing, invalid, or not fulfillable.
7. Add loop prevention so challenge-driven retries are bounded per outbound request.
8. Keep `specDeterminer` persistence coherent with successful negotiated outcomes: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/sig/http.ts#L1251-L1269

## Out of scope
Server-side challenge emission in inbox 401 responses, response-signing APIs, and request-response binding support from [RFC 9421 §2.4](https://www.rfc-editor.org/rfc/rfc9421#section-2.4) are intentionally excluded and will be tracked in issue #584.

## Acceptance criteria
- [ ] Unit tests cover valid and invalid `Accept-Signature` parsing cases.
- [ ] Signing tests cover challenge-driven label/component/parameter customization.
- [ ] `doubleKnock()` integration tests verify challenge-aware retry behavior.
- [ ] Regression tests confirm current fallback behavior remains intact when no usable challenge exists.
- [ ] User docs are updated to describe outbound challenge-aware negotiation with links to [RFC 9421 §5](https://www.rfc-editor.org/rfc/rfc9421#section-5).

### LINKED ISSUE #584: RFC 9421 inbound negotiation: emit `Accept-Signature` challenges on inbox authentication failures
## Background
This issue is the server-side follow-up to #583 and targets RFC 9421 `Accept-Signature` usage in HTTP responses so Fedify can request properly signed next client requests as described in [RFC 9421 §5](https://www.rfc-editor.org/rfc/rfc9421#section-5), [§5.1](https://www.rfc-editor.org/rfc/rfc9421#section-5.1), and [§5.2](https://www.rfc-editor.org/rfc/rfc9421#section-5.2).

## Related issue
Outbound processing of `Accept-Signature` challenges is tracked separately in https://github.com/fedify-dev/fedify/issues/583.

## Current implementation references
- Inbox verification failure currently returns plain `401` without `Accept-Signature`: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/federation/handler.ts#L739-L766
- Actor/key mismatch currently returns plain `401` without `Accept-Signature`: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/federation/handler.ts#L787-L808
- Inbox handler parameters currently have no unauthorized/challenge callback surface: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/federation/handler.ts#L516-L541
- Middleware wires inbox handling directly through `handleInbox(...)`: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/federation/middleware.ts#L1474-L1489
- Existing KV prefixes have no dedicated namespace for challenge nonces: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/federation/middleware.ts#L146-L175
- Existing tests currently assert `401` outcomes but no challenge header behavior: https://github.com/fedify-dev/fedify/blob/0ce42c7c5e958150fde932cb5794d507d3359496/packages/fedify/src/federation/handler.test.ts#L1103-L1127

## Problem statement
When inbox authentication fails, Fedify returns `401` without server-side signature negotiation hints, so senders do not learn the expected signature shape and cannot reliably retry with RFC 9421 parameters such as `created` and, when enabled, `nonce` from [RFC 9421 §5.1](https://www.rfc-editor.org/rfc/rfc9421#section-5.1).

## Goal
Add inbound challenge emission so `401` responses from inbox authentication failures can include `Accept-Signature` requests that are valid for request messages and actionable by clients.

## Scope
This issue covers inbox and shared inbox authentication-failure responses only, and it does not include outbound challenge processing or response-signing APIs.

## Proposed work
1. Add a configurable inbox challenge policy that can emit an `Accept-Signature` request for the client’s next request according to [RFC 9421 §5](https://www.rfc-editor.org/rfc/rfc9421#section-5).
2. Define a default requested component set that is request-applicable and verifiable by Fedify, and explicitly avoid response-only identifiers such as `@status` per [RFC 9421 §5](https://www.rfc-editor.org/rfc/rfc9421#section-5).
3. Support challenge metadata parameters from [RFC 9421 §5.1](https://www.rfc-editor.org/rfc/rfc9421#section-5.1), with `created` requested by default and `nonce` explicitly optional.
4. If optional nonce mode is enabled, add nonce storage/consumption with a new KV prefix and bounded TTL for replay resistance.
5. Attach `Accept-Signature` on `401` responses for missing/invalid HTTP signature failures, and keep cache behavior explicit (for example `Cache-Control: no-store`) to avoid stale challenges.
6. Keep actor/key mismatch handling separate by default, because re-signing does not resolve impersonation, and only challenge that path if an explicit policy opts in.

## Out of scope
This issue does not implement outbound `Accept-Signature` processing (covered by #583), does not add full response-signature support from [RFC 9421 §2.4](https://www.rfc-editor.org/rfc/rfc9421#section-2.4), and does not change non-inbox endpoint authorization flows.

## Acceptance criteria
- [ ] Unsigned or invalidly signed inbox POST requests return `401` with an `Accept-Signature` header when challenge policy is enabled.
- [ ] Emitted `Accept-Signature` values only use request-applicable covered components and conform to [RFC 9421 §5.1](https://www.rfc-editor.org/rfc/rfc9421#section-5.1).
- [ ] Nonce support remains optional, and nonce checks are enforced only when optional nonce mode is enabled.
- [ ] Existing successful inbox processing remains unchanged for valid signatures.
- [ ] Existing non-challenge behavior remains available when challenge policy is disabled.
- [ ] Documentation is updated to describe inbox `401` challenge behavior and its RFC references.

2chanhaeng:
Summary
-------

Implement `Accept-Signature` challenge-response negotiation (RFC 9421 §5)
for both inbound and outbound sides of Fedify's federation layer.

Related issue
-------------

 -  closes #583
 -  closes #584

Changes
-------

 -  Added RFC 9421 §5 `Accept-Signature` negotiation for both outbound and inbound paths.  On the outbound side, `doubleKnock()` now parses `Accept-Signature` challenges from `401` responses and retries with a compatible RFC 9421 signature before falling back to legacy spec-swap.
- On the inbound side, a new `InboxChallengePolicy` option in `FederationOptions` enables emitting `Accept-Signature` headers on inbox `401` responses, with optional one-time nonce support for replay protection.

Benefits
--------

 -  Enables standards-based signature negotiation with servers that
    require specific RFC 9421 parameters
 -  Provides inbox operators an opt-in mechanism to enforce signature
    requirements and replay protection
 -  Maintains full backward compatibility — opt-in on inbound,
    transparent on outbound

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -36,6 +36,7 @@
     "@cloudflare/workers-types": "npm:@cloudflare/workers-types@^4.20250529.0",
     "@david/dax": "jsr:@david/dax@^0.43.2",
     "@fxts/core": "npm:@fxts/core@^1.21.1",
+    "@hongminhee/localtunnel": "jsr:@hongminhee/localtunnel@^0.3.0",
     "@js-temporal/polyfill": "npm:@js-temporal/polyfill@^0.5.1",
     "@logtape/file": "jsr:@logtape/file@^2.0.0",
     "@logtape/logtape": "jsr:@logtape/logtape@^2.0.0",
--- a/deno.lock
+++ b/deno.lock
@@ -7511,6 +7511,7 @@
   "workspace": {
     "dependencies": [
       "jsr:@david/dax@~0.43.2",
+      "jsr:@hongminhee/localtunnel@0.3",
       "jsr:@hono/hono@^4.8.3",
       "jsr:@logtape/file@2",
       "jsr:@logtape/logtape@2",
@@ -7575,7 +7576,6 @@
       },
       "examples/rfc-9421-test": {
         "dependencies": [
-          "jsr:@hongminhee/localtunnel@0.3",
           "jsr:@hono/hono@^4.7.1"
         ]
       },
--- a/examples/rfc-9421-test/deno.json
+++ b/examples/rfc-9421-test/deno.json
@@ -1,6 +1,5 @@
 {
   "imports": {
-    "@hongminhee/localtunnel": "jsr:@hongminhee/localtunnel@^0.3.0",
     "hono": "jsr:@hono/hono@^4.7.1"
   },
   "tasks": {
--- a/examples/test-examples/mod.ts
+++ b/examples/test-examples/mod.ts
@@ -22,6 +22,7 @@
  */
 
 import $, { type CommandChild } from "@david/dax";
+import { openTunnel, type Tunnel } from "@hongminhee/localtunnel";
 import { configure, getConsoleSink, getLogger } from "@logtape/logtape";
 import { fromFileUrl, join } from "@std/path";
 
@@ -283,6 +284,11 @@ const SKIPPED_EXAMPLES: SkippedExample[] = [
     reason:
       "No actor dispatcher configured; federation lookup cannot be verified",
   },
+  {
+    name: "rfc-9421-test",
+    reason:
+      "Requires live interaction with external fediverse servers (Bonfire, Mastodon)",
+  },
 ];
 
 // ─── ANSI Colors ──────────────────────────────────────────────────────────────
@@ -395,70 +401,22 @@ function forceKillChild(child: CommandChild): void {
 // ─── Tunnel ───────────────────────────────────────────────────────────────────
 
 /**
- * Starts `fedify tunnel -s pinggy.io <port>` and waits up to `timeoutMs`
- * for the tunnel URL to appear in its output.  The tunnel process is kept
- * alive and returned to the caller; it must be killed when no longer needed.
- *
- * Returns `null` if the URL was not found before the timeout.
+ * Opens a tunnel via `@hongminhee/localtunnel` (pinggy.io) to expose
+ * a local port.  Returns the {@link Tunnel} object or `null` on failure.
  */
-async function startTunnel(
-  port: number,
-  timeoutMs: number,
-): Promise<{ child: CommandChild; url: string } | null> {
+async function startTunnel(port: number): Promise<Tunnel | null> {
   const tunnelLogger = getLogger(["fedify", "examples", "tunnel"]);
-  tunnelLogger.info("Opening localhost.run tunnel on port {port}", { port });
-
-  const child = $`deno task cli tunnel -s pinggy.io ${String(port)}`
-    .cwd(REPO_ROOT)
-    .stdout("piped")
-    .stderr("piped")
-    .noThrow()
-    .spawn();
-
-  // Accumulate text from both streams while logging each chunk at DEBUG.
-  const textChunks: string[] = [];
-  const decoder = new TextDecoder();
-
-  const readStream = (stream: ReadableStream<Uint8Array>) => {
-    (async () => {
-      const reader = stream.getReader();
-      try {
-        while (true) {
-          const { done, value } = await reader.read();
-          if (done) break;
-          const text = decoder.decode(value, { stream: true });
-          textChunks.push(text);
-          const trimmed = text.trim();
-          if (trimmed) tunnelLogger.debug("{output}", { output: trimmed });
-        }
-      } catch {
-        // Stream may error when the process is killed.
-      }
-    })();
-  };
-
-  readStream(child.stdout());
-  readStream(child.stderr());
-
-  // Poll until we find an https URL in the accumulated output.
-  // The `message` template tag from @optique/run may wrap the URL in double
-  // quotes in non-TTY output, so we stop matching at whitespace or quotes.
-  const deadline = Date.now() + timeoutMs;
-  while (Date.now() < deadline) {
-    const match = textChunks.join("").match(/https:\/\/[^\s"']+/);
-    if (match) {
-      tunnelLogger.info("Tunnel established at {url}", { url: match[0] });
-      return { child, url: match[0] };
-    }
-    await new Promise((r) => setTimeout(r, 200));
+  tunnelLogger.info("Opening tunnel on port {port}", { port });
+  try {
+    const tunnel = await openTunnel({ port, service: "pinggy.io" });
+    tunnelLogger.info("Tunnel established at {url}", {
+      url: tunnel.url.href,
+    });
+    return tunnel;
+  } catch (error) {
+    tunnelLogger.error("Failed to open tunnel: {error}", { error });
+    return null;
   }
-
-  tunnelLogger.error(
-    "Tunnel did not produce a URL within {timeout} ms",
-    { timeout: timeoutMs },
-  );
-  forceKillChild(child);
-  return null;
 }
 
 // ─── Test Runners ─────────────────────────────────────────────────────────────
@@ -549,7 +507,7 @@ async function testServerExample(
   const collectServerOutput = () =>
     stdoutChunks.join("") + stderrChunks.join("");
 
-  let tunnelChild: CommandChild | null = null;
+  let activeTunnel: Tunnel | null = null;
 
   try {
     console.log(
@@ -572,18 +530,18 @@ async function testServerExample(
     });
     console.log(c.dim(`  server ready — opening tunnel on port ${port}…`));
 
-    const tunnel = await startTunnel(port, 30_000);
+    const tunnel = await startTunnel(port);
     if (tunnel == null) {
-      const error = "fedify tunnel did not produce a URL within 30s";
+      const error = "Failed to open tunnel";
       serverLogger.error("{error}", { error });
       return { name, status: "fail", error, output: collectServerOutput() };
     }
 
-    tunnelChild = tunnel.child;
-    const tunnelHostname = new URL(tunnel.url).hostname;
+    activeTunnel = tunnel;
+    const tunnelHostname = tunnel.url.hostname;
     const handle = `@${actor}@${tunnelHostname}`;
 
-    console.log(c.dim(`  tunnel URL : ${tunnel.url}`));
+    console.log(c.dim(`  tunnel URL : ${tunnel.url.href}`));
     console.log(c.dim(`  running    : fedify lookup ${handle} -d`));
     serverLogger.info("Running fedify lookup {handle}", { handle });
 
@@ -606,10 +564,10 @@ async function testServerExample(
     serverLogger.error("{error}", { error });
     return { name, status: "fail", error, output: lookupOutput };
   } finally {
-    // Force-kill tunnel first (it holds a connection to the server).
-    if (tunnelChild != null) {
-      serverLogger.debug("Force-killing tunnel process");
-      forceKillChild(tunnelChild);
+    // Close tunnel first (it holds a connection to the server).
+    if (activeTunnel != null) {
+      serverLogger.debug("Closing tunnel");
+      await activeTunnel.close();
     }
     serverLogger.debug("Force-killing server process");
     forceKillChild(serverChild);


2chanhaeng:
@codex review

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1679,68 @@ export async function respondWithObjectIfAcceptable(
...
+];
+
+function generateNonce(): string {
+  const bytes = new Uint8Array(16);
+  crypto.getRandomValues(bytes);
+  // Base64url encoding without padding
+  return btoa(String.fromCharCode(...bytes))
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_")
+    .replace(/=+$/, "");
```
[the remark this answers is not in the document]
2chanhaeng: `btoa()` fails when a character's code point exceeds 255 and `Uint8Array` values are always 0–255, so `String.fromCharCode(...bytes)` is no problem.

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1679,68 @@ export async function respondWithObjectIfAcceptable(
...
+  for (const sig of globalThis.Object.values(parsed)) {
+    const nonce = sig.nonce;
+    if (nonce == null) continue;
+    const key = [...noncePrefix, nonce] as unknown as KvKey;
+    const stored = await kv.get(key);
+    if (stored != null) {
+      // Consume the nonce (one-time use)
+      await kv.delete(key);
+      return true;
+    }
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 26c5725ad10bb2b094b52b8aabc2a12691554df6

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1679,68 @@ export async function respondWithObjectIfAcceptable(
...
+
+async function buildAcceptSignatureHeader(
+  policy: InboxChallengePolicy,
+  kv: KvStore,
+  noncePrefix: KvKey,
+): Promise<string> {
+  const params: AcceptSignatureMember["parameters"] = {};
+  if (policy.requestCreated !== false) {
+    params.created = true;
+  }
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in b6975afc466a9041ebbec3ed3f379d9c5386e11a

packages/fedify/src/sig/http.ts
```diff
@@ -1551,12 +1612,67 @@ export async function doubleKnock(
...
+          "Received Accept-Signature challenge; retrying with " +
+            "label {label} and components {components}.",
+          { label: rfc9421.label, components: rfc9421.components },
+        );
+        signedRequest = await signRequest(
+          request,
+          identity.privateKey,
+          identity.keyId,
+          { spec: "rfc9421", tracerProvider, body, rfc9421 },
+        );
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 24a3b2561694fcd306f65f0420ddece806943672

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1679,68 @@ export async function respondWithObjectIfAcceptable(
...
+  kv: KvStore,
+  noncePrefix: KvKey,
+): Promise<boolean> {
+  const signatureInput = request.headers.get("Signature-Input");
+  if (signatureInput == null) return false;
+  const parsed = parseRfc9421SignatureInput(signatureInput);
+  // Check each signature for a nonce
+  for (const sig of globalThis.Object.values(parsed)) {
+    const nonce = sig.nonce;
+    if (nonce == null) continue;
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 26c5725ad10bb2b094b52b8aabc2a12691554df6

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1679,68 @@ export async function respondWithObjectIfAcceptable(
...
+    const ttl = Temporal.Duration.from({
+      seconds: policy.nonceTtlSeconds ?? 300,
+    });
+    const key = [...noncePrefix, nonce] as unknown as KvKey;
+    await kv.set(key, true, { ttl });
+    params.nonce = nonce;
+  }
+  return formatAcceptSignature([{
+    label: "sig1",
+    components: policy.components ?? DEFAULT_CHALLENGE_COMPONENTS,
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 80f5ac8c3973c485a1c21cd5cc8b69eb709bd268

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1679,68 @@ export async function respondWithObjectIfAcceptable(
...
+  // Check each signature for a nonce
+  for (const sig of globalThis.Object.values(parsed)) {
+    const nonce = sig.nonce;
+    if (nonce == null) continue;
+    const key = [...noncePrefix, nonce] as unknown as KvKey;
+    const stored = await kv.get(key);
+    if (stored != null) {
+      // Consume the nonce (one-time use)
+      await kv.delete(key);
+      return true;
```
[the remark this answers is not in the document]
2chanhaeng: It doesn't look useful.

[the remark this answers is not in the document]
2chanhaeng: It doesn't look useful.

dahlia:
@codex review

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1680,84 @@ export async function respondWithObjectIfAcceptable(
...
+    const key = [...noncePrefix, nonce] as unknown as KvKey;
+    await kv.set(key, true, { ttl });
+    params.nonce = nonce;
+  }
+  return formatAcceptSignature([{
+    label: "sig1",
+    components: (policy.components ?? DEFAULT_CHALLENGE_COMPONENTS)
+      .filter((c) => c !== "@status"),
+    parameters: params,
+  }]);
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 4660cbba3573d35d9ca27b2fb3054911e7ca3421

packages/fedify/src/federation/handler.ts
```diff
@@ -797,6 +815,38 @@ async function handleInboxInternal<TContextData>(
...
+              kvPrefixes.acceptSignatureNonce,
+            );
+            headers["Cache-Control"] = "no-store";
+            headers["Vary"] = "Accept, Signature";
+            return new Response(
+              "Signature nonce verification failed.",
+              { status: 401, headers },
+            );
+          }
+        }
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 4660cbba3573d35d9ca27b2fb3054911e7ca3421

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1680,84 @@ export async function respondWithObjectIfAcceptable(
...
+    if (nonce == null) continue;
+    const key = [...noncePrefix, nonce] as unknown as KvKey;
+    const stored = await kv.get(key);
+    if (stored != null) {
+      // Consume the nonce (one-time use)
+      await kv.delete(key);
+      return true;
+    }
+  }
+  return false;
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 4660cbba3573d35d9ca27b2fb3054911e7ca3421

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1680,84 @@ export async function respondWithObjectIfAcceptable(
...
+    const sig = parsed[verifiedLabel];
+    if (sig == null) return false;
+    const nonce = sig.nonce;
+    if (nonce == null) return false;
+    const key = [...noncePrefix, nonce] as unknown as KvKey;
+    const stored = await kv.get(key);
+    if (stored != null) {
+      await kv.delete(key);
+      return true;
+    }
```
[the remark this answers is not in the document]
2chanhaeng: Similar to [this comment](https://github.com/fedify-dev/fedify/pull/626#discussion_r2940385110), the modification doesn't offer enough benefit to justify the added code complexity. So, rejected.

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1687,89 @@ export async function respondWithObjectIfAcceptable(
...
+  if (verifiedLabel == null) return false;
+  const sig = parsed[verifiedLabel];
+  if (sig == null) return false;
+  const nonce = sig.nonce;
+  if (nonce == null) return false;
+  const key = [...noncePrefix, nonce] as unknown as KvKey;
+  const stored = await kv.get(key);
+  if (stored != null) {
+    await kv.delete(key);
+    return true;
```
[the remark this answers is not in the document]
2chanhaeng: Read https://github.com/fedify-dev/fedify/pull/626#discussion_r2944496961.

packages/fedify/src/sig/accept.ts
```diff
@@ -0,0 +1,314 @@
...
+  if (
+    entry.parameters.keyid != null && entry.parameters.keyid !== localKeyId
+  ) {
+    return null;
+  }
+  return {
+    label: entry.label,
+    components: concatMinimumComponents(entry.components),
+    nonce: entry.parameters.nonce,
+    tag: entry.parameters.tag,
```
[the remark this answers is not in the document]
2chanhaeng: `expires` is a boolean flag requesting a timestamp, not a value.

packages/fedify/src/sig/http.ts
```diff
@@ -217,12 +262,22 @@ export interface Rfc9421SignatureParameters {
...
-  return `alg="${params.algorithm}";keyid="${params.keyId.href}";created=${params.created}`;
+  return Array.from(iterRfc9421(params)).join(";");
+}
+
+function* iterRfc9421(params: Rfc9421SignatureParameters): Iterable<string> {
+  yield `alg="${params.algorithm}"`;
+  yield `keyid="${params.keyId.href}"`;
+  yield `created=${params.created}`;
+  if (params.nonce != null) yield `nonce="${params.nonce}"`;
+  if (params.tag != null) yield `tag="${params.tag}"`;
```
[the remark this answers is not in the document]
2chanhaeng: No code path allows dangerous characters to reach the interpolation.

packages/fedify/src/sig/http.ts
```diff
@@ -297,11 +345,12 @@ export function formatRfc9421Signature(
...
   components: string[],
   parameters: string,
+  label = "sig1",
 ): [string, string] {
-  const signatureInputValue = `sig1=("${
+  const signatureInputValue = `${label}=("${
     components.join('" "')
   }");${parameters}`;
-  const signatureValue = `sig1=:${encodeBase64(signature)}:`;
+  const signatureValue = `${label}=:${encodeBase64(signature)}:`;
```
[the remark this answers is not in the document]
2chanhaeng: Eh, that's too much.

packages/fedify/src/sig/http.ts
```diff
@@ -237,55 +292,48 @@ export function createRfc9421SignatureBase(
...
 }
 
+const derivedComponents: Record<string, (request: Request) => string> = {
+  "@method": (request) => request.method.toUpperCase(),
+  "@target-uri": (request) => request.url,
+  "@authority": (request) => new URL(request.url).host,
+  "@scheme": (request) => new URL(request.url).protocol.slice(0, -1),
+  "@request-target": (request) => {
+    const url = new URL(request.url);
+    return `${request.method.toLowerCase()} ${url.pathname}${url.search}`;
```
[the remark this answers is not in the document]
2chanhaeng: This isn't even related to the PR issue, I just touch a little because I saw it. Somebody will do it someday.

packages/fedify/src/federation/handler.ts
```diff
@@ -1630,3 +1687,89 @@ export async function respondWithObjectIfAcceptable(
...
+  }
```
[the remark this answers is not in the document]
2chanhaeng: https://github.com/fedify-dev/fedify/pull/626#discussion_r2944496961

packages/fedify/src/federation/handler.ts
```diff
@@ -840,6 +868,35 @@ async function handleInboxInternal<TContextData>(
       headers: { "Content-Type": "text/plain; charset=utf-8" },
     });
   }
+  // Perform deferred nonce verification now that actor/key ownership is confirmed.
+  if (pendingNonceLabel !== null) {
+    const nonceValid = await verifySignatureNonce(
+      request,
+      kv,
+      kvPrefixes.acceptSignatureNonce,
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in cbea355e40d69e079976653994842fb32f104b94.

packages/fedify/src/federation/handler.ts
```diff
@@ -797,6 +818,13 @@ async function handleInboxInternal<TContextData>(
           },
         );
       } else {
+        if (
+          inboxChallengePolicy?.enabled && inboxChallengePolicy.requestNonce
+        ) {
+          // Defer nonce consumption until after actor/key ownership check to
+          // avoid burning nonces on requests that will be rejected anyway.
+          pendingNonceLabel = verification.signatureLabel;
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in cbea355e40d69e079976653994842fb32f104b94.

packages/fedify/src/federation/handler.ts
```diff
@@ -677,6 +686,9 @@ async function handleInboxInternal<TContextData>(
     }
   }
   let httpSigKey: CryptographicKey | null = null;
+  // Nonce verification is deferred until after actor/key ownership is checked
+  // to avoid consuming nonces on requests that will be rejected anyway.
+  let pendingNonceLabel: string | undefined | null = null;
```
dahlia: `pendingNonceLabel` is typed as `string | undefined | null` and initialized to `null`, but `verification.signatureLabel` is `string | undefined`. The three-way type makes the sentinel logic harder to follow. Consider initializing to `undefined` and using a `boolean` flag (e.g., `shouldVerifyNonce`) to make the intent clearer, or at least adding a comment explaining why `null` is the initial sentinel vs `undefined` from the verification result.

packages/fedify/src/sig/http.ts
```diff
@@ -237,55 +292,48 @@ export function createRfc9421SignatureBase(
...
+    const search = new URL(request.url).search;
+    return search.startsWith("?") ? search.slice(1) : search;
+  },
+  "@query-param": () => {
+    throw new Error("@query-param requires a parameter name");
+  },
+  "@status": () => {
+    throw new Error("@status is only valid for responses");
+  },
+};
```
dahlia: The original code parsed the URL once at the top of the function. This refactored lookup table calls `new URL(request.url)` independently in each handler (`@authority`, `@scheme`, `@request-target`, `@path`, `@query`). When a challenge requests multiple derived components (which is the common case—e.g., `@method`, `@target-uri`, `@authority`), the URL gets parsed repeatedly per signature base construction.

Consider passing a pre-parsed `URL` object to each handler, or lazily initializing it once.

packages/fedify/src/sig/http.ts
```diff
@@ -1551,12 +1613,76 @@ export async function doubleKnock(
...
+          );
+        }
+        fulfilled = true;
+        break;
+      }
+      // If the challenge retry succeeded, remember spec and return
+      if (
+        fulfilled && response.status < 300
+      ) {
+        await specDeterminer?.rememberSpec(origin, "rfc9421");
```
dahlia: If the challenge retry returns a non-negotiation failure (e.g., `403 Forbidden`), `fulfilled` is `true` but `response.status < 300` fails, causing a fall-through to the legacy spec-swap fallback. A `403 Forbidden` is not a signature negotiation issue, so retrying with draft-cavage seems incorrect in that case.

Consider refining this to only fall through to legacy retry on `400 Bad Request`/`401 Unauthorized`, and returning the challenge-retry response directly for other status codes.

packages/fedify/src/federation/handler.ts
dahlia, line 1706: The `get()` then `delete()` sequence is not atomic—between the two calls, a concurrent request with the same nonce could also pass `get()`, leading to a race condition where the nonce is consumed twice.

`KvStore` supports an optional `cas()` ([compare-and-swap]) method since 1.8.0. Using `kv.cas?.(key, true, undefined)` (or falling back to `get` + `delete` when `cas` is unavailable) would make this atomic where the store supports it:

```typescript
if (kv.cas != null) {
  return await kv.cas(key, true, undefined);
}
const stored = await kv.get(key);
if (stored != null) {
  await kv.delete(key);
  return true;
}
return false;
```

[compare-and-swap]: https://en.wikipedia.org/wiki/Compare-and-swap

packages/fedify/src/sig/accept.ts
```diff
@@ -0,0 +1,314 @@
...
+  evolve,
+  filter,
+  fromEntries,
+  isArray,
+  map,
+  pick,
+  pipe,
+  toArray,
+  uniq,
+} from "@fxts/core";
```
dahlia: This module imports heavily from `@fxts/core` (11 functions). The rest of the codebase (e.g., `http.ts` in this same PR) uses `es-toolkit` for utility functions. Mixing two utility libraries increases the overall dependency footprint, which is a concern for a framework like Fedify.

Consider using `es-toolkit` (or plain JS) here instead for consistency. Most of these operations (`pick`, `filter`, `map`, `entries`, `fromEntries`, `uniq`, `concat`) have direct equivalents in `es-toolkit` or are trivial with native JS.

CHANGES.md
```diff
@@ -34,8 +34,19 @@ To be released.
...
     GoToSocial that have authorized fetch enabled.  [[#473], [#589]]
 
+ -  Added RFC 9421 §5 `Accept-Signature` negotiation for both outbound and
+    inbound paths.  On the outbound side, `doubleKnock()` now parses
+    `Accept-Signature` challenges from `401` responses and retries with a
+    compatible RFC 9421 signature before falling back to legacy spec-swap.
+    On the inbound side, a new `InboxChallengePolicy` option in
+    `FederationOptions` enables emitting `Accept-Signature` headers on
+    inbox `401` responses, with optional one-time nonce support for replay
+    protection.  [[#583], [#584] by ChanHaeng Lee]
```
dahlia: The PR number (#626) should also be included in the reference markers alongside the issue numbers, per the project's changelog conventions (e.g., `[[#583], [#584], [#626] by ChanHaeng Lee]`).

packages/fedify/src/sig/accept.ts
```diff
@@ -0,0 +1,314 @@
...
+
+const parseEachSignature = (
+  dict: Dictionary,
+): IterableIterator<AcceptSignatureMember> =>
+  pipe(
+    dict,
+    entries,
+    filter(([_, item]) => isArray(item.value)),
+    map(([label, item]) =>
+      ({
```
dahlia: RFC compliance issue: component parameters are dropped.

[RFC 9421 §5.1] states:

> The list of component identifiers indicates the exact set of component identifiers to be included in the requested signature, *including all applicable component parameters.*

Component identifiers in the inner list can carry parameters like `;sf`, `;bs`, `;req`, `;key`, `;name`, `;tr` (e.g., `"content-type";sf`, `"@query-param";name="foo"`). Currently, `parseEachSignature` extracts only `subitem.value` and discards `subitem.params`. This means a challenge requesting `("@query-param";name="foo")` would lose the `;name` parameter and produce a malformed signature.

At minimum, the `components` field should preserve structured field item parameters, or `fulfillAcceptSignature` should reject entries containing parameterized components that it cannot handle.

[RFC 9421 §5.1]: https://www.rfc-editor.org/rfc/rfc9421#name-the-accept-signature-field

packages/fedify/src/sig/accept.ts
```diff
@@ -0,0 +1,314 @@
...
+export function fulfillAcceptSignature(
+  entry: AcceptSignatureMember,
+  localKeyId: string,
+  localAlg: string,
+): FulfillAcceptSignatureResult | null {
+  // Check algorithm compatibility
+  if (entry.parameters.alg != null && entry.parameters.alg !== localAlg) {
+    return null;
+  }
+  // Check key ID compatibility
```
dahlia, line 314: *RFC compliance concern: adding components not requested by the challenger.*

[RFC 9421 §5.2] states:

> By this process, a signature applied to a target message **MUST** have the same label, **MUST** include the *same set* of covered components, **MUST** process all requested parameters, and MAY have additional *parameters*.

Note the distinction: covered _components_ must be the same, while additional _parameters_ (metadata like `created`, `expires`) are allowed. `concatMinimumComponents` merges `@method`, `@target-uri`, `@authority` into the challenge's component list, which changes the covered component set.

The “MAY ignore any signature request that does not fit application parameters” clause allows _rejecting_ an entry entirely, but not _partially fulfilling_ it with a modified component set.

Suggestion: if the challenge's components don't include Fedify's minimum set, either fulfill it exactly as requested (trusting the challenger's intent), or reject it (return `null`) and let the fallback handle it.

[RFC 9421 §5.2]: https://www.rfc-editor.org/rfc/rfc9421#name-processing-an-accept-signat

packages/fedify/src/sig/accept.ts
```diff
@@ -0,0 +1,314 @@
...
+/**
+ * Attempts to translate an {@link AcceptSignatureMember} challenge into
+ * RFC 9421 signing options that the local signer can fulfill.
+ *
+ * Returns `null` if the challenge cannot be fulfilled—for example, if
+ * the requested `alg` or `keyid` is incompatible with the local key.
+ *
+ * Safety constraints:
+ * - `alg`: only honored if it matches `localAlg`.
+ * - `keyid`: only honored if it matches `localKeyId`.
```
dahlia, line 294: *RFC compliance issue: `expires` parameter is silently ignored.*

[RFC 9421 §5.1] defines `expires` as a requestable parameter:

> `expires`: The signer is requested to generate and include an expiration time. This parameter has no associated value when sent as a signature request.

And [§5.2] says the resulting signature “**MUST** process all requested parameters.”

`FulfillAcceptSignatureResult` only includes `nonce` and `tag`, but not `expires` (or `created`, though `created` is always generated by `signRequestRfc9421`). If a challenge requests `expires`, the signer should generate and include an expiration timestamp—currently this is silently dropped.

Consider adding an `expires` field to `FulfillAcceptSignatureResult` and `Rfc9421SignRequestOptions`, and generating an expiration timestamp in `signRequestRfc9421` when requested.

[RFC 9421 §5.1]: https://www.rfc-editor.org/rfc/rfc9421#name-the-accept-signature-field
[§5.2]: https://www.rfc-editor.org/rfc/rfc9421#name-processing-an-accept-signat

packages/fedify/src/sig/http.ts
```diff
@@ -1551,12 +1613,76 @@ export async function doubleKnock(
...
+        ) {
+          const location = response.headers.get("Location")!;
+          return doubleKnock(
+            createRedirectRequest(request, location, body),
+            identity,
+            { ...options, body },
+          );
+        }
+        fulfilled = true;
+        break;
```
dahlia: *RFC compliance concern: only the first matching entry is fulfilled.*

[RFC 9421 §5] states:

> The target message of an `Accept-Signature` field **MUST** include *all* labeled signatures indicated in the `Accept-Signature` field.

And [§5.2]:

> The receiver of an `Accept-Signature` field **MAY** ignore any signature request that does not fit application parameters.

The current loop fulfills the first compatible entry, then breaks. If the `Accept-Signature` header contains multiple entries (e.g., `sig1` and `sig2`), only one signature is produced.

The MAY-ignore clause allows _skipping_ incompatible entries, but compatible entries should ideally all be fulfilled. That said, `signRequest()` currently only supports producing a single signature per call (it uses `headers.set()` not `headers.append()`), so supporting multiple signatures would require a larger refactor.

At minimum, this limitation should be documented. Longer term, consider supporting multi-signature production when the challenge requests it.

[RFC 9421 §5]: https://www.rfc-editor.org/rfc/rfc9421#name-requesting-signatures
[§5.2]: https://www.rfc-editor.org/rfc/rfc9421#name-processing-an-accept-signat

CHANGES.md
2chanhaeng: Fixed in 623204bd1374d5b0f1cca6fa5e6169ad738860c6!

packages/fedify/src/federation/handler.ts
2chanhaeng: Fixed in 83fb7398f353dc2b0ccece406aae4bd28f66d14b!

packages/fedify/src/federation/handler.ts
2chanhaeng: Fixed in 314c519fbfc427b6593b8a0c2d25d4591e371362!

packages/fedify/src/sig/accept.ts
2chanhaeng: Fixed in 4042e7da924c90c9eb1575b06b5302c9c78bec30!

I've removed `@fxts/core` for now, but the reason I chose it over `es-toolkit` in the first place was because its type inference is more accurate. You can see the difference immediately just by comparing `flow` from `es-toolkit` with `pipe` from `@fxts/core`.

packages/fedify/src/sig/accept.ts
2chanhaeng: Fixed in 4042e7da924c90c9eb1575b06b5302c9c78bec30!

I misunderstood some parts of RFC 9421 because sync with the existing code. Like, `Rfc9421SignRequestOptions` or something. 😂

packages/fedify/src/sig/accept.ts
2chanhaeng: Fixed in 470323d6374b40f8529c6c38c7e9b8d80c609918!

packages/fedify/src/sig/accept.ts
2chanhaeng: Fixed in ca50dea19af0c6d2ae16acf91634281f9c8e4945!

I've changed that fulfill the challenge exactly as requested. I think that users might find it a bit complicated about the minimum set.

packages/fedify/src/sig/http.ts
2chanhaeng: Fixed in d1f9eab5fd797584bd0da38a3eb4118119a95dc7!

I wrote the original code that way because I don't like passing values as separate arguments when they can already be calculated from existing ones. However, this version is definitely much more readable and saner. 😂

packages/fedify/src/sig/http.ts
2chanhaeng: Fixed in ee19f0bb3a5b443c8c4fd2184c3820a9b6bd5a21!

packages/fedify/src/sig/http.ts
2chanhaeng: Fixed in ab6efec82f63a44edd1c576e4a0ded0cbdf43e60!

I didn't think it would be that difficult, so I applied it. What do you think? Did I missed something?

packages/fedify/src/sig/http.ts
```diff
@@ -1551,12 +1642,93 @@ export async function doubleKnock(
...
+            identity.privateKey,
+            identity.keyId,
+            {
+              spec: "rfc9421",
+              tracerProvider,
+              body,
+              rfc9421: {
+                ...rfc9421,
+                components: rfc9421.components.map((c) => c.value),
+              },
```
dahlia: Preserve the full challenged covered components here instead of collapsing them to `c.value`. `fulfillAcceptSignature()` keeps component parameters like `;name="foo"` and `;sf`, but `components.map((c) => c.value)` drops them before the retry is signed. That changes the covered component set (or throws for bare `@query-param`) and breaks RFC 9421 challenge negotiation for valid `Accept-Signature` challenges that rely on component parameters.

packages/fedify/src/sig/http.ts
```diff
@@ -217,12 +269,24 @@ export interface Rfc9421SignatureParameters {
...
+  return Array.from(iterRfc9421(params)).join(";");
+}
+
+function* iterRfc9421(params: Rfc9421SignatureParameters): Iterable<string> {
+  yield `alg="${params.algorithm}"`;
+  yield `keyid="${params.keyId.href}"`;
+  yield `created=${params.created}`;
+  if (params.expires != null) yield `expires=${params.expires}`;
+  if (params.nonce != null) yield `nonce="${params.nonce}"`;
+  if (params.tag != null) yield `tag="${params.tag}"`;
```
dahlia: `nonce` and `tag` are interpolated directly into `Signature-Input` here without structured-field escaping. RFC 8941 strings can legitimately contain escaped `\"` or `\\`, so a challenge such as `;tag="a\"b"` or `;nonce="x\\y"` produces an invalid retry header even though the `Accept-Signature` challenge itself is valid. These parameters need structured-field serialization, or equivalent escaping, before being emitted.

packages/fedify/src/federation/handler.ts
```diff
@@ -797,6 +819,13 @@ async function handleInboxInternal<TContextData>(
           },
```
dahlia: When `inboxChallengePolicy` is enabled, other inbox 401 paths already advertise the retry shape with `Accept-Signature` plus `Cache-Control: no-store`, but this fallback after `unverifiedActivityHandler` returns `undefined` still emits a plain `401`. That makes the new challenge-response behavior inconsistent across verification-failure paths and creates an interoperability gap: senders that hit this branch do not learn what RFC 9421 signature shape they should retry with.

2chanhaeng: Fixed in d90f4a5576af8697da2d886c763814c5e727c68c!

packages/fedify/src/sig/http.ts
2chanhaeng: Fixed in 95b6eccc33142ab96fd23ce7368b0f77d40c638c!

packages/fedify/src/sig/http.ts
2chanhaeng: Fixed in 71fdcae4ff1188d536db4d8d9146444cef6da272!

examples/rfc-9421-test/README.md
```diff
@@ -0,0 +1,132 @@
+RFC 9421 Interoperability Field Test
+=====================================
+
+A Fedify-based server for testing RFC 9421 HTTP Message Signatures
+interoperability with Bonfire, Mastodon, and other fediverse implementations.
+
+See [../../plans/field-test.md](../../plans/field-test.md) for the full test
+plan.
```
dahlia: I think it's a dangling pointer?

examples/rfc-9421-test/tunnel.ts
```diff
@@ -0,0 +1,85 @@
...
+import { getLogger } from "@logtape/logtape";
+
+/**
+ * Starts `fedify tunnel -s pinggy.io <port>` and waits up to `timeoutMs`
+ * for the tunnel URL to appear in its output.  The tunnel process is kept
+ * alive and returned to the caller; it must be killed when no longer needed.
+ *
+ * Returns `null` if the URL was not found before the timeout.
+ */
+export default async function startTunnel(
```
dahlia, line 11: FYI, there's the [@hongminhee/localtunnel](https://github.com/dahlia/localtunnel) package, which powers `fedify tunnel`.

2chanhaeng: Oh, good! I copied it from [examples/test-examples](https://github.com/fedify-dev/fedify/tree/main/examples/test-examples/mod.ts).

2chanhaeng:
Added tags about signature spec in log!

dahlia (approved):
I've just conducted a field test against Bonfire myself, and it apparently works perfectly! Great job!

Issue #627, opened 2026-03-17
Refactor preloaded contexts to per-context static JSON assets


dahlia:
## Summary
This issue is a follow-up from PR #622, where https://github.com/fedify-dev/fedify/pull/622#discussion_r2944685365 suggested moving preloaded context definitions out of TypeScript implementation code, and the change was noted as a good candidate for a separate task.

## Goal
Refactor context data loading so each context is stored as an individual static JSON asset instead of being bundled together in one large in-code definition, while keeping runtime behavior and public API behavior stable.

## Proposed approach
Create one JSON file per context under *packages/vocab-runtime/src/contexts/* and load them through JSON imports in *packages/vocab-runtime/src/contexts.ts* using the native `import ... with { type: "json" };` syntax, then normalize them into the existing internal runtime structure.

## Data layout
Store context definitions separately, for example *packages/vocab-runtime/src/contexts/activitystreams.json* and *packages/vocab-runtime/src/contexts/gotosocial.json*, and build the registry/index from these files so each context can be updated independently.

## Why this helps
A per-context file layout reduces merge conflicts, narrows review scope to only the changed context, and makes future updates and regeneration easier because data-only modifications stay isolated from runtime code.

## Acceptance criteria
The refactor is complete when the effective behavior of context lookup and rendering is unchanged, tests pass across Deno/Node/Bun, exported API contracts remain unchanged, and updating a single context does not require editing unrelated context data in shared code.

## References
PR discussion: https://github.com/fedify-dev/fedify/pull/622#discussion_r2944685365

PR #641, opened 2026-03-23
Refactor preloaded contexts into static JSON assets

dahlia:
Summary
-------

This PR refactors the preloaded JSON-LD contexts in *@fedify/vocab-runtime* so that each context is stored as its own static JSON asset instead of being embedded together in one large TypeScript object.

Details
-------

Each preloaded context now lives under *packages/vocab-runtime/src/contexts/*, and *packages/vocab-runtime/src/contexts.ts* rebuilds the existing `preloadedContexts` registry through static JSON imports while preserving the same URL keys, exported API shape, and runtime behavior.

Rationale
---------

This keeps data-only context updates isolated from runtime code, reduces merge conflicts when multiple contexts change independently, and makes context changes easier to review because unrelated definitions no longer live in the same source block.

Testing
-------

I verified the change with `deno task -f @fedify/vocab-runtime test`, `pnpm --filter @fedify/vocab-runtime test`, `deno test --check --allow-all packages/fedify/src/utils/kv-cache.test.ts`, `node --test utils/kv-cache.test.js` from *packages/fedify/dist*, and `bun test utils/kv-cache.test.js --timeout 60000` from *packages/fedify/dist*.

Notes
-----

This is an internal refactor with no user-visible behavior change, so *CHANGES.md* was intentionally not updated.

Related issue
-------------

Fixes https://github.com/fedify-dev/fedify/issues/627


packages/vocab-runtime/src/contexts/gotosocial.json
```diff
@@ -0,0 +1,86 @@
+{
+  "@context": {
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "gts": "https://gotosocial.org/ns#",
+    "LikeRequest": "gts:LikeRequest",
+    "ReplyRequest": "gts:ReplyRequest",
+    "AnnounceRequest": "gts:AnnounceRequest",
+    "QuoteRequest": "gts:QuoteRequest",
+    "LikeAuthorization": "gts:LikeApproval",
```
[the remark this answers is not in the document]
dahlia: No, it's intended.

packages/vocab-runtime/src/contexts.ts
2chanhaeng, line 1: How about move this file to `contexts/mod.ts`? Just a thought.

dahlia: Thanks for the suggestion. I kept *contexts.ts* in this PR to minimize churn and keep the refactor scoped to moving the data into per-context JSON assets. I think *contexts/mod.ts* could be a reasonable follow-up cleanup, but it does not materially affect this refactor, so I'd prefer to leave it as-is here.

PR #674, opened 2026-04-13
Add `@fedify/nuxt` package for Nuxt integration

### LINKED ISSUE #149: Nuxt integration
It would be nice to have the official integration with [Nuxt](https://nuxt.com/)!

2chanhaeng:
Closes #149

> [!TIP]
> Of the 4,315 lines of additions, 3,349 lines are *pnpm-lock.yaml*
> changes.  The actual code to review is under 1,000 lines.

Changes
-------

 -  Added `@fedify/nuxt` package that provides a Nuxt module integrating
    Fedify into Nitro middleware with deferred `406 Not Acceptable`
    handling for content negotiation.
 -  Added Nuxt section to *docs/manual/integration.md*.
 -  Updated workspace configs, package listings, and *CHANGES.md*.
 -  Updated agent skill docs with clarified guidance (non-functional).

---

Assisted-by: GPT-5.3-codex


packages/nuxt/src/runtime/server/middleware.ts
```diff
@@ -0,0 +1,46 @@
...
+      "@fedify/nuxt: Federation instance is missing. " +
+        "Export default Federation (or named 'federation') from the configured module.",
+    );
+  }
+}
+
+export function createFedifyMiddleware(
+  federation: unknown,
+  contextDataFactory?: (event: unknown, request: Request) => unknown,
+) {
```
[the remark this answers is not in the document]
2chanhaeng: Won't fix — `createFedifyMiddleware` is only called from a virtual template auto-generated by the Nuxt module, not by end users. It is not part of the public API surface (`src/mod.ts`), so JSDoc is unnecessary here.

packages/nuxt/src/runtime/server/logic.ts
```diff
@@ -0,0 +1,53 @@
...
+  return { kind: "handled", response };
+}
+
+export function resolveDeferredNotAcceptable(
+  isDeferred: boolean,
+  frameworkStatus: number,
+): Response | undefined {
+  if (!isDeferred || frameworkStatus !== 404) return undefined;
+  return createNotAcceptableResponse();
+}
```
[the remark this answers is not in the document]
2chanhaeng: Won't fix — same reason as above. `resolveDeferredNotAcceptable` is an internal helper not exposed in the public API. No other integration package documents internal helpers with JSDoc.

packages/nuxt/src/runtime/server/logic.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (response === DUMMY_NOT_FOUND_RESPONSE) {
+    return { kind: "not-found" };
+  }
+
+  if (response.status === 406) {
+    return { kind: "not-acceptable" };
+  }
+
+  return { kind: "handled", response };
+}
```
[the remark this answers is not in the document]
2chanhaeng: Won't fix — `fetchWithFedify` is an internal runtime helper inside `src/runtime/server/` and is not re-exported from the package's public API (`src/mod.ts`). Other integration packages (fastify, hono, h3, express) follow the same pattern without JSDoc on internal helpers.

packages/nuxt/src/mod.ts
```diff
@@ -0,0 +1,115 @@
+import {
+  addServerHandler,
+  addServerPlugin,
+  addTemplate,
+  createResolver,
+  defineNuxtModule,
+  resolveAlias,
+} from "@nuxt/kit";
```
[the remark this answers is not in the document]
2chanhaeng: Wrong. `Nuxt` type declares only internal, not exported.

.agents/skills/add-to-fedify-init/SKILL.md
```diff
@@ -34,9 +34,9 @@ modifications consistent with the existing code style and context.
 ### Write the `WebFrameworkDescription` object
 
 Create a `packages/init/src/webframeworks/framework.ts` file and write the
-`WebFrameworkDescription` object, referring to <init/framework.ts>.  Check
-the specifications in the comments in `packages/init/src/types.ts` for
-details.
+`WebFrameworkDescription` object, referring to
+[init/framework.ts](./init/framework.ts).  Check the specifications in the
+comments in `packages/init/src/types.ts` for details.
```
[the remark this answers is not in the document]
2chanhaeng: This file is an internal agent skill document, not user-facing documentation. The inline link style here is perfectly discoverable via `grep` and doesn't warrant the overhead of reference-style links. Won't fix.

packages/nuxt/src/runtime/server/plugin.ts
```diff
@@ -0,0 +1,48 @@
...
+        getResponseStatus(event),
+      );
+      if (negotiatedResponse == null) return;
+
+      setResponseStatus(event, negotiatedResponse.status);
+      negotiatedResponse.headers.forEach((value: string, key: string) => {
+        setResponseHeader(event, key, value);
+      });
+
+      payload.body = NOT_ACCEPTABLE_BODY;
```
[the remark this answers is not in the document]
2chanhaeng: Unlike Express (which exposes `req.route` to distinguish "route matched but returned 404" from "no route found"), Nuxt/Nitro does not provide a route-miss signal. There is no reliable way to differentiate between an intentional 404 from a matched route and a 404 from no matching route at the `beforeResponse` hook level.

This is not unique to the Nuxt integration — other Fedify integration packages (h3, Hono, SvelteKit, Fastify, Elysia) all use the same status-code-based approach. Won't fix.

.agents/skills/create-integration-package/SKILL.md
```diff
@@ -167,14 +247,19 @@ Add keywords related to the framework in `.hongdown.toml` and `cspell.json` in
 root path. Especially, the package name `@fedify/framework` should be added to
 the `.hongdown.toml`.
 
-After implementation, run `mise run fmt && mise check`.
+After implementation, run `mise run fmt && mise run check`.
```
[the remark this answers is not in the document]
2chanhaeng: `run` is an optional subcommand in mise that simply makes explicit that the following argument is a task name. `mise run check` and `mise check` are equivalent. Not an issue.

packages/nuxt/src/mod.ts
```diff
@@ -0,0 +1,115 @@
...
+            );
+          }
+
+          const contextFactoryResolver = contextDataFactoryModule == null
+            ? "const contextDataFactory = undefined;"
+            : [
+              "const contextDataFactory =",
+              "  contextFactoryModule.default ??",
+              "  contextFactoryModule.contextDataFactory;",
+            ].join("\n");
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in f5b22a1e4b4378818b61fa820802802b8817f55b

.agents/skills/create-example-app-with-integration/example/README.md
```diff
@@ -2,17 +2,19 @@
 
 <!-- Replace `프레임워크` with the name of the framework you are integrating with -->
 
-프레임워크 example application
-==============================
+Fedify-프레임워크 integration example application
```
dahlia: [En dash!](https://www.thepunctuationguide.com/en-dash.html)

```suggestion
Fedify–프레임워크 integration example application
```

.agents/skills/create-integration-package/SKILL.md
```diff
@@ -67,6 +93,34 @@ When a request arrives, the integration middleware calls
 `application/activity+json`, Fedify generates and returns the JSON-LD
 response directly.  Framework-side routing does not execute.
 
+#### Fallthrough on 404 (not found)
```
dahlia:
```suggestion
#### Fallthrough on `404 Not Found`
```

.agents/skills/create-integration-package/SKILL.md
```diff
@@ -67,6 +93,34 @@ When a request arrives, the integration middleware calls
...
+
+When `federation.fetch()` does not recognize the path, it invokes the
+`onNotFound` callback.  The integration should delegate to the framework's
+next handler so that non-federation routes (HTML pages, API endpoints,
+etc.) are served normally.
+
+#### Deferred 406 (not acceptable)
```
dahlia:
```suggestion
#### Deferred `406 Not Acceptable`
```

CHANGES.md
```diff
@@ -47,6 +47,19 @@ To be released.
 [#601]: https://github.com/fedify-dev/fedify/pull/601
 [#652]: https://github.com/fedify-dev/fedify/pull/652
 
+### @fedify/nuxt
+
+ -  Added `@fedify/nuxt` package for integrating Fedify with [Nuxt].
+    It provides a Nuxt module that delegates non-federation requests to Nuxt,
+    supports shared-route content negotiation, and returns deferred
+    `406 Not acceptable` when Fedify routes are requested without
```
```suggestion
    `406 Not Acceptable` when Fedify routes are requested without
```

packages/nuxt/src/runtime/server/logic.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export function resolveDeferredNotAcceptable(
+  isDeferred: boolean,
+  frameworkStatus: number,
+): Response | undefined {
+  if (!isDeferred || frameworkStatus !== 404) return undefined;
```
dahlia, line 57: This helper currently assumes that `isDeferred && frameworkStatus === 404` means “Fedify matched the path, Nuxt had no page for it”, but those two facts are not equivalent on shared routes.

For a browser request to a shared route like `/users/[handle]`, Fedify can call `onNotAcceptable` before the application has determined whether the underlying actor or object actually exists. That means the deferred flag is set even when the Nuxt page *did* match and then intentionally returned its own `404 Not Found` for a missing resource.

In that case, this helper rewrites the page's genuine 404 into the generic `406 Not Acceptable` fallback, which changes the semantics of matched HTML routes. A missing actor page should remain the page's real 404 response, not get converted into a negotiation failure.

packages/nuxt/src/runtime/server/logic.ts
```diff
@@ -0,0 +1,53 @@
...
+    contextData,
+    onNotFound: () => DUMMY_NOT_FOUND_RESPONSE,
+    onNotAcceptable: createNotAcceptableResponse,
+  });
+
+  if (response === DUMMY_NOT_FOUND_RESPONSE) {
+    return { kind: "not-found" };
+  }
+
+  if (response.status === 406) {
```
dahlia: This `response.status === 406` check conflates the synthetic response produced by `onNotAcceptable` with any real `406` returned by Fedify handlers themselves.

`Federation.fetch()` can surface user-provided `Response` objects from extension points such as object, collection, or inbox handlers. If one of those handlers intentionally returns `406` with its own explanatory body or headers, this branch classifies it as the deferred negotiation case and drops the original response. The plugin then later replaces it with the generic plain-text fallback.

So the problem here is not “406 exists”, but “406 came from our synthetic `onNotAcceptable` sentinel”. We need to distinguish those two cases instead of keying only on the numeric status code.

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,124 @@
...
+    },
+    setup(options: ModuleOptions, nuxt: Nuxt) {
+      const resolver = createResolver(import.meta.url);
+      const federationModule = resolveAlias(
+        options.federationModule,
+        nuxt.options.alias,
+      );
+      const contextDataFactoryModule = options.contextDataFactoryModule == null
+        ? undefined
+        : resolveAlias(options.contextDataFactoryModule, nuxt.options.alias);
```
dahlia: `resolveAlias()` only expands aliases like `~` and `@`; it does not anchor plain relative specifiers against the Nuxt app root.

Because these strings are written into `.nuxt/fedify-nuxt-options.mjs`, a config such as `federationModule: './server/federation'` or `contextDataFactoryModule: './server/fedify-context'` ends up being imported relative to `.nuxt/`, not the project root, and fails with `ERR_MODULE_NOT_FOUND` at startup.

So aliased paths work here, but non-aliased relative module paths are currently broken. These options need to be normalized against the Nuxt root directory before they are serialized into the generated template.

dahlia (changes requested):

dahlia:
@codex review

.agents/skills/create-example-app-with-integration/example/README.md
2chanhaeng: Fixed in 1ec06ab92c91d40aa8fe68a5743c013b84622ca9!
Always the dash... 😂

.agents/skills/create-integration-package/SKILL.md
2chanhaeng: Fixed in 1ec06ab92c91d40aa8fe68a5743c013b84622ca9!

.agents/skills/create-integration-package/SKILL.md
2chanhaeng: Fixed in 1ec06ab92c91d40aa8fe68a5743c013b84622ca9!

CHANGES.md
2chanhaeng: Fixed in 1ec06ab92c91d40aa8fe68a5743c013b84622ca9!

packages/nuxt/src/runtime/server/logic.ts
2chanhaeng: Fixed in 6cddda7d508b5cdd64abea1fc11b990b10a266fe!
I blocked it by adding a flag argument to the third parameter of `resolveDeferredNotAcceptable` to ensure that `matchedRoute` is not nullable.

packages/nuxt/src/runtime/server/logic.ts
2chanhaeng: Fixed in 6cddda7d508b5cdd64abea1fc11b990b10a266fe!
I modified the logic to store the value returned by `onNotAcceptable` in a variable within the parent `fetchWithFedify` scope, and then compare the two values to check if it is the `Not Acceptable` handled by the `fetcher`.

packages/nuxt/src/module.ts
2chanhaeng: Fixed in 6d98b887ca4c0930d07ff690f4f059e1503cb205!
I defined and applied a `resolveModulePath` function that utilizes `resolveAlias`. While creating an example app, an error occured related to path. Maybe caused by this part?

packages/nuxt/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "@fedify/fedify": "workspace:^",
+    "@nuxt/kit": "^4.4.0",
+    "h3": "catalog:",
+    "nuxt": "catalog:"
+  },
+  "devDependencies": {
+    "@types/node": "catalog:",
+    "@nuxt/schema": "^4.4.0",
+    "tsdown": "catalog:",
+    "typescript": "catalog:"
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 37c57d0a7b962273c538faa8d37efd7a761cb871

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,124 @@
...
+  contextDataFactoryModule: string | null,
+): string {
+  if (contextDataFactoryModule == null) {
+    return "const contextDataFactory = undefined;";
+  }
+  return `
+const contextDataFactory =
+  contextFactoryModule.default ??
+  contextFactoryModule.contextDataFactory;
+if (contextDataFactory != null && typeof contextDataFactory !== 'function') {
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 12d3d772c53165f423cfb7b037be15b41f2090e8
Adding an error message for null export.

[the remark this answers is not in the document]
2chanhaeng: Fixed in 12d3d772c53165f423cfb7b037be15b41f2090e8
Adding an error message for null export.

2chanhaeng:
@coderabbitai review

2chanhaeng:
@coderabbitai review

docs/manual/integration.md
```diff
@@ -359,6 +361,76 @@ app.use(router);
...
+
+By default, _@fedify/nuxt_ loads the federation instance from
+*~/server/federation*. You can customize module paths using `fedify` options:
+
+~~~~ typescript
+export default defineNuxtConfig({
+  modules: ["@fedify/nuxt"],
+  fedify: {
+    federationModule: "~/server/federation",
+    contextDataFactoryModule: "~/server/fedify-context",
```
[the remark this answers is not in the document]
2chanhaeng: The right path is `#server`. Check the [document](https://nuxt.com/docs/4.x/api/nuxt-config#alias).

packages/nuxt/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+If your project uses a different file path or context data factory,
+configure the module options:
+
+~~~~ typescript
+export default defineNuxtConfig({
+  modules: ["@fedify/nuxt"],
+  fedify: {
+    federationModule: "~/server/federation",
+    contextDataFactoryModule: "~/server/fedify-context",
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 0986264afd623ea432f423a0672684a19f0fbff1

docs/manual/integration.md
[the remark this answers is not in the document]
2chanhaeng: Fixed in 0986264afd623ea432f423a0672684a19f0fbff1

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export function resolveModulePath(
+  modulePath: string,
+  aliases: Record<string, string>,
+  rootDir: string,
+): string {
+  const resolved = resolveAlias(modulePath, aliases);
+  if (isAbsolute(resolved)) return resolved;
+  return resolve(rootDir, resolved);
+}
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 4b23e69e264ffe7e09c7f3949e371810b6eece85

.agents/skills/commit/SKILL.md
```diff
@@ -42,8 +42,8 @@ EOF
 
 Don't put `Generated with [Claude Code](https://claude.ai/code)`, or
 `Co-authored-by` trailers at the end of the commit message.  Instead,
-use the `Assiged-by` trailer to indicate that the commit was generated by
-an AI assistant, if necessary.  The format of the `Assiged-by` trailer should
+use the `Assigned-by` trailer to indicate that the commit was generated by
+an AI assistant, if necessary.  The format of the `Assigned-by` trailer should
```
dahlia: Oh, these should be `Assisted-by`, neither `Assigned-by` nor `Assiged-by`! It's my fault though.

dahlia (changes requested):

.agents/skills/commit/SKILL.md
2chanhaeng: Fixed in 30ddd9f13a2783351a703a779308d322e3e01760!
Ah yeah, right. AI is supposed to assist and humans assign, but it almost ended up the other way around with humans assisting and AI assigning. 😂

dahlia (changes requested):
Just one more thing: shouldn't we update the table in the *README.md* in the root?

2chanhaeng:
> Just one more thing: shouldn't we update the table in the _README.md_ in the root?

[Already did!](https://github.com/fedify-dev/fedify/pull/674/changes#diff-a0bafe998b33e1f1406b0198fa2a72d9a43d209a73c4a0d99002ffa899a462f5)

dahlia (approved):

PR #698, opened 2026-04-19
Allow private addresses for explicit URLs in fedify lookup

### LINKED ISSUE #696: `fedfiy lookup` is not working without `-a` option in 2.1.x
## Summary

Since v2.1, `fedify lookup` no longer works locally without the `-a` flag.

## Steps to Reproduce

1. Run `fedify lookup <actor-uri>` locally (without `-a`)
2. Observe that the command fails

## Expected Behavior

`fedify lookup` should work without `-a` for local usage, as it did before v2.1.

## Actual Behavior

The command does not execute unless `-a` is explicitly provided.

## Environment

- Fedify version: 2.1.x

## Test

Tested with `examples/hono-sample`.

Image

2chanhaeng:
Closes #696.

Since v2.1, `fedify lookup` rejected localhost URLs unless `-p`/`--allow-private-address` was passed, because the CLI began forwarding `allowPrivateAddress=false` to the vocab-runtime document loader, whose `validatePublicUrl` check blocks loopback addresses.

Split the document/auth loaders into two:

 -  An "initial" loader that always allows private addresses, used for URLs explicitly provided on the command line (plain lookup, the first fetch in `--traverse`, and the first fetch in `--recurse`).
 -  The existing loader, which continues to honor `--allow-private-address`, used for URLs discovered from remote responses (traversal pages and recursion targets).

This preserves SSRF protection against `http://localhost/...` URLs embedded in remote `first`/`next`/`inReplyTo` fields while letting users look up local servers without extra flags.

Assisted-by: Claude Code:claude-opus-4-7

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,9 +15,10 @@ To be released.
     regression introduced in Fedify 2.1.0 when the CLI began forwarding
     the `allowPrivateAddress` option to the underlying document loader.
     URLs explicitly provided on the command line now always allow private
-    addresses, while URLs discovered via [`-t`/`--traverse`] or [`--recurse`]
-    still honor the option to mitigate SSRF attacks against private
-    addresses.  [[#696], [#698] by Chanhaeng Lee]
+    addresses, while URLs discovered during [`-t`/`--traverse`] honor the
+    option to mitigate SSRF attacks against private addresses.  Recursive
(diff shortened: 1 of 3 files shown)


2chanhaeng:
@coderabbitai review

dahlia (approved):

PR #726, opened 2026-04-25
Wire Nuxt init logging

### LINKED ISSUE #724: `fedify init -w nuxt` scaffolds server/logging.ts but never imports it, leaving Fedify silent
## Summary

Running `fedify init -w nuxt -p npm -k in-memory -m in-process` produces a project where *server/logging.ts* exists with a complete LogTape `configure({...})` call, but no other file ever imports it. Other framework templates (Hono, Express, Elysia, bare-bones) include `import "./logging.ts"` from their entry point, so this looks like an oversight specific to the Nuxt template.

The visible symptom is that `getLogger("fedify")` and `getLogger("<your-package>")` produce no output. Federation-time bugs (HTTP signature failures, inbox listener errors, queued sends) are silent during development, which makes interop bugs against Mastodon, Pixelfed, etc. very hard to debug.

## Reproduction

```sh
fedify init -w nuxt -p npm -k in-memory -m in-process my-app
cd my-app
grep -r "logging" .
# server/logging.ts is the only hit; no `import "./logging.ts"` anywhere
npm run dev
# trigger a federation event, e.g. fedify lookup http://localhost:3000/users/foo
# expected: Fedify logs to stdout
# actual: no Fedify logs appear
```

## Expected behavior

Same as the Hono/Express/etc. templates: *server/logging.ts* should be wired into the runtime so its `configure(...)` runs at server startup and Fedify's `getLogger(...)` calls produce output by default.

## Suggested fix

The Nuxt template should add a Nitro server plugin that imports the configuration. Roughly:

```ts
// packages/init/src/templates/nuxt/server/plugins/logging.ts.tpl
import { loggingConfigured } from "../logging";

export default defineNitroPlugin(async () => {
  await loggingConfigured;
});
```

…with *server/logging.ts.tpl* re-exporting the `Promise<void>` returned by `configure(...)` instead of using top-level `await` (Nitro's default Vite target is `es2019`, which rejects top-level await):

```ts
// packages/init/src/templates/nuxt/server/logging.ts.tpl
import { AsyncLocalStorage } from "node:async_hooks";
import { configure, getConsoleSink } from "@logtape/logtape";

export const loggingConfigured: Promise<void> = configure({
  contextLocalStorage: new AsyncLocalStorage(),
  sinks: { console: getConsoleSink() },
  filters: {},
  loggers: [
    { category: "{{packageName}}", lowestLevel: "debug", sinks: ["console"] },
    { category: "fedify", lowestLevel: "info", sinks: ["console"] },
    { category: ["logtape", "meta"], lowestLevel: "warning", sinks: ["console"] },
  ],
});
```

(Alternative: keep top-level await but raise the Nitro/Vite ES target. The plugin approach is closer to what other Nuxt projects look like, and it preserves compatibility with the default ES target.)

## Environment

- `@fedify/cli` 2.2.0-dev.938 (npm install via mise)
- Node.js 22.22.2
- Nuxt 4.4.2, Nitro 2.13.3
- Discovered while writing the content-sharing tutorial (Fedify issue #693)

dahlia:
## Summary

This fixes #724 by wiring the logging configuration generated by `fedify init -w nuxt` into the generated Nuxt/Nitro runtime.

Previously, *server/logging.ts* was generated but never imported, so LogTape configuration did not run and Fedify logs stayed silent in generated Nuxt apps. The Nuxt template now generates a Nitro server plugin at *server/plugins/logging.ts* that imports and awaits the logging setup during Nitro startup.

## Changes

- Added a Nuxt-specific logging template at *packages/init/src/templates/nuxt/server/logging.ts.tpl*.
- Changed that template to default-export the `configure(...)` promise instead of using top-level `await`.
- Added *packages/init/src/templates/nuxt/server/plugins/logging.ts.tpl* to await the logging setup from a Nitro plugin.
- Allowed framework initializers to override the default logging template.
- Added a regression test that verifies generated Nuxt projects include the logging plugin and avoid top-level `await configure(...)`.

## Verification

- `deno test -A packages/init/src/action/configs.test.ts`
- `deno task -f @fedify/init check`
- `mise run test:init -- -w nuxt -p npm -k in-memory -m in-process --no-dry-run`

### DIFF
--- a/packages/init/src/action/configs.test.ts
+++ b/packages/init/src/action/configs.test.ts
@@ -7,6 +7,7 @@ import { message } from "@optique/core";
 import { kvStores, messageQueues } from "../lib.ts";
 import type { InitCommandData } from "../types.ts";
 import bareBonesDescription from "../webframeworks/bare-bones.ts";
+import nuxtDescription from "../webframeworks/nuxt.ts";
 import { loadDenoConfig } from "./configs.ts";
 import { patchFiles } from "./patch.ts";
 
@@ -133,6 +134,28 @@ test("patchFiles creates a Biome config matching the npm package version", async
   }
 });
 
+test("patchFiles wires Nuxt logging through a Nitro plugin", async () => {
+  const dir = await mkdtemp(join(tmpdir(), "fedify-init-nuxt-"));
+
+  try {
+    const data = await createNuxtNpmInitData(dir);
+    await patchFiles(data);
+
+    const logging = await readFile(join(dir, "server/logging.ts"), "utf8");
+    const plugin = await readFile(
+      join(dir, "server/plugins/logging.ts"),
+      "utf8",
+    );
+
+    assert.match(logging, /export default configure\(/);
+    assert.doesNotMatch(logging, /await configure\(/);
+    assert.match(plugin, /import loggingConfigured from "\.\.\/logging";/);
+    assert.match(plugin, /await loggingConfigured;/);
+  } finally {
+    await rm(dir, { recursive: true, force: true });
+  }
+});
+
 async function createNpmInitData(dir: string): Promise<InitCommandData> {
   const initializer = await bareBonesDescription.init({
     command: "init",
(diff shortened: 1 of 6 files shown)


2chanhaeng (approved):
Looks great! However, I'm wondering if `"nuxt/server/logging.ts"` is the right approach. It might be worth exploring a more versatile method in case we encounter other frameworks where top-level await isn't possible – I hope NOT, but who knows? 😂 Since it's not an urgent requirement, we can discuss this later.

PR #760, opened 2026-05-10
Add oxlint plugin to @fedify/lint

### LINKED ISSUE #702: Add official Oxlint support to @fedify/lint
Add official Oxlint support to `@fedify/lint` by exposing a dedicated entrypoint that Oxlint can load directly.

Background
----------

Right now `@fedify/lint` has two clear entrypoints with two different jobs. On JSR, the root export is a Deno lint plugin. On npm, the root export is the Fedify ESLint flat config, and the actual ESLint plugin object is a named export.

That split works well for Deno and ESLint. It is awkward for Oxlint. Oxlint can load ESLint-compatible JS plugins, but it expects a plugin module, not a package whose default export is already a flat config. In practice, anyone who wants to use `@fedify/lint` from Oxlint needs a small local wrapper module that re-exports the plugin object as the default export.

That extra file is not a serious blocker, but it is unnecessary friction. If Oxlint is going to be a supported way to run Fedify's lint rules in Node.js and Bun projects, `@fedify/lint` should provide a clean entrypoint for it.

Proposal
--------

Add a dedicated entrypoint such as `@fedify/lint/oxlint` that exports the Fedify lint plugin in the shape Oxlint expects.

The existing root exports should stay as they are, so current Deno and ESLint users do not need to change anything.

Scope
-----

 -  add an Oxlint-specific subpath export in `@fedify/lint`
 -  document Oxlint usage in the package README
 -  add tests that exercise the package through Oxlint
 -  keep the current JSR and npm root exports unchanged

Notes
-----

Oxlint's JS plugin support is still marked alpha upstream, so the documentation should say that plainly. Even so, this looks like a reasonable addition because `@fedify/lint` is already very close to working in that environment.

nyanrus:
fix #702 

Expose the existing rule set through oxlint's JS plugin API at the new `@fedify/lint/oxlint` subpath export. The same rules used by the ESLint plugin are reused verbatim — oxlint claims an ESLint-compatible API, so no rule logic is duplicated.

Includes an integration test that spawns the oxlint binary against a tmpdir fixture, plus an example project under examples/lint/oxlint/ that demonstrates the configuration with both a violating and a corrected fixture.

edit:
Deno support added!

### DIFF
--- a/docs/manual/lint.md
+++ b/docs/manual/lint.md
@@ -1,8 +1,8 @@
 ---
 description: >-
-  Fedify provides linting plugins for Deno Lint and ESLint to help you catch
-  common mistakes and enforce best practices when building federated server
-  apps.
+  Fedify provides linting plugins for Deno Lint, ESLint, and Oxlint to help you
+  catch common mistakes and enforce best practices when building federated
+  server apps.
 ---
 
 Linting
@@ -15,8 +15,9 @@ _This package is available since Fedify 2.0.0._
 > app to catch common mistakes early and enforce best practices.
 
 Fedify provides the [`@fedify/lint`] package, which includes lint rules
-specifically designed for Fedify applications. It supports both [Deno Lint] and
-[ESLint], so you can use it regardless of your JavaScript/TypeScript runtime.
+specifically designed for Fedify applications. It supports [Deno Lint],
+[ESLint], and [Oxlint], so you can use it regardless of your
+JavaScript/TypeScript runtime.
 
 The plugin includes rules that check for:
 
@@ -29,6 +30,7 @@ The plugin includes rules that check for:
 [`@fedify/lint`]: https://jsr.io/@fedify/lint
 [Deno Lint]: https://docs.deno.com/runtime/reference/lint_plugins/
 [ESLint]: https://eslint.org/
+[Oxlint]: https://oxc.rs/docs/guide/usage/linter/
 
 
 Installation
@@ -257,6 +259,114 @@ yarn eslint .
 
 ~~~~ sh [Bun]
 bunx eslint .
+~~~~
+
+:::
+
+
+Oxlint
+------
+
+[Oxlint] is a fast Rust-based linter that supports ESLint-compatible JS
+plugins. `@fedify/lint` exposes its rules through Oxlint's [JS plugin API]
+via the `@fedify/lint/oxlint` subpath export.
+
+> [!NOTE]
+> Oxlint's JS plugin API is currently in alpha and not subject to semver.
+
+[JS plugin API]: https://oxc.rs/docs/guide/usage/linter/writing-js-plugins.html
+
+### Basic setup
+
+Add the plugin to your _.oxlintrc.json_ via the `jsPlugins` field, then enable
+the rules you want:
+
+~~~~ json
+{
+  "$schema": "https://raw.githubusercontent.com/oxc-project/oxc/main/npm/oxlint/configuration_schema.json",
+  "jsPlugins": ["@fedify/lint/oxlint"],
+  "rules": {
+    "@fedify/lint/actor-id-required": "error",
+    "@fedify/lint/actor-id-mismatch": "error"
+  }
+}
+~~~~
+
+Rule IDs are namespaced under `@fedify/lint/`, matching the ESLint preset.
+
+### Custom configuration
+
+Each rule accepts `"error"`, `"warn"`, or `"off"`. Enable any subset listed in
+the [*Rules* section](#rules) below:
+
+~~~~ json
+{
+  "jsPlugins": ["@fedify/lint/oxlint"],
+  "rules": {
+    "@fedify/lint/actor-id-required": "error",
+    "@fedify/lint/actor-id-mismatch": "error",
+    "@fedify/lint/actor-inbox-property-required": "warn",
+    "@fedify/lint/actor-outbox-property-required": "warn",
+    "@fedify/lint/actor-followers-property-required": "warn",
+    "@fedify/lint/actor-public-key-required": "warn",
+    "@fedify/lint/actor-assertion-method-required": "warn",
+    "@fedify/lint/collection-filtering-not-implemented": "warn"
+  }
+}
+~~~~
+
+### Running Oxlint
+
+Add a script to _package.json_:
+
+~~~~ jsonc
+{
+  "scripts": {
+    "lint": "oxlint ."
+  }
+}
+~~~~
+
+Then run the linter:
+
+::: code-group
+
+~~~~ sh [npm]
+npm run lint
+~~~~
+
+~~~~ sh [pnpm]
+pnpm lint
+~~~~
+
+~~~~ sh [Yarn]
+yarn lint
+~~~~
+
+~~~~ sh [Bun]
+bun lint
+~~~~
+
+:::
+
+Or invoke Oxlint directly:
+
+::: code-group
+
+~~~~ sh [npm]
+npx oxlint .
+~~~~
+
+~~~~ sh [pnpm]
+pnpx oxlint .
+~~~~
+
+~~~~ sh [Yarn]
+yarn oxlint .
+~~~~
+
+~~~~ sh [Bun]
+bunx oxlint .
 ~~~~
 
 :::
@@ -1217,10 +1327,12 @@ See also
  -  [`@fedify/lint` on npm]
  -  [Deno Lint plugins documentation]
  -  [ESLint documentation]
+ -  [Oxlint documentation]
  -  [Example project]
 
 [`@fedify/lint` on JSR]: https://jsr.io/@fedify/lint
 [`@fedify/lint` on npm]: https://www.npmjs.com/package/@fedify/lint
 [Deno Lint plugins documentation]: https://docs.deno.com/runtime/reference/lint_plugins/
 [ESLint documentation]: https://eslint.org/
+[Oxlint documentation]: https://oxc.rs/docs/guide/usage/linter/
 [Example project]: https://github.com/fedify-dev/fedify/tree/main/examples/lint
--- a/examples/lint/oxlint/README.md
+++ b/examples/lint/oxlint/README.md
@@ -1,0 +1,72 @@
+<!-- deno-fmt-ignore-file -->
+
+@fedify/lint with Oxlint
+========================
+
+This example demonstrates how to use [`@fedify/lint`] together with [Oxlint]
+to catch common Fedify federation mistakes.  Note that Oxlint's JS plugin
+support is upstream alpha and may be unstable.
+
+[`@fedify/lint`]: https://www.npmjs.com/package/@fedify/lint
+[Oxlint]: https://oxc.rs/docs/guide/usage/linter/
+
+
+Layout
+------
+
+ -  *.oxlintrc.json* — Oxlint configuration that enables `@fedify/lint`
+    via the JS plugin API.
+ -  *federation.ts* — code that intentionally violates several rules
+    (missing `id`, `inbox`, `outbox`, `followers`).
+ -  *federation.fixed.ts* — corrected version that passes all rules.
+
+
+Usage
+-----
+
+Install dependencies and run the linter:
+
+~~~~ sh
+pnpm install
+pnpm lint
+~~~~
+
+You should see at least one `@fedify/lint(actor-id-required)` error on
+*federation.ts*. Running against *federation.fixed.ts* alone produces no
+diagnostics:
+
+~~~~ sh
+pnpm lint:fixed
+~~~~
+
+The same tasks are also wired into *deno.json*, so you can invoke Oxlint
+through Deno (the plugin still resolves out of *node\_modules*, so
+`pnpm install` is required first):
+
+~~~~ sh
+deno task lint
+deno task lint:fixed
+~~~~
+
+
+How it works
+------------
+
+The plugin is loaded via the `jsPlugins` field in *.oxlintrc.json*:
+
+~~~~ json
+{
+  "jsPlugins": ["@fedify/lint/oxlint"],
+  "rules": {
+    "@fedify/lint/actor-id-required": "error"
+  }
+}
+~~~~
+
+`@fedify/lint/oxlint` is a subpath export that exposes the same rules as the
+ESLint plugin in Oxlint's plugin shape. Rule IDs are namespaced under
+`@fedify/lint/`.
+
+See the [Linting] manual for the full rule reference.
+
+[Linting]: https://fedify.dev/manual/lint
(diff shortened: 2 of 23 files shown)


docs/manual/lint.md
```diff
@@ -1,8 +1,8 @@
 ---
 description: >-
-  Fedify provides linting plugins for Deno Lint and ESLint to help you catch
-  common mistakes and enforce best practices when building federated server
-  apps.
+  Fedify provides linting plugins for Deno Lint, ESLint, and oxlint to help you
```
dahlia: It's trivial, but their official name is written as Oxlint (capital O).

```suggestion
  Fedify provides linting plugins for Deno Lint, ESLint, and Oxlint to help you
```

docs/manual/lint.md
```diff
@@ -15,8 +15,9 @@ _This package is available since Fedify 2.0.0._
 > app to catch common mistakes early and enforce best practices.
 
 Fedify provides the [`@fedify/lint`] package, which includes lint rules
-specifically designed for Fedify applications. It supports both [Deno Lint] and
-[ESLint], so you can use it regardless of your JavaScript/TypeScript runtime.
+specifically designed for Fedify applications. It supports [Deno Lint],
+[ESLint], and [oxlint], so you can use it regardless of your
+JavaScript/TypeScript runtime.
```
dahlia:
```suggestion
[ESLint], and [Oxlint], so you can use it regardless of your
JavaScript/TypeScript runtime.
```

docs/manual/lint.md
```diff
@@ -29,6 +30,7 @@ The plugin includes rules that check for:
 [`@fedify/lint`]: https://jsr.io/@fedify/lint
 [Deno Lint]: https://docs.deno.com/runtime/reference/lint_plugins/
 [ESLint]: https://eslint.org/
+[oxlint]: https://oxc.rs/docs/guide/usage/linter/
```
dahlia:
```suggestion
[Oxlint]: https://oxc.rs/docs/guide/usage/linter/
```

docs/manual/lint.md
```diff
@@ -262,6 +264,118 @@ bunx eslint .
 :::
 
 
+Oxlint
+------
+
+[oxlint] is a fast Rust-based linter that supports ESLint-compatible JS
```
dahlia:
```suggestion
[Oxlint] is a fast Rust-based linter that supports ESLint-compatible JS
```

docs/manual/lint.md
```diff
@@ -262,6 +264,118 @@ bunx eslint .
...
+plugins. `@fedify/lint` exposes its rules through oxlint's [JS plugin API]
+via the `@fedify/lint/oxlint` subpath export.
+
+> [!NOTE]
+> oxlint's JS plugin API is currently in alpha and the loader runs only under
```
dahlia:
```suggestion
> Oxlint's JS plugin API is currently in alpha and the loader runs only under
```

docs/manual/lint.md
```diff
@@ -262,6 +264,118 @@ bunx eslint .
...
+  }
+}
+~~~~
+
+Rule IDs are namespaced under `@fedify/lint/`, matching the ESLint preset.
+
+### Custom configuration
+
+Each rule accepts `"error"`, `"warn"`, or `"off"`. Enable any subset listed in
+the [Rules] section below:
```
dahlia:
```suggestion
the [*Rules* section](#rules) below:
```

docs/manual/lint.md
```diff
@@ -262,6 +264,118 @@ bunx eslint .
...
+yarn lint
+~~~~
+
+~~~~ sh [Bun]
+bun lint
+~~~~
+
+:::
+
+Or invoke oxlint directly:
```
dahlia:
```suggestion
Or invoke Oxlint directly:
```

docs/manual/lint.md
```diff
@@ -1217,10 +1331,12 @@ See also
  -  [`@fedify/lint` on npm]
  -  [Deno Lint plugins documentation]
  -  [ESLint documentation]
+ -  [oxlint documentation]
```
dahlia:
```suggestion
 -  [Oxlint documentation]
```

docs/manual/lint.md
```diff
@@ -1217,10 +1331,12 @@ See also
...
  -  [Example project]
 
 [`@fedify/lint` on JSR]: https://jsr.io/@fedify/lint
 [`@fedify/lint` on npm]: https://www.npmjs.com/package/@fedify/lint
 [Deno Lint plugins documentation]: https://docs.deno.com/runtime/reference/lint_plugins/
 [ESLint documentation]: https://eslint.org/
+[oxlint documentation]: https://oxc.rs/docs/guide/usage/linter/
```
dahlia:
```suggestion
[Oxlint documentation]: https://oxc.rs/docs/guide/usage/linter/
```

packages/lint/src/tests/oxlint.test.ts
```diff
@@ -0,0 +1,120 @@
...
+ * Skipped when the `oxlint` binary is not on PATH or the built plugin is
+ * missing under `dist/`.
+ */
+import { ok } from "node:assert/strict";
+import { spawnSync } from "node:child_process";
+import { existsSync, mkdtempSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join, resolve } from "node:path";
+import process from "node:process";
+import { test } from "node:test";
```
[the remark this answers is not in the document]
nyanrus: Oxlint did not mention that it only support nodejs only! I will fix with using `@fedify/fixture`
(nodejs-based oxlint requirement was for TS config for oxlint, not jsplugin)

dahlia (changes requested):
Left several nitpicking comments. I guess @2chanhaeng, the original author of `@fedify/lint`, could give you more helpful feedback.

docs/manual/lint.md
nyanrus: Thank you for feedback!

packages/lint/src/tests/oxlint.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    { encoding: "utf8" },
+  );
+  if (where.status === 0 && where.stdout) {
+    return where.stdout.trim().split(/\r?\n/)[0];
+  }
+  return null;
+}
+
+const oxlintBin = findOxlint();
+const ignore = !pluginBuilt || !oxlintBin;
```
2chanhaeng: Don't just ignore-speak to the tester what they should do (e.g., execute `mise run prepare`).

packages/lint/src/tests/oxlint.test.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+test(
+  "oxlint plugin: actor-id-required fires on missing id",
+  { ignore },
+  () => {
+    const dir = mkdtempSync(join(tmpdir(), "fedify-lint-oxlint-"));
+    try {
+      writeFileSync(
+        join(dir, ".oxlintrc.json"),
```
2chanhaeng: ~~Please consider to add comment line: `// cspell: ignore oxlintrc`.~~ `oxlintrc` also exists in *README.md*. Please just add the word to `cspell.json`, too.

packages/lint/src/oxlint.ts
```diff
@@ -0,0 +1,130 @@
...
+  [RULE_IDS.outboxListenerDeliveryRequired]: outboxListenerDeliveryRequired,
+};
+
+/**
+ * Plugin object compatible with oxlint's JS plugin API.
+ */
+export interface OxlintPlugin {
+  meta: { name: string; version: string };
+  rules: Record<string, Rule.RuleModule>;
+}
```
2chanhaeng: Will users import `OxlintPlugin` interface directly from this module? If not, the module shouldn't export this interface. If they will, import the interface from `oxlint` or `@oxlint/plugins` and export it to prevent type sync errors.

packages/lint/src/oxlint.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export const plugin: OxlintPlugin = {
+  meta: {
+    name: metadataRaw.name,
+    version: metadataRaw.version,
+  },
+  rules,
+};
+
+export default plugin;
```
2chanhaeng: `plugin` is exported as default and as `plugin` in the same file.

packages/lint/deno.json
```diff
@@ -3,7 +3,8 @@
   "version": "2.3.0",
   "license": "MIT",
   "exports": {
-    ".": "./src/mod.ts"
+    ".": "./src/mod.ts",
+    "./oxlint": "./src/oxlint.ts"
   },
   "imports": {
     "eslint": "npm:eslint@^9.0.0",
```
2chanhaeng, line 10: If the oxlint module suports Deno, add `oxlint` package to `imports` in *deno.json*.

2chanhaeng (changes requested):
Thanks for contributions! I'm really sorry about late for review.

Would you add `"oxlint"` to *cspell.json" in root?
And if the Oxlint plugin supports Deno, would you add *deno.json* to the oxlint examples?

While the review was delayed, there was a minor update for oxlint. I'm truly sorry for the late review, again, but could you also update that package version to the latest? Thank you once again for your contribution!

packages/lint/src/tests/oxlint.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+if (ignore) {
+  const missing: string[] = [];
+  if (!pluginBuilt) missing.push(`built loader at ${pluginPath}`);
+  if (!oxlintBin) missing.push("oxlint binary on PATH or in node_modules");
+  console.warn(
+    `Skipping oxlint plugin integration test — missing: ${
+      missing.join(", ")
+    }.\n` +
+      "To enable it, run `mise run install` (or `pnpm install && pnpm --filter @fedify/lint build`) " +
```
2chanhaeng: `mise run install` does not build packages. If building the `@fedify/lint` package is necessary, the `prepare` or `prepare-each` tasks might help. Currently, there are two scenarios in which these test files are executed:
1. When the entire test suite is run.
2. When only the `@fedify/lint` package tests are run.

Don't care about case 1 because the `test:*` tasks depend on the `prepare` task, meaning all packages get built. However, for case 2, building at least the individual `@fedify/lint` package is necessary. To handle this, please add instructions telling users to run `mise run install && mise run prepare-each lint`. Avoiding `pnpm` commands is recommended because they might become legacy in the future. Furthermore, it would be even better to add a note recommending that users exclusively use `mise run test` (for full tests) or `mise run test-each lint` (for individual tests) when running tests.

packages/lint/src/tests/oxlint.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  code?: string;
+  message?: string;
+  severity?: string;
+}
+
+interface OxlintJsonReport {
+  diagnostics?: OxlintJsonDiagnostic[];
+}
+
+test(
```
2chanhaeng, line 54: I think your test code is great. However, this test only executes a very small portion of the entire test suite. I would suggest considering integrating this with *integration.test.ts* to test the entire suite. How about defining `testOxlint: (code: string) => Diagnostic[]` and then inspecting the `Diagnostic[]` result in `assertNoErrors` and `assertHasError`? I won't insist, so you can just dismiss this comment.

examples/lint/oxlint/README.md
```diff
@@ -0,0 +1,72 @@
+<!-- deno-fmt-ignore-file -->
+
+@fedify/lint with Oxlint
```
2chanhaeng, line 3: `mise run test:examples <dirs-in-examples>` can test examples from args in `<dirs-in-examples>`, or the entire examples if you omit the args. But for that, configurations are needed in *examples/test-examples/mod.ts*. Would you add configuration for `examples/lint`?
For now, the code uses *pinggy.io* to tunnel, but the *pinggy.io* service is unstable. So before running `mise run test:examples`, the tunnel service should be replaced with *localhost.run*. Also, some example tests could fail, but don't worry about them. Just config about `examples/lint`, please.

2chanhaeng (commented):
Thanks for your contributions, again!
Sorry for late. I've been sick for the past few days and still haven't fully recovered, so my review is late. I should have at least let you know. I apologize if you've been waiting a long time.
Before getting into the review, I recommend that in the future, you leave a comment on the review explaining how you applied the feedback along with the relevant commit hash if you accepted it, or the reasoning behind your decision if you dismissed it. This will help the reviewer review your revisions more quickly.

packages/lint/src/tests/integration.test.ts
2chanhaeng, line 1: Excellent work on integrating the Oxlint test code into the existing *integration.test.ts* file! Moving forward, it would be better to consider refactoring both *oxlint.test.ts* and *integration.test.ts* to modularize redundant logic, as there is significant overlap between the two. The tests in *oxlint.test.ts* share nearly identical setup code with the `testOxlint` function in *integration.test.ts*, and the `findOxlint` implementation is virtually the same. Furthermore, the ignore logic from *oxlint.test.ts* appears to be necessary for *integration.test.ts* as well. Please consider extracting this shared logic into *packages/lint/src/lib/oxlint.ts* to minimize code duplication.

2chanhaeng (changes requested):
Thanks for your awesome contributions! The code is almost complete. If there are no major issues, this will likely be my final review before approval. First, please add `prepare` as a dependency to the `tasks."test:examples"` entry in *mise.toml*.

```diff
 [tasks."test:examples"]
 description = "Run tests for all example projects"
+ depends = ["prepare"]
 run = "deno run -A examples/test-examples/mod.ts"
```

`test:examples` runs with Deno, and as you know, using Oxlint with Deno requires generating the Oxlint binary through a build process. It would be even better to create and add a separate task specifically for generating the Oxlint binary, but I am hesitant to ask because I am not sure how complex that might be. Please look into it and consider implementing it that way if you can.
Secondly, before merging, you need to organize the commit logs using `git rebase` to make them ready for merging. If you ask @daliah for help with the process, I am sure they will be happy to assist. I would love to help you myself, but @daliah knows much more than I do, whereas I am just barely following along with help from LLM agents. 😂
Anyway, thanks again. I hope this is my last review. See ya!

dahlia:
This pull request has some conflicts; could you resolve them?

packages/lint/src/tests/integration.test.ts
nyanrus: done with https://github.com/nyanrus/fedify/blob/d3f7e5f540f511cad50c802358623763fb4e1c69/packages/lint/src/lib/oxlint.ts !

dahlia (approved):
@2chanhaeng Could you review this again?

2chanhaeng (approved):
Looks great! By the way, `findOxlint` finds `oxlint` using `which` or `where`, depending on the `platform`. Since this repository uses `mise`, I thought it might be a good idea to use `oxlint` with `mise` instead. However, since the PR seems to be dragging on for too long, it might be better to leave this as a follow-up issue. Please consider it. Thank you so much for your first contribution!

PR #776, opened 2026-05-22
Add `--skip-install` option to `fedify init`

### LINKED ISSUE #720: Add a `--skip-install` flag to `fedify init`
## Summary

Add a `--skip-install` flag to `fedify init` that skips automatic dependency installation after scaffolding.

## Motivation

`fedify init` currently runs the package manager's install command automatically after generating a project. This is convenient in most cases, but can be undesirable when:

- Working in a CI environment where installation is handled separately.
- Using a monorepo where dependencies are installed from the root.
- The user prefers to inspect the generated files before installing.

## Proposed behavior

- Default (no flag): install dependencies as soon as the files are ready.
- `--skip-install`: skip the install step entirely.

No interactive prompt is needed since installing is the sensible default.

fru1tworld:
Closes #720.

`fedify init` always ran the package manager's install command after scaffolding, with no way to skip it.  This is disruptive in CI pipelines that install from a fixed lockfile in a separate step, in monorepo workspaces that install from the root, and when developers want to inspect the generated files before installing.

Adds `--skip-install` to the `initOptions` schema and threads an `isSkipInstall` predicate through the action pipeline so `installDependencies` is bypassed when the flag is set.  When the run is not in dry-run mode, `noticeSkippedInstall` prints the exact install command for the selected package manager.

## Verification

 -  `mise run check`
 -  `mise run test`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -265,6 +265,15 @@ To be released.
  -  Added `SqliteMessageQueue.getDepth()` for reporting queued, ready, and
     delayed message counts.  [[#735], [#748]]
 
+### @fedify/cli
+
+ -  Added the `--skip-install` option to `fedify init`, following the
+    corresponding `@fedify/init` update, which skips automatic dependency
+    installation after scaffolding.  [[#720], [#776] by fru1tworld]
+
+[#720]: https://github.com/fedify-dev/fedify/issues/720
+[#776]: https://github.com/fedify-dev/fedify/pull/776
+
 ### @fedify/init
 
  -  Added a `--skip-install` option to `fedify init` that skips automatic
@@ -272,9 +281,6 @@ To be released.
     environments, monorepo workspaces that install dependencies from the
     root, or when you want to inspect the generated files before
     installing.  [[#720], [#776] by fru1tworld]
-
-[#720]: https://github.com/fedify-dev/fedify/issues/720
-[#776]: https://github.com/fedify-dev/fedify/pull/776
 
 ### Claude Code plugin
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -265,6 +265,17 @@ To be released.
  -  Added `SqliteMessageQueue.getDepth()` for reporting queued, ready, and
     delayed message counts.  [[#735], [#748]]
 
+### @fedify/init
+
+ -  Added a `--skip-install` option to `fedify init` that skips automatic
+    dependency installation after scaffolding.  This is useful for CI
+    environments, monorepo workspaces that install dependencies from the
+    root, or when you want to inspect the generated files before
+    installing.  [[#720], [#776] by fru1tworld]
+
+[#720]: https://github.com/fedify-dev/fedify/issues/720
+[#776]: https://github.com/fedify-dev/fedify/pull/776
+
 ### Claude Code plugin
 
  -  Added a Claude Code plugin at *claude-plugin/*, installable with:
--- a/packages/init/src/action/mod.ts
+++ b/packages/init/src/action/mod.ts
@@ -39,11 +39,11 @@ import {
  * 5. Converts `InitCommandOptions` into `InitCommandData` via `setData`.
  * 6. Branches based on `isDry`:
  *    - If dry run, executes `handleDryRun`.
- *    - Otherwise, executes `handleHydRun`.
+ *    - Otherwise, executes `handleHydRun`, which installs dependencies, or—when
+ *      `--skip-install` is set—prints how to install them manually via
+ *      `noticeSkippedInstall`.
  * 7. Recommends configuration environment via `recommendConfigEnv`.
- * 8. If installation was skipped and not a dry run, prints how to install
- *    dependencies manually via `noticeSkippedInstall`.
- * 9. Shows how to run the project via `noticeHowToRun`.
+ * 8. Shows how to run the project via `noticeHowToRun`.
  */
 const runInit = (options: InitCommand) =>
   pipe(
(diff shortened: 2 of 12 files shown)


sij411:
cc @2chanhaeng

2chanhaeng:
Hi there! Thank you so much for your interest in contributing to our package. When initializing an app with the `@fedify/init` package, it executes the initialization command of the web framework that the user intends to use for that app. This is stored in the `WebFrameworkInitializer.command` of the respective web framework. For now, the currently supported frameworks prevent package installation when running this command, but frameworks added in the future might allow it. Could you also add a flow that prevents package installation while this command is running when the `--skip-install` flag is used? Thanks again for your interest in contributing to our package! I hope your contributions merge into our package as well!

2chanhaeng (changes requested):
Thanks for your contributions, again!
First, sorry for late. I've been sick for the past few days and still haven't fully recovered, so my review is late. I should have at least let you know. I apologize if you've been waiting a long time.
Before getting into the review, I recommend that in the future, you leave a comment on the review explaining how you applied the feedback along with the relevant commit if you accepted it, or the reasoning behind your decision if you rejected it. This will help the reviewer review your revisions more quickly.
The flow you added looks more complex than I thought. Since `InitCommandOptions` has a `skipInstall` prop, I think it would be enough to just add the related command in `WebFrameworkDescription.init` when `skipInstall` is present. I'm really sorry, but could you revert 39bb2df and then only add the part that handles the `skipInstall` argument in *packages/init/src/webframeworks/next.ts*? It should be sufficient to add the `skipInstall: boolean` param to `getNextInitCommand` and then append `"--skip-install"` if that argument is `true`.
Also, this repo doesn't ban contributions using AI agents, but you should add trailers in the commit messages if you used any AI agents. You can find explanations about trailers in *AI_POLICY.md*. Are you sure you didn't use any AI agents at all?

fru1tworld:
Thanks for the feedback. I've simplified the approach by removing the withSkipInstallArgs logic and now directly passing skipInstall: boolean to getNextInitCommand in next.ts (0cccc38c).
Additionally, I apologize for omitting the AI disclosure trailer in my previous commits. I missed checking AI_POLICY.md regarding this requirement. The Assisted-by trailer is now included.

packages/init/src/action/mod.ts
```diff
@@ -52,6 +56,7 @@ const runInit = (options: InitCommand) =>
     when(isDry, handleDryRun),
     unless(isDry, handleHydRun),
     tap(recommendConfigEnv),
+    tap(unless(isDry, when(isSkipInstall, noticeSkippedInstall))),
```
2chanhaeng: It seems more sense to place `when(isSkipInstall, noticeSkippedInstall)` inside `handleHydRun`. Or is there a specific reason you intended for it to be here?

packages/init/src/action/notice.ts
```diff
@@ -117,6 +117,16 @@ ${instruction}
 Start by editing the ${text(federationFile)} file to define your federation!
 `;
 
+/** Prints a notice that dependency installation was skipped and how to install them manually. */
+export const noticeSkippedInstall = (
+  { packageManager }: InitCommandData,
+) =>
+  printMessage`
+Dependencies were not installed.  Run ${
+    text(`${packageManager} install`)
```
2chanhaeng: The `text` function outputs its arguments so that they appear as plain text. I think [`commandLine`](https://optique.dev/concepts/messages#command-line-examples) is better for this.

2chanhaeng (changes requested):
Thanks for your contributions, and I like your work! I have some suggestions to make your contribution even better. Please consider applying these reviews.

fru1tworld:
Thanks for the review! Moved `noticeSkippedInstall` into `handleHydRun` so it sits right after the install step and only runs in the non-dry-run path. Also switched the manual install hint from `text()` to `commandLine()` as suggested.

2chanhaeng (approved):
LGTM! Thanks for your hard work, and welcome to being one of the contributors to our repository!

fru1tworld:
Thanks for the warm welcome.😃 
Happy to help improve the project.  🍅

CHANGES.md
```diff
@@ -286,6 +265,17 @@ To be released.
  -  Added `SqliteMessageQueue.getDepth()` for reporting queued, ready, and
     delayed message counts.  [[#735], [#748]]
 
+### @fedify/init
+
+ -  Added a `--skip-install` option to `fedify init` that skips automatic
+    dependency installation after scaffolding.  This is useful for CI
+    environments, monorepo workspaces that install dependencies from the
+    root, or when you want to inspect the generated files before
+    installing.  [[#720], [#776] by fru1tworld]
```
dahlia, line 283: I think this item should belong to *@fedify/cli* rather than *@fedify/init*, but what's your thoughts @2chanhaeng?

2chanhaeng: From the user's perspective, it's a feature added to the CLI, but since `@fedify/init` itself can be used independently, I think it's right to keep it in `@fedify/init`. Or how about we add the details and explanation about the `--skip-install` option to the `@fedify/init` section, and then for `@fedify/cli`, just mention that the `--skip-install` option was added to `fedify init` following the `@fedify/init` update?

dahlia: Although `@fedify/init` can be used by other party in theory, it's only consumed by `@fedify/cli` in practice, right? At least, I think it should belong to the both sections.

2chanhaeng: Agreed. @fru1tworld, would you the update changelog about this? I think this is the last one!

fru1tworld: Done in 4212e8b. Thanks for the reviews! 🍅

dahlia (approved):

PR #789, opened 2026-06-05
fix(cli): use ReturnType<typeof setTimeout> for timer map

chenrui333:
## Summary

Deno 2.8+ / TypeScript 6.0 changed `setTimeout`'s return type from `number` to `Timeout`, causing TS2345 when storing timer IDs in the `signalTimers` WeakMap.

## Fix

Use `ReturnType<typeof setTimeout>` which is compatible across all TypeScript/Deno versions.

```
TS2345 [ERROR]: Argument of type 'Timeout' is not assignable to parameter of type 'number'.
  signalTimers.set(controller.signal, timerId);
                                      ~~~~~~~
    at packages/cli/src/lookup.ts:498:39
```

- https://github.com/Homebrew/homebrew-core/pull/286203

### DIFF
--- a/packages/cli/src/lookup.ts
+++ b/packages/cli/src/lookup.ts
@@ -482,7 +482,7 @@ export async function collectAsyncItems<T>(
   }
 }
 
-const signalTimers = new WeakMap<AbortSignal, number>();
+const signalTimers = new WeakMap<AbortSignal, ReturnType<typeof setTimeout>>();
 
 export function createTimeoutSignal(
   timeoutSeconds?: number,


dahlia:
Thanks for your contribution! Since it should be fixed in 2.2.x versions too, I will cherry-pick your commit instead of merging it.

dahlia:
Fixed in v2.0.20, v2.1.16, and v2.2.5.

chenrui333:
Thanks @dahlia!

PR #800, opened 2026-06-09
Speed up codebase type-checking by collapsing dispatcher path overloads

### LINKED ISSUE #613: Optimize `deno task check-all` performance without reducing `pre-commit` coverage
We currently run `deno task check-all` from the Git `pre-commit` hook. That full validation step is important and should remain in place, but the type-checking phase currently takes long enough to make routine commits slower than they need to be.

This issue is about improving the performance of `check-all`, not weakening the checks it performs. In particular, we should preserve full type checking in the `pre-commit` path and instead investigate why the current `deno check` step is so expensive across the workspace.

Areas to investigate:

- Identify which packages or dependency graphs dominate the total `deno check` time
- Determine whether the current workspace-wide invocation causes redundant type checking
- Look for opportunities to improve caching, task structure, or command composition
- Review whether generated artifacts or package boundaries are causing unnecessary work
- Measure before/after timings so any improvement is concrete and verifiable

Acceptance criteria:

- `deno task check-all` becomes meaningfully faster
- Full type-check coverage in the `pre-commit` hook is preserved
- The main sources of the slowdown are understood and documented
- The change does not reduce the reliability of local validation

2chanhaeng:
Fixes #613.

## Background

The `setObjectDispatcher()`, `setCollectionDispatcher()`, and `setOrderedCollectionDispatcher()` methods on `FederationBuilderImpl` typed their `path` parameter as RFC 6570 template-literal unions. `Rfc6570Expression<TParam>` is an 8-member union, and these overloads interpolated it up to four times, so each `path` type expanded to a Cartesian product of up to $8^4 = 4096$ template-literal variants.

TypeScript spent almost all of `deno check`'s time structurally comparing these enormous unions. The cost was not confined to a single file: the context boundary checks (`ContextImpl`/`FederationImpl` vs `Context`/`Federation`) dragged these path types in throughout the type graph, so the combinatorial blowup propagated across the whole project.

## Change

Collapse the affected dispatcher overloads on `FederationBuilderImpl` to a single `path: string` signature:

- `setObjectDispatcher()` — removed 6 template-literal overloads (interpolated `Rfc6570Expression` up to three times, 8^3 = 512 variants, plus several simple multi-variable forms).
- `setCollectionDispatcher()` and `setOrderedCollectionDispatcher()` — removed the template-literal overloads (up to 8^4 = 4096 variants each).

The net diff is **183 deletions in a single file** (`packages/fedify/src/federation/builder.ts`).

## Public API is unchanged

The template-literal overloads are kept on the public `Federatable` interface (and thus on `Federation` and `FederationBuilder`, which extend it). Users still get the same compile-time path validation and autocompletion. Only the internal implementation class signature was collapsed; because `path: string` is a wider, compatible implementation signature, it still satisfies the interface's narrower overloads.

## Nothing is lost at runtime

Path validity—including the "one or more variables" requirement—is still enforced at runtime via `assertPath()` and `Router.variables()` (backed by `@fedify/uri-template`), exactly as before. The only thing removed is the costly compile-time RFC 6570 shape checking.

## Impact

Full-codebase type-checking (`mise run check:types`) drops from **~99s to ~13s**.


dahlia (commented):
Although it's not a hard change, but since speed-ups are good news, we should add an entry about this speed-up to *CHANGES.md*.

2chanhaeng:
> Although it's not a hard change, but since speed-ups are good news, we should add an entry about this speed-up to _CHANGES.md_.

Addressed in 878ab20a2b67e01b8d88431aa13e2e2d1b458384.

dahlia (dismissed):

sij411 (commented):
I can reproduce that this meaningfully improves the expensive type-check phase, so I agree this is relevant to #613.

  In my local A/B check against the PR base, the workspace `deno check` phase got much faster after this change. So this does address one real source of the slowdown.

  That said, I would be careful about saying this fully fixes #613. The issue asks for the main slowdown sources to be understood and documented, and for `check-all`/pre-commit validation to become meaningfully faster without reducing coverage. This PR identifies and improves one major source, but it does not by itself show that the remaining `check-all` cost is understood, nor does it include a reusable benchmark/diagnostic artifact that future changes can compare against.

  There is also a subtle API-boundary point worth documenting. Calls through the supported factory path still preserve the public overloads:

  ```ts
  createFederationBuilder<void>().setObjectDispatcher(
    Note,
    "/notes/{id}",
    (_ctx, values) => {
      values.id satisfies string;
      // @ts-expect-error still rejected
      values.missing;
      return null;
    },
  );
```
  But direct use of the implementation class no longer has that compile-time protection:

```ts
  new FederationBuilderImpl<void>().setObjectDispatcher(
    Note,
    "/notes/{id}",
    (_ctx, values) => {
      // @ts-expect-error now unused
      values.missing;
      return null;
    },
  );
```
  This seems acceptable if `FederationBuilderImpl` is intentionally internal; it is not re-exported from `@fedify/fedify` or `@fedify/fedify/federation`. But I think the PR description should say that explicitly: the public `FederationBuilder`/`Federatable` surface is unchanged, while direct source-level use of `FederationBuilderImpl` moves those checks from compile time to runtime.

  So I would suggest either:
  - clarify that this is a partial fix for #613 focused on the dispatcher overload hot path; and
  - add a small type-level regression test for the supported factory/public path, or document `FederationBuilderImpl` as internal.

2chanhaeng:
> I can reproduce that this meaningfully improves the expensive type-check phase, so I agree this is relevant to #613.
> 
> In my local A/B check against the PR base, the workspace `deno check` phase got much faster after this change. So this does address one real source of the slowdown.
> 
> That said, I would be careful about saying this fully fixes #613. The issue asks for the main slowdown sources to be understood and documented, and for `check-all`/pre-commit validation to become meaningfully faster without reducing coverage. This PR identifies and improves one major source, but it does not by itself show that the remaining `check-all` cost is understood, nor does it include a reusable benchmark/diagnostic artifact that future changes can compare against.
> 
> There is also a subtle API-boundary point worth documenting. Calls through the supported factory path still preserve the public overloads:
> 
> ```ts
> createFederationBuilder<void>().setObjectDispatcher(
>   Note,
>   "/notes/{id}",
>   (_ctx, values) => {
>     values.id satisfies string;
>     // @ts-expect-error still rejected
>     values.missing;
>     return null;
>   },
> );
> ```
> 
> But direct use of the implementation class no longer has that compile-time protection:
> 
> ```ts
>   new FederationBuilderImpl<void>().setObjectDispatcher(
>     Note,
>     "/notes/{id}",
>     (_ctx, values) => {
>       // @ts-expect-error now unused
>       values.missing;
>       return null;
>     },
>   );
> ```
> 
> This seems acceptable if `FederationBuilderImpl` is intentionally internal; it is not re-exported from `@fedify/fedify` or `@fedify/fedify/federation`. But I think the PR description should say that explicitly: the public `FederationBuilder`/`Federatable` surface is unchanged, while direct source-level use of `FederationBuilderImpl` moves those checks from compile time to runtime.
> 
> So I would suggest either:
> 
>     * clarify that this is a partial fix for #613 focused on the dispatcher overload hot path; and
> 
>     * add a small type-level regression test for the supported factory/public path, or document `FederationBuilderImpl` as internal.

Addressed in bb2ef0a17c7e2961231dcb6308ee4e24ecd77d26. Thanks!

Issue #797, opened 2026-06-08
Custom background tasks: core `defineTask` / `enqueueTask` API


2chanhaeng:
First sub-issue of #206.

Background
----------

#206 generalizes the fire-and-then-process pattern that `Context.sendActivity` already uses — enqueue work, return the HTTP response immediately, let a separate worker process the payload later — to arbitrary application-defined long-running jobs (digest mailers, image transcoding, periodic cleanup, …). Tasks run on a separate background worker, not in-request; in-process execution via `setTimeout`/`queueMicrotask` is explicitly out of scope.

This sub-issue delivers the end-to-end core: define a task, enqueue it from a `Context`, run it on a worker, retry on failure. It deliberately excludes deduplication (#798 sub-issue) and task-specific telemetry attributes (#799 sub-issue) so the first PR stays small and reviewable.

Public API
----------

### Registration

`defineTask` is declared on a new `TaskRegistry<TContextData>` interface; `Federatable<TContextData>` extends it, so both `Federation` and `FederationBuilder` inherit the method (they already extend `Federatable`).

~~~~ typescript
import type { StandardSchemaV1 } from "@standard-schema/spec";

interface TaskRegistry<TContextData> {
  defineTask<TSchema extends StandardSchemaV1>(
    name: string,
    options: TaskDefinitionOptions<TContextData, TSchema>,
  ): TaskDefinition<TContextData, StandardSchemaV1.InferOutput<TSchema>>;
}

interface Federatable<TContextData> extends TaskRegistry<TContextData> { /* … */ }
~~~~

The returned `TaskDefinition<TContextData, TData>` carries `name` and `schema` plus a phantom `TContextData` marker — no runtime methods. Dispatch lives on `Context` (below).

### Payload handling

`schema` is **required**. Every task declares a [Standard Schema] object (Zod, Valibot, ArkType, …) and `TData` is inferred from it via `StandardSchemaV1.InferOutput`; there is no explicit-`TData` escape hatch and no `unknown` fallback. The schema validates the *envelope* of the payload, and an Activity Vocabulary object carried inside it (`Note`, `Create`, …) is described as an opaque `instanceof` leaf (`z.instanceof(Create)` / `v.instance(Create)`), so the schema stays small even when the payload carries ActivityPub objects.

[...]

Issue #250, opened 2025-06-09
Add automatic retry capability flag to `MessageQueue` interface


dahlia:
## Problem

Currently, Fedify implements its own retry mechanism for all `MessageQueue` implementations, assuming that message queue backends do not handle retries automatically. However, some message queue backends provide their own sophisticated retry mechanisms:

- `AmqpMessageQueue`: RabbitMQ supports Dead Letter Exchanges, TTL, and native retry policies
- `WorkersMessageQueue`: Cloudflare Queues provides automatic retry with exponential backoff and Dead Letter Queues
- `RedisMessageQueue`: No native retry (messages are lost on handler failure)
- `PostgresMessageQueue`: No native retry (application-level implementation needed)
- `DenoKvMessageQueue`: Deno KV Queues supports automatic retries with exponential backoff
- `InProcessMessageQueue`: No native retry (development only)

This approach has several issues:

1. **Redundant retries**: When both Fedify and the backend implement retries, it can cause unexpected behavior or unnecessary overhead
2. **Underutilization of backend features**: Sophisticated backends like RabbitMQ and Cloudflare Queues provide optimized retry mechanisms that aren't being used
3. **Inconsistent behavior**: Different retry behaviors compared to other systems using the same message queue backend

## Proposed Solution

Add metadata to the `MessageQueue` interface to indicate whether the backend supports automatic retries:

### Option 1: Simple boolean flag
```typescript
interface MessageQueue {
  readonly nativeRetrial?: boolean; // defaults to false
  // existing methods...
}
```

### Option 2: Enum for clarity
```typescript
interface MessageQueue {
  readonly retryHandling?: 'automatic' | 'manual'; // defaults to 'manual'
  // existing methods...
}
```

### Option 3: Capabilities object (for future extensibility)
```typescript
interface MessageQueueCapabilities {
  readonly automaticRetry: boolean;
  readonly deadLetterQueue: boolean;
  readonly delayedRetry: boolean;
}

interface MessageQueue {
  readonly capabilities?: MessageQueueCapabilities; // defaults to all false
  // existing methods...
}
```

## Implementation Strategy

1. **Backward compatibility**: Add the new property as optional to avoid breaking existing implementations
2. **Default behavior**: When the property is undefined/omitted, assume `nativeRetrial: false` (current behavior)
3. **Per-backend configuration**: Each implementation explicitly sets the appropriate value:
   - `AmqpMessageQueue`: `nativeRetrial: true`
[...]

Issue #206, opened 2025-02-09
Support custom background tasks in worker


dahlia:
Currently, Fedify has a built-in background worker that processes incoming and outgoing activities through a message queue. However, applications using Fedify might need their own background tasks as well. This issue proposes to extend Fedify's worker system to support custom tasks, allowing users to utilize Fedify's background worker for their own tasks.

## Scope consideration

Before diving into implementation details, we should consider whether providing custom task support aligns with Fedify's core mission and scope. Fedify is primarily an ActivityPub server framework, and adding general-purpose background task functionality might be stepping outside its intended scope. Some concerns:

- It could make Fedify more complex and harder to maintain
- Users might be better served by dedicated task queue solutions (e.g., Bull, Celery)
- It might dilute Fedify's focus on ActivityPub/fediverse features

However, there are also arguments in favor of this feature:
- Many fediverse applications need background tasks that are closely tied to ActivityPub operations
- Providing integrated task support could simplify application architecture
- Users wouldn't need to set up and maintain a separate task system

This is an important consideration that should be discussed before proceeding with implementation.

## Goals

- Allow users to define and register custom tasks
- Allow users to enqueue custom tasks from anywhere in their application
- Provide type safety for task data
- Support task-specific retry policies and other options
- Maintain backward compatibility

## Proposed solutions

We have two potential approaches to implement this feature:

### Approach 1: Extension of current design

This approach adds new APIs to the existing `Federation` object without major architectural changes:

```typescript
// Define custom task
interface CustomTask<TData = unknown> {
  type: string;
  data: TData;
}

// Register task handler
federation.registerTaskHandler<TData>(
  type: string,
  handler: (ctx: Context<TContextData>, data: TData) => Promise<void> | void
);

// Enqueue task
context.enqueueTask<TData>(
  type: string,
  data: TData,
  options?: MessageQueueEnqueueOptions
);
```

#### Pros
- Minimal changes to existing architecture
- Easier to implement
- Maintains backward compatibility
- Quick to ship

#### Cons
- Keeps worker functionality tightly coupled with `Federation`
- Less flexibility for worker process management
- Limited scalability options

### Approach 2: Separate worker architecture

This approach separates worker functionality into its own module:

```typescript
// Worker configuration
interface WorkerOptions<TContextData> {
  queue: MessageQueue;
  tasks?: Record<TaskType, boolean>;
  contextData?: TContextData;
}

// Dedicated worker class
class Worker<TContextData> {
[...]

PR #806, opened 2026-06-16
Deduplicate custom background tasks

### LINKED ISSUE #798: Custom background tasks: deduplication
Second sub-issue of #206.

Background
----------

Background tasks frequently need at-most-once-per-key enqueue semantics: a digest mailer should not send twice if a request is retried, a cleanup job should coalesce duplicate triggers. The maintainer asked that this mirror the `nativeRetrial` capability flag introduced in #250 — backends that deduplicate natively own the check; otherwise Fedify provides a best-effort KV fallback, with the race-condition tradeoff documented explicitly.

This is kept separate from the core API so the first PR stays small and the deduplication semantics (including the documented best-effort limitation) get their own reviewable boundary.

Public API
----------

### MQ-layer primitives

~~~~ typescript
// mq.ts
export interface MessageQueue {
  readonly nativeRetrial?: boolean;       // existing, #250
  readonly nativeDeduplication?: boolean; // new — backend dedups same deduplicationKey
  // …
}

export interface MessageQueueEnqueueOptions {
  readonly delay?: Temporal.Duration;     // existing
  readonly orderingKey?: string;          // existing
  readonly deduplicationKey?: string;     // new
}
~~~~

These are MQ-layer primitives, not task-layer concepts, so they survive the Approach 2 `Worker` extraction unchanged and are reusable by any future enqueue path.

### Task-API surface

 -  Add `deduplicationKey?: string` to `TaskEnqueueOptions` (the core sub-issue ships `TaskEnqueueOptions` without it; adding an optional field is non-breaking).
 -  `FederationOptions.taskDeduplicationTtl?: Temporal.DurationLike` (default 1 hour) — TTL for the KV fallback entry.
 -  `FederationOptions.taskDeduplicationFallback?: "open" | "closed"` (default `"open"`) — behavior when `deduplicationKey` is set but the queue does not declare `nativeDeduplication` *and* the KV adapter exposes no conditional-write primitive: `"open"` logs at debug and proceeds; `"closed"` throws `TypeError` synchronously.
 -  New `taskDeduplication` KV prefix (default `["_fedify", "taskDeduplication"]`), separate from `activityIdempotence`.

Resolution path
---------------

Inside `#enqueueTasks`, after the queue is resolved, when `deduplicationKey` is supplied:

1.  If `queue.nativeDeduplication === true`: forward `deduplicationKey` in `MessageQueueEnqueueOptions`; the backend owns the check; Fedify does not touch KV.
2.  Otherwise: attempt a conditional KV write under `taskDeduplication` with the TTL, using an `onlyIfNotExists`-style guard where supported (Deno KV `atomic().check()`, Postgres `INSERT … ON CONFLICT DO NOTHING`, Redis `SET NX`, SQLite `INSERT OR IGNORE`). Key present → skip the enqueue; write succeeded → proceed.
3.  KV adapter has no conditional-write primitive → branch on `taskDeduplicationFallback` (`"open"` proceeds, `"closed"` throws).

For `enqueueTaskMany`, a single `deduplicationKey` applies to the whole batch (documented restriction; per-item dedup means calling `enqueueTask` in a loop). This preserves the atomicity guarantee of `nativeDeduplication` backends, which accept one key per call.

Documented limitation
----------------------

The check-then-enqueue sequence in the KV fallback is **not atomic**: two concurrent enqueuers can both observe a missing key and both write. This is best-effort and stated in the public JSDoc for `deduplicationKey`; production deployments needing strict guarantees use a backend with `nativeDeduplication: true`. Cleanup is by TTL expiry, not active deletion on handler success (active cleanup introduces a success→crash-before-delete window; deferred to a later enhancement).

Out of scope
------------

 -  Active KV cleanup on handler success (TTL-only for v1).
 -  Adding `nativeDeduplication: true` to the first-party adapter packages (*packages/postgres*, *packages/redis*, etc.) — track per-adapter follow-ups; this sub-issue ships the core flag + KV fallback, and each adapter opts in separately.

Acceptance criteria
-------------------

 -  [ ] `deduplicationKey` on a `nativeDeduplication: true` queue is forwarded; Fedify does not write KV.
 -  [ ] `deduplicationKey` on a default queue: a second enqueue inside the TTL is skipped; re-enqueue after TTL expiry succeeds.
 -  [ ] `taskDeduplicationFallback: "closed"` throws synchronously when no conditional write is available; `"open"` proceeds with a debug log.
 -  [ ] `taskDeduplication` KV prefix does not collide with `activityIdempotence`.
 -  [ ] `enqueueTaskMany` applies one batch-level `deduplicationKey`.
 -  [ ] Best-effort race limitation documented in JSDoc and *docs/manual/tasks.md*; `CHANGES.md` updated; AI usage disclosed per *AI_POLICY.md*.

References
----------

 -  Umbrella issue (design discussion + background): #206 · depends on the core API sub-issue
 -  `nativeRetrial` prior art (the native-capability flag pattern this mirrors): #250

2chanhaeng:
Resolves #798, the second sub-issue of #206.

> [!IMPORTANT]
> This branch is **stacked on #803** (#797, the first sub-issue), which is **not yet merged**. Until #803 lands, this PR's range includes #803's commits too, so the diff shown here is the core `defineTask` / `enqueueTask` API **plus** this sub-issue's deduplication work. Please merge #803 first; once it lands on `feat/custom-worker`, rebasing this branch drops the overlap and leaves this sub-issue's deduplication changes alone.

Summary
-------

Background tasks frequently need *at-most-once-per-key* enqueue semantics: a digest mailer must not send twice when a request is retried, and a cleanup job should coalesce duplicate triggers. This sub-issue adds an opt-in `deduplicationKey` to the task enqueue path, mirroring the native-capability flag pattern that `nativeRetrial` (#250) established—a backend that deduplicates natively owns the check; otherwise Fedify provides a best-effort key–value fallback with the race-condition tradeoff documented explicitly.

It is kept separate from the core API (#797 / #803) so the first PR stays small and the deduplication semantics—including the documented best-effort limitation—get their own reviewable boundary.

Public API
----------

### MQ-layer primitives

These are message-queue-layer primitives, not task-layer concepts, so they survive a future `Worker` extraction unchanged and are reusable by any enqueue path:

~~~~ typescript
interface MessageQueueEnqueueOptions {
  readonly delay?: Temporal.Duration;     // existing
  readonly orderingKey?: string;          // existing
  readonly deduplicationKey?: string;     // new — at-most-once enqueue for messages sharing it
}

interface MessageQueue {
  readonly nativeRetrial?: boolean;       // existing, #250
  readonly nativeDeduplication?: boolean; // new — backend dedups same deduplicationKey natively
  // …
}
~~~~

`InProcessMessageQueue` declares `nativeDeduplication = false`; `ParallelMessageQueue` inherits the wrapped queue's flag and rejects a batch carrying a `deduplicationKey` when the wrapped queue has no `enqueueMany` (it would otherwise fan out to single `enqueue()` calls that cannot share one key atomically).

### Task-API surface

~~~~ typescript
interface TaskEnqueueOptions {
  readonly delay?: Temporal.DurationLike;
  readonly orderingKey?: string;
  readonly deduplicationKey?: string;     // new (optional → non-breaking on top of #803)
}

interface FederationOptions<TContextData> {
  // …
  taskDeduplicationTtl?: Temporal.DurationLike;     // new, default { hours: 1 }
  taskDeduplicationFallback?: "open" | "closed";    // new, default "open"
}
~~~~

A new `taskDeduplication` key–value prefix (default `["_fedify", "taskDeduplication"]`) holds fallback markers, kept separate from `activityIdempotence`.

Resolution path
---------------

The enqueue pipeline was extracted out of `ContextImpl` into a dedicated `tasks/enqueue.ts` module (handle validation, deduplication planning, payload encoding, and queue dispatch live in one place). Deduplication is decided once, before any payload is encoded:

1.  **Native backend** — when the task's queue declares `nativeDeduplication: true`, Fedify forwards `deduplicationKey` in `MessageQueueEnqueueOptions` and the backend owns the check. The key–value store is never touched.
2.  **Key–value fallback** — otherwise, when the configured `KvStore` exposes the optional compare-and-swap (`KvStore.cas`) primitive, Fedify claims the key under the `taskDeduplication` prefix with `taskDeduplicationTtl`. A present marker skips the enqueue; a won claim proceeds. If the subsequent dispatch fails, the marker is rolled back (via a conditional `cas` clear) so a transient failure does not suppress the retry.
3.  **No conditional write** — when neither applies (no `nativeDeduplication`, and a `KvStore` without `cas`), behavior follows `taskDeduplicationFallback`: `"open"` (default) proceeds without deduplication after a debug-level log; `"closed"` throws a `TypeError` before enqueuing.

Among the first-party adapters, the in-memory, Deno KV, SQLite, and MySQL key–value stores implement `cas`; PostgreSQL and Redis do not yet, so those deployments take the fallback branch until per-adapter follow-ups add it.

Batch semantics
---------------

For `enqueueTaskMany`, a single `deduplicationKey` applies to the whole batch—it enqueues as a unit or is skipped as a unit, never partially. Per-item deduplication means calling `enqueueTask` in a loop, each with its own key. When deduplication is actually applied (a native queue, or a `KvStore` with `cas`), a multi-item batch with a `deduplicationKey` on a queue without `enqueueMany` is rejected rather than risking duplicates, since fanning the key across separate `enqueue()` calls cannot enqueue the batch as one unit. Under the `"open"` fallback no marker is taken, so the batch simply fans out.

Documented limitation
---------------------

The key–value fallback is **best-effort, not transactional**: the marker write and the enqueue are separate operations. Fedify rolls the marker back on enqueue failure, but a crash before that rollback, the `"open"` fallback under concurrency, a non-atomic third-party `cas`, or key reuse within the TTL window can still admit a duplicate or suppress a task. Cleanup is by TTL expiry, not active deletion on handler success (active cleanup introduces a success→crash-before-delete window; deferred). This is stated in the public JSDoc for `deduplicationKey` and in a warning callout in *docs/manual/tasks.md*. Deployments needing strict guarantees use a queue with `nativeDeduplication: true`.

Out of scope
------------

 -  Active key–value cleanup on handler success (TTL-only for now).
 -  Adding `nativeDeduplication: true` / `cas` to the remaining first-party adapters (PostgreSQL, Redis) — tracked as per-adapter follow-ups; this sub-issue ships the core flag plus the key–value fallback.

Acceptance criteria
-------------------

 -  [x] `deduplicationKey` on a `nativeDeduplication: true` queue is forwarded; Fedify does not write to the key–value store.
 -  [x] `deduplicationKey` on a default queue: a second enqueue inside the TTL is skipped; re-enqueue after TTL expiry succeeds.
 -  [x] `taskDeduplicationFallback: "closed"` throws synchronously when no conditional write is available; `"open"` proceeds with a debug log.
 -  [x] `taskDeduplication` key–value prefix does not collide with `activityIdempotence`.
 -  [x] `enqueueTaskMany` applies one batch-level `deduplicationKey`.
 -  [x] Best-effort race limitation documented in JSDoc and *docs/manual/tasks.md*; *CHANGES.md* updated.

Tests
-----

 -  `tasks/enqueue.test.ts` covers the three resolution paths, TTL skip/expiry, the `"open"`/`"closed"` fallback, the prefix isolation, batch-level dedup, the `enqueueMany`-required rejection, and marker rollback on dispatch failure.
 -  `mq.test.ts` covers the new `nativeDeduplication` flag and the `ParallelMessageQueue` batch rejection.
 -  The testing mock (`@fedify/testing`) honors `deduplicationKey` so applications can assert dedup behavior in their own tests.
 -  Tests use `@fedify/fixture` `test()` and pass under Deno, Node.js, and Bun.

Notes for reviewers
-------------------

 -  **Base branch is `feat/custom-worker`** and this is **stacked on the unmerged #803**; see the callout at the top.
 -  `KvStore.cas` is a pre-existing primitive; this PR consumes it rather than introducing it.

AI disclosure
-------------

Assisted-by: Claude Code:claude-opus-4-8
Assisted-by: Codex:gpt-5.5

Codex was used only in review.


2chanhaeng:
@codex review
@coderabbit review
@gemini-code-assist review

AGENTS.md
```diff
@@ -127,11 +127,18 @@ Development workflow
...
+    built. To rebuild a specific package and its dependencies, run
+    `mise run prepare`, or run `mise run prepare-each <pkgs>` to build
+    specific packages.
+ -  **Checking Code**: Run `mise run check` before committing, or run
+    `mise run check-each <pkgs>` to check specific packages.
+ -  **Running Tests**:
+     -  `mise run test`: Executes all the tests in every runtime.
+     -  `mise run test:<runtime:deno,node,bun>`: Executes all the tests by the
+        runtime. If some specific tests are needed, execute
+        `mise run test:<runtime> <test-file-path-from-the-root>`.
```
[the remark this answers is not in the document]
2chanhaeng: Not related.

packages/testing/src/mock.test.ts
```diff
@@ -1711,3 +1712,119 @@ test("MockContext.getActorKeyPairs() returns empty array when no dispatcher regi
...
+    new URL("https://example.com"),
+    undefined,
+  );
+  await assertRejects(
+    () => context.enqueueTask(foreign, 1),
+    TypeError,
+    "is not defined on this federation",
+  );
+  assertEquals(called, 0);
+});
```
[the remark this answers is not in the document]
2chanhaeng: Unrelated.

packages/fedify/src/federation/tasks/codec.ts
```diff
@@ -0,0 +1,191 @@
...
+      if (seen.has(node)) return seen.get(node);
+      for (const reviver of this.#classRevivers) {
+        const out = reviver(seen, inner, node);
+        if (out !== undefined) return await out;
+      }
+      // devalue can handle non-container objects.
+      return node;
+    };
+    return inner;
+  };
```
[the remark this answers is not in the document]
2chanhaeng: Wrong. See [#803](https://github.com/fedify-dev/fedify/pull/803#discussion_r3401416233).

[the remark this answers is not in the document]
2chanhaeng: Wrong. See [#803](https://github.com/fedify-dev/fedify/pull/803#discussion_r3401416233).

packages/fedify/src/federation/tasks/enqueue.ts
```diff
@@ -0,0 +1,294 @@
...
+      items.length,
+    );
+    const messages: TaskMessage[] = await Promise.all(
+      items.map(encodeTaskMessage(ctx.codec, ctx.origin, task, options)),
+    );
+    const claim = await claimDeduplication(ctx, plan, task.name);
+    if (!claim.proceed) return;
+    if (!ctx.federation.manuallyStartQueue) {
+      ctx.federation._startQueueInternal(ctx.data);
+    }
```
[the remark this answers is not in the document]
2chanhaeng: Dismissed. Because the code was replaced at #803.

packages/fedify/src/federation/tasks/codec.ts
```diff
@@ -0,0 +1,191 @@
...
+
+export default class TaskCodec {
+  constructor(readonly options: TaskCodecLoaders) {}
+
+  serialize = (data: unknown): Promise<string> =>
+    stringifyAsync(data, { Vocab: this.#stringifyVocab });
+
+  deserialize = (raw: string): Promise<unknown> =>
+    this.#revive(new Map())(parse(raw, { Vocab: VocabHolder.from }));
+
```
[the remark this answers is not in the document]
2chanhaeng: Addressed in 3c6ec35d. Marked `TaskCodec` `@internal` and added brief JSDoc to `serialize()`, `deserialize()`, and `static validate()`.

packages/fedify/src/federation/middleware.test.ts
```diff
@@ -10480,3 +10504,235 @@ test("createFederation() omits instrumentation when no meterProvider is set", ()
...
+  });
+  return Promise.race([promise, timeout]).finally(() => clearTimeout(timer));
+};
+
+const taskFederationOptions = {
+  kv: new MemoryKvStore(),
+  documentLoaderFactory: () => mockDocumentLoader,
+  contextLoaderFactory: () => mockDocumentLoader,
+  manuallyStartQueue: true,
+};
```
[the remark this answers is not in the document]
2chanhaeng: Addressed in 465147c0c. Replaced with a `mockOptions()` factory.

packages/fedify/src/federation/tasks/tasks.test.ts
```diff
@@ -0,0 +1,637 @@
...
+      signal: controller.signal,
+      queue: "task",
+    });
+    strictEqual(taskQueue.listenCount, 1);
+    strictEqual(inbox.listenCount, 0);
+    strictEqual(outbox.listenCount, 0);
+    strictEqual(fanout.listenCount, 0);
+    controller.abort();
+    await listening;
+  });
```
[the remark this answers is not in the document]
2chanhaeng: In 7ea20a4c3 two steps were added to the `startQueue() task worker` test: with the outbox fallback (no task queue) and with `queue.task` sharing the outbox instance, both assert that `queue: "task"` listens the effective queue exactly once (`listenCount === 1`) and leaves the others at 0.

packages/fedify/src/testing/tasks.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  listen(
+    _handler: (message: TaskMessage) => Promise<void> | void,
+    options?: MessageQueueListenOptions,
+  ): Promise<void> {
+    this.listenCount++;
+    return new Promise((resolve) => {
+      options?.signal?.addEventListener("abort", () => resolve());
+    });
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 7262a07ab. `listen()` now returns `Promise.resolve()` when `signal.aborted` is already set, and registers the abort listener with `{ once: true }`. Added a test asserting `listen()` resolves immediately for a pre-aborted signal.

packages/fedify/src/federation/middleware.ts
```diff
@@ -894,12 +941,31 @@ export class FederationImpl<TContextData>
     return this.tracerProvider.getTracer(metadata.name, metadata.version);
   }
 
+  resolveTaskQueue(taskName: string): MessageQueue | undefined {
+    const def = this.taskDefinitions.get(taskName);
+    const resolved = def?.queue ?? this.taskQueue;
+    if (resolved != null) return resolved;
+    return this.taskQueueResolution === "strict" ? undefined : this.outboxQueue;
+  }
```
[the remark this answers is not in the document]
2chanhaeng: Fixed in 7ea20a4c3. `_startQueueInternal` now iterates a (role, queue) target list and starts each instance once via an identity Set. The targets include the outbox fallback and per-task queues, so the `queue: "task"` selector starts a shared or fallback-routed effective task queue. Added two regression tests for the fallback and shared cases.

packages/fedify/src/federation/mq.ts
```diff
@@ -412,6 +451,15 @@ export class ParallelMessageQueue implements MessageQueue {
...
     if (this.queue.enqueueMany == null) {
+      if (options?.deduplicationKey != null) {
+        throw new TypeError(
+          "Cannot enqueue a batch with a deduplicationKey: the wrapped queue " +
+            "does not implement enqueueMany, so ParallelMessageQueue would " +
+            "have to fan out to individual enqueue() calls that cannot share " +
+            "one deduplicationKey atomically.  Wrap a queue that implements " +
+            "enqueueMany instead.",
+        );
+      }
```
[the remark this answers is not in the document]
2chanhaeng: Addressed in 9b2ab8ee5. Since the atomicity limitation only applies to multi-item fan-out, the guard now allows a single-item batch to fall back to `enqueue()`. Note the task path never calls `enqueueMany` for a single item, but this improves robustness for direct `ParallelMessageQueue` callers.

packages/fedify/src/federation/context.ts
```diff
@@ -429,6 +430,54 @@ export interface Context<TContextData> {
...
+   * @template TData The type of the task payload, inferred from the task's
+   *                 schema.
+   * @param task The handle returned by {@link TaskRegistry.defineTask}.
+   * @param payloads The task payloads.  Each is validated against the
+   *                 task's schema before being enqueued.
+   * @param options Options for enqueuing the tasks.
+   * @throws {TypeError} If the task is not defined on this federation,
+   *                     if no message queue is configured for tasks, or if
+   *                     a payload fails schema validation.
+   * @since 2.x.x
```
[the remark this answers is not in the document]
2chanhaeng: Addressed in 3c6ec35d. Made the fallback wording conditional and added the missing `@throws {TypeError}` for a deduplicated multi-item batch that cannot be enqueued atomically.

packages/fedify/src/federation/tasks/codec.ts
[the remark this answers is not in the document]
2chanhaeng: Wrong. See [#803](https://github.com/fedify-dev/fedify/pull/803#discussion_r3401416233).

CHANGES.md
```diff
@@ -285,8 +285,18 @@ To be released.
...
+        `MessageQueue.nativeDeduplication` capability owns the check and
+        receives the key through the new
+        `MessageQueueEnqueueOptions.deduplicationKey`; otherwise Fedify
+        performs a best-effort key–value guard through the optional
+        `KvStore.cas` primitive, under a new `taskDeduplication` key prefix.
+        The marker TTL and the no-`cas` fallback are tunable with the new
+        `FederationOptions.taskDeduplicationTtl` and
+        `FederationOptions.taskDeduplicationFallback` options.
+
+    [[#206], [#797], [#798], [#803] by ChanHaeng Lee]
```
dahlia: I think we need to rebase the upstream `feat/custom-worker` branch first so that the branch catches up the recent v2.3.0 release. This entry should go inside *Version 2.4.0* instead.

2chanhaeng: Addressed in a1b0bb3ed5ef8ccc2963cfda36c72402ecaa450a. After this, if the implementation is not applied immediately, I should add it as a "To be added" section or something similar at the top to make it easier to manage instead of in the latest version section.

docs/manual/tasks.md
```diff
@@ -257,12 +263,96 @@ delivered it.
...
+
+~~~~ typescript
+await ctx.enqueueTask(sendDigest, payload, {
+  deduplicationKey: `digest:${payload.userId}`,  // [!code highlight]
+});
+~~~~
+
+How the key is resolved depends on the queue and the key–value store:
+
+1.  **Native backend.**  When the task's queue declares
```
dahlia: This label does not need bold emphasis; italics should be enough here: `*Native backend.*`.

docs/manual/tasks.md
```diff
@@ -257,12 +263,96 @@ delivered it.
...
+    `~MessageQueue.nativeDeduplication`, Fedify forwards the key in the
+    message queue's `~MessageQueueEnqueueOptions.deduplicationKey` and the
+    backend owns the check.  Fedify does not touch the key–value store.
+
+2.  **Key–value fallback.**  Otherwise, if the configured `~KvStore` exposes
```
dahlia: Same here. This reads more like a section label than a warning, so italics would be lighter: `*Key–value fallback.*`.

docs/manual/tasks.md
```diff
@@ -257,12 +263,96 @@ delivered it.
...
+    configurable with `~FederationOptions.taskDeduplicationTtl`:
+
+    ~~~~ typescript
+    const federation = createFederation<void>({
+      // ...
+      taskDeduplicationTtl: { minutes: 10 },  // [!code highlight]
+    });
+    ~~~~
+
+3.  **No conditional write.**  When neither applies—no native deduplication and
```
dahlia: Same here: `*No conditional write.*` should be enough without bold emphasis.

docs/manual/tasks.md
```diff
@@ -257,12 +263,96 @@ delivered it.
...
+risking duplicates.  Under the `"open"` fallback (no native deduplication and no
+`cas`), no marker is taken, so the batch simply fans out without deduplication.
+
+This applies through `~ParallelMessageQueue` as well: wrapping a queue that
+lacks `~MessageQueue.enqueueMany()` does not make batch enqueue atomic, so a
+deduplicated multi-item batch on such a wrapper is likewise rejected rather than
+collapsed onto one message.
+
+> [!WARNING]
+> The key–value fallback is **best-effort, not transactional**.  The marker
```
dahlia: This is already inside a warning block, so bold feels stronger than necessary. Italics should carry the emphasis: `*best-effort, not transactional*`.

docs/manual/tasks.md
dahlia: `~KvStore` appears to misuse markdown-it-jsr-ref syntax. The `~` prefix is for member references such as `~ClassName.fieldName` or `~ClassName.methodName()`, where the owner type is hidden in the rendered text; a standalone type name should not use it. See the Syntax section: https://github.com/dahlia/markdown-it-jsr-ref#syntax

docs/manual/tasks.md
```diff
@@ -257,12 +263,96 @@ delivered it.
...
+key out across separate `~MessageQueue.enqueue()` calls cannot enqueue a whole
+batch as one unit: a native per-message key cannot cover it, and a key–value
+marker could not be rolled back cleanly if only some of the fanned-out enqueues
+failed.  So when deduplication is actually applied—a native queue, or a
+key–value store with `~KvStore.cas`—Fedify rejects a multi-item batch with a
+`deduplicationKey` on a queue without `~MessageQueue.enqueueMany()` instead of
+risking duplicates.  Under the `"open"` fallback (no native deduplication and no
+`cas`), no marker is taken, so the batch simply fans out without deduplication.
+
+This applies through `~ParallelMessageQueue` as well: wrapping a queue that
```
dahlia: Same issue here: `~ParallelMessageQueue` is a standalone type reference, so the `~` prefix does not match markdown-it-jsr-ref syntax. It should be written without `~`, unlike member references such as `~ClassName.methodName()`. See https://github.com/dahlia/markdown-it-jsr-ref#syntax

packages/fedify/src/federation/tasks/task.ts
```diff
@@ -159,6 +159,26 @@ export interface TaskEnqueueOptions {
...
+   *
+   * For {@link Context.enqueueTaskMany}, one key governs the whole batch.  When
+   * deduplication is actually applied—a native queue, or the key–value
+   * fallback through {@link KvStore.cas}—a multi-item batch with a
+   * `deduplicationKey` requires the queue to implement
+   * {@link MessageQueue.enqueueMany} so it enqueues atomically, or the call
+   * throws a `TypeError`.  Under the `"open"` fallback with no `cas`, no marker
+   * is taken, so such a batch instead fans out without deduplication.
+   *
+   * @since 2.x.x
```
dahlia: `@since 2.x.x` should be `@since 2.4.0` for this API.

packages/fedify/src/federation/mq.ts
```diff
@@ -25,6 +25,20 @@ export interface MessageQueueEnqueueOptions {
...
+   * An optional key requesting at-most-once enqueue semantics for messages
+   * that share it.  A backend that declares
+   * {@link MessageQueue.nativeDeduplication} `true` owns the check: a message
+   * whose `deduplicationKey` was already seen within the backend's
+   * deduplication window is dropped instead of enqueued.  Backends without
+   * native deduplication ignore this field; Fedify performs its own
+   * best-effort deduplication before reaching them on the paths that support
+   * it.
+   *
+   * @since 2.x.x
```
dahlia: `@since 2.x.x` should be `@since 2.4.0` for this API.

packages/fedify/src/federation/mq.ts
```diff
@@ -87,6 +101,18 @@ export interface MessageQueue {
...
 
+  /**
+   * Whether the message queue backend deduplicates messages that share a
+   * {@link MessageQueueEnqueueOptions.deduplicationKey} natively.  When `true`,
+   * Fedify forwards the `deduplicationKey` and relies on the backend to drop
+   * duplicates; when `false` or omitted, Fedify applies its own best-effort
+   * key–value deduplication on the paths that request it.
+   *
+   * @default `false`
+   * @since 2.x.x
```
dahlia: `@since 2.x.x` should be `@since 2.4.0` for this API.

packages/fedify/src/federation/mq.ts
```diff
@@ -176,6 +202,12 @@ export class InProcessMessageQueue implements MessageQueue {
    */
   readonly nativeRetrial = false;
 
+  /**
+   * In-process message queue does not deduplicate messages natively.
+   * @since 2.x.x
```
dahlia: `@since 2.x.x` should be `@since 2.4.0` for this API.

packages/fedify/src/federation/mq.ts
```diff
@@ -365,6 +397,12 @@ export class ParallelMessageQueue implements MessageQueue {
    * @since 1.7.0
    */
   readonly nativeRetrial?: boolean;
+
+  /**
+   * Inherits the native deduplication capability from the wrapped queue.
+   * @since 2.x.x
```
dahlia: `@since 2.x.x` should be `@since 2.4.0` for this API.

packages/fedify/src/federation/middleware.ts
```diff
@@ -512,6 +513,15 @@ export interface FederationKvPrefixes {
    * @since 2.3.0
    */
   readonly circuitBreaker: KvKey;
+
+  /**
+   * The key prefix used for storing custom background task deduplication
+   * markers.  Kept separate from {@link activityIdempotence} so the two key
+   * spaces never collide.
+   * @default `["_fedify", "taskDeduplication"]`
+   * @since 2.x.x
```
dahlia: `@since 2.x.x` should be `@since 2.4.0` for this API.

packages/fedify/src/federation/federation.ts
```diff
@@ -1103,6 +1103,32 @@ export interface FederationOptions<TContextData> {
...
   taskQueueResolution?: "fallback" | "strict";
 
+  /**
+   * The time-to-live for a {@link TaskEnqueueOptions.deduplicationKey} marker
+   * stored in the key–value deduplication fallback.  A second enqueue with the
+   * same key within this window is skipped; once it expires, the key may
+   * enqueue again.  Ignored when the task's queue declares
+   * {@link MessageQueue.nativeDeduplication} (the backend owns the window).
+   * @default `{ hours: 1 }`
+   * @since 2.x.x
```
dahlia: `@since 2.x.x` should be `@since 2.4.0` for this API.

packages/fedify/src/federation/federation.ts
```diff
@@ -1103,6 +1103,32 @@ export interface FederationOptions<TContextData> {
...
+   * but the task's queue does not declare
+   * {@link MessageQueue.nativeDeduplication} *and* the configured
+   * {@link KvStore} exposes no `cas` (compare-and-swap) primitive:
+   *
+   * - `"open"` (the default): proceeds without deduplication after logging at
+   *   debug level.
+   * - `"closed"`: rejects with a `TypeError` before enqueuing.
+   *
+   * @default `"open"`
+   * @since 2.x.x
```
dahlia: `@since 2.x.x` should be `@since 2.4.0` for this API.

packages/fedify/src/federation/context.ts
```diff
@@ -459,17 +459,21 @@ export interface Context<TContextData> {
...
    * @param options Options for enqueuing the tasks.
    * @throws {TypeError} If the task is not defined on this federation,
-   *                     if no message queue is configured for tasks, or if
-   *                     a payload fails schema validation.
+   *                     if no message queue is configured for tasks, if
+   *                     a payload fails schema validation, or if a
+   *                     deduplicated multi-item batch cannot be enqueued
+   *                     atomically because the queue does not implement
+   *                     bulk enqueue.
    * @since 2.x.x
```
dahlia: `@since 2.x.x` should be `@since 2.4.0` for this API.

docs/manual/tasks.md
```diff
@@ -257,12 +263,96 @@ delivered it.
 > queue and set `taskQueueResolution: "strict"`.
 
 
+Deduplication
+-------------
+
+A task often needs *at-most-once-per-key* enqueue: a digest mailer must not
+send twice when a request is retried, and a cleanup job should coalesce
+duplicate triggers.  Passing a `deduplicationKey` requests this—a second
```
dahlia: This sentence reads as if `deduplicationKey` always drops duplicate enqueues. That is not true for the default `"open"` fallback when the queue has no native deduplication and the configured key–value store has no `cas`; in that case Fedify logs and proceeds without deduplication. It would be clearer to say this is what the key requests, and that the later fallback rules decide whether the request can actually be enforced.

packages/fedify/src/federation/tasks/enqueue.ts
```diff
@@ -0,0 +1,294 @@
...
+  queue: MessageQueue,
+  taskName: string,
+  options: TaskEnqueueOptions,
+  itemCount: number,
+): DedupPlan {
+  if (options.deduplicationKey == null) return { kind: "none" };
+  const key = options.deduplicationKey;
+  const native = queue.nativeDeduplication === true;
+  const canCas = ctx.federation.kv.cas != null;
+  if (itemCount > 1 && queue.enqueueMany == null && (native || canCas)) {
```
dahlia: This capability check is fooled by `ParallelMessageQueue`: the wrapper always has an `enqueueMany()` method even when the wrapped queue does not. On the CAS fallback path that means a deduplicated multi-item batch can claim one KV marker, then `ParallelMessageQueue.enqueueMany()` fans out individual `enqueue()` calls because no forwarded `deduplicationKey` is present. If one enqueue fails, the batch is no longer atomic, and a successful partial fan-out can still reserve the marker for the whole batch. This also contradicts the docs that say a deduplicated multi-item batch through such a wrapper is rejected. Could this check detect that the underlying queue can actually enqueue the batch atomically, or otherwise reject before claiming the marker?

dahlia (changes requested):
Left several comments. Most of them are just nitpicks, but some of them look important.

packages/fedify/src/federation/tasks/task.ts
2chanhaeng: Addressed in 5a66b3f049e9def32387cda317152a27da1b5f61.

packages/fedify/src/federation/mq.ts
2chanhaeng: Addressed in 5a66b3f049e9def32387cda317152a27da1b5f61.

packages/fedify/src/federation/mq.ts
2chanhaeng: Addressed in 5a66b3f049e9def32387cda317152a27da1b5f61.

packages/fedify/src/federation/mq.ts
2chanhaeng: Addressed in 5a66b3f049e9def32387cda317152a27da1b5f61.

packages/fedify/src/federation/mq.ts
2chanhaeng: Addressed in 5a66b3f049e9def32387cda317152a27da1b5f61.

packages/fedify/src/federation/middleware.ts
2chanhaeng: Addressed in 5a66b3f049e9def32387cda317152a27da1b5f61.

packages/fedify/src/federation/federation.ts
2chanhaeng: Addressed in 5a66b3f049e9def32387cda317152a27da1b5f61.

packages/fedify/src/federation/federation.ts
2chanhaeng: Addressed in 5a66b3f049e9def32387cda317152a27da1b5f61.

packages/fedify/src/federation/context.ts
2chanhaeng: Addressed in 5a66b3f049e9def32387cda317152a27da1b5f61. In my defense, I had already fixed it locally, but I was waiting to push it together after applying the next review to avoid re-running the CI/CD.

docs/manual/tasks.md
2chanhaeng: Addressed in 799087d55601fbdbf5ba64888ad4aca9ea195705.

docs/manual/tasks.md
2chanhaeng: Addressed in 799087d55601fbdbf5ba64888ad4aca9ea195705.

docs/manual/tasks.md
2chanhaeng: Addressed in 799087d55601fbdbf5ba64888ad4aca9ea195705.

docs/manual/tasks.md
2chanhaeng: Addressed in 799087d55601fbdbf5ba64888ad4aca9ea195705.

docs/manual/tasks.md
2chanhaeng: Addressed in 799087d55601fbdbf5ba64888ad4aca9ea195705.

docs/manual/tasks.md
2chanhaeng: Addressed in 799087d55601fbdbf5ba64888ad4aca9ea195705. I will make sure to familiarize myself with the usage of `~` prefix.

docs/manual/tasks.md
2chanhaeng: Addressed in 40311c436. Reworded the section to state that the key *requests* at-most-once-per-key enqueue, and that whether the drop is actually enforced depends on the fallback rules below (native / KV `cas` / `"open"`·`"closed"`).

packages/fedify/src/federation/tasks/enqueue.ts
2chanhaeng: Addressed in 7fff167cf5d778bd960af2b7ad184648bb398549. The check now looks through the wrapper (`queue.queue.enqueueMany`) and rejects before claiming any marker. Added a cas-path regression test to prove it.

dahlia (approved):
Great job!

PR #817, opened 2026-06-23
Add FEP-0837 Agreement and Commitment vocabulary types

### LINKED ISSUE #775: Add vocabulary for commitment and agreement
### Summary

I would like to have the full vocabulary for [FEP-0837: Federated Marketplace; Status: Draft](https://codeberg.org/fediverse/fep/src/branch/main/fep/0837/fep-0837.md) in fedify. After #578 the merge of #579 the first part of the vocabulary is already available. The next step is to all add commitment and agreement. 

### Problem

Currently, there is no vocabulary included for commitment and agreement, and no way to add custom vocabulary. 

### Proposed Solution

Adding the vocabulary for commitment and agreement based on the draft Version of FEP-0837. 

### Alternatives Considered

It would also suit me if it were possible to add custom vocabulary to Fedify projects like outlined in #207 

### Scope / Dependencies

Adding the YAML files for commitment and agreement, and adding the according vocab-tools snapshots.

scammo:
Add FEP-0837 Agreement and Commitment vocabulary types https://github.com/fedify-dev/fedify/issues/775
Assisted-by: Claude Code:claude-opus-4-8

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -409,6 +409,19 @@ To be released.
     identifiers (`_:b0`) against `options.baseUrl`; blank nodes are left
     as `null` in the resulting instance's `id` field.  [[#792]]
 
+ -  Added the second-stage vocabulary types for [FEP-0837], economic
+    resource coordination in federated networks.
+    [[#775] by Samuel Brinkmann]
+
+     -  Added `Agreement` class, representing the agreement reached between
+        parties responding to a `Proposal`, wrapped in an `Offer` and
+        finalized as the `result` of an `Accept`.
+     -  Added `Commitment` class, representing a promised economic
+        transaction that references an `Intent` via `satisfies` and carries
+        the committed quantity via `resourceQuantity`.
+
+[FEP-0837]: https://w3id.org/fep/0837
+[#775]: https://github.com/fedify-dev/fedify/issues/775
 [#790]: https://github.com/fedify-dev/fedify/issues/790
 
 ### @fedify/vocab-tools
@@ -704,7 +717,6 @@ Released on April 28, 2026.
         measure, with `unit` and `numericalValue` properties.
 
 [FEP-044f]: https://w3id.org/fep/044f
-[FEP-0837]: https://w3id.org/fep/0837
 [#452]: https://github.com/fedify-dev/fedify/issues/452
 [#578]: https://github.com/fedify-dev/fedify/issues/578
 [#645]: https://github.com/fedify-dev/fedify/issues/645
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -425,7 +425,7 @@ To be released.
 
  -  Added the second-stage vocabulary types for [FEP-0837], economic
     resource coordination in federated networks.
-    [[#775] by Samuel Brinkmann]
+    [[#775], [#817] by Samuel Brinkmann]
 
      -  Added `Agreement` class, representing the agreement reached between
         parties responding to a `Proposal`, wrapped in an `Offer` and
@@ -437,6 +437,7 @@ To be released.
 [FEP-0837]: https://w3id.org/fep/0837
 [#775]: https://github.com/fedify-dev/fedify/issues/775
 [#790]: https://github.com/fedify-dev/fedify/issues/790
+[#817]: https://github.com/fedify-dev/fedify/pull/817
 
 ### @fedify/vocab-tools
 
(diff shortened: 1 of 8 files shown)


CHANGES.md
```diff
@@ -423,6 +423,19 @@ To be released.
     identifiers (`_:b0`) against `options.baseUrl`; blank nodes are left
     as `null` in the resulting instance's `id` field.  [[#792]]
 
+ -  Added the second-stage vocabulary types for [FEP-0837], economic
+    resource coordination in federated networks.
+    [[#775] by Samuel Brinkmann]
```
dahlia: Please add a link to this pull request as well!

scammo: Thank you for this information. I fixed this in [b49a383](https://github.com/fedify-dev/fedify/pull/817/commits/b49a383cd01ff1fc0fbf55003b7725dc9586b904)

packages/vocab/src/vocab.test.ts
```diff
@@ -2802,6 +2811,338 @@ test(
...
+    "currencyAmount",
+  );
+  deepStrictEqual(
+    restored.stipulatesReciprocal?.resourceQuantity?.numericalValue,
+    parseDecimal("30.00"),
+  );
+});
+
+test("FEP-0837: Agreement parses verbatim Accept-result example from spec", async () => {
+  // Verbatim from FEP-0837 "Accepting an agreement" example.
```
dahlia: Not quite verbatim: the spec's “Accepting an agreement” example maps `hasUnit` and `hasNumericalValue` to the `vf:` prefix and never defines `om2:`. This `@context` rewrites them to `om2:` (and adds the `om2` prefix) so Fedify's `Measure` can parse it.

The data is the spec's, but the context isn't. As written, “verbatim” implies Fedify parses the spec's literal example, which it can't, since the spec's `vf:hasUnit` term wouldn't be recognized. Could you reword the comment and test name? Something like “adapted from … (context rewritten to use the `om2` prefix Fedify's `Measure` expects).”

packages/vocab/src/commitment.yaml
```diff
@@ -0,0 +1,45 @@
...
+  hasNumericalValue: "om2:hasNumericalValue"
+
+properties:
+- singularName: satisfies
+  functional: true
+  compactName: satisfies
+  uri: "https://w3id.org/valueflows/ont/vf#satisfies"
+  description: |
+    The {@link Intent} of a {@link Proposal} that this commitment
+    satisfies.  The value is typically a fragment identifier such as
```
dahlia: The `satisfies` value isn't a bare fragment. The spec and the tests all use absolute URIs that *end* in such a fragment (e.g. `https://market.example/proposals/…#primary`), and the range here is `anyURI`, so “a fragment identifier such as `#primary`” could be read as endorsing a bare `#primary`. Maybe “a URI ending in a fragment such as `…#primary` or `…#reciprocal`.”

dahlia (commented):
Thanks for finishing the FEP-0837 second stage. The `Agreement` and `Commitment` definitions track the existing `Proposal`/`Intent`/`Measure` pattern, the `defaultContext` blocks cover the embedded objects, and the roundtrip tests are solid. Two small wording nits below, neither blocking.

_Disclosure: this review was assisted by Claude Code (claude-opus-4-8)._

packages/vocab/src/vocab.test.ts
scammo: Thank you for catching this. I fixed this.

packages/vocab/src/commitment.yaml
scammo: Thank you for catching this detail. I fixed this, and tried to clarify it. I'm not sure if the text is enough.
I also did add link to "fragment" to a MDN doc to make clear what is meant, because from my taste, using fragments in this area feels a bit exotic. I can remove it if it does not fit into the desired text of the docs.

packages/vocab/src/commitment.yaml
```diff
@@ -0,0 +1,47 @@
...
+- singularName: satisfies
+  functional: true
+  compactName: satisfies
+  uri: "https://w3id.org/valueflows/ont/vf#satisfies"
+  description: |
+    The {@link Intent} of a {@link Proposal} that this commitment
+    satisfies.  The value is the proposal's URI with a
+    [fragment](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment)
+    appended to identify the intent.  Recommended fragments are
+    `primary` and`reciprocal`.
```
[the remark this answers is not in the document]
scammo: From my point of view the link to the mdn docs is more useful than using the #. In the FEP they also don't use the # https://fediverse.codeberg.page/fep/fep/0837/#proposals.

dahlia (approved):
Thanks for your contribution!

PR #842, opened 2026-06-25
Fall back when Linked Data signing hits JSON-LD processing errors

### LINKED ISSUE #824: Linked Data signing should not crash outbound activity delivery on canonization depth errors
Hackers' Pub is seeing repeated production failures from Fedify 2.2.4 while sending a `Follow` `Accept` activity. The Sentry issue is `GRAPHQL-30` and the event count was 14 when triaged.

Observed behavior
-----------------

The application calls `fedCtx.sendActivity()` from *federation/inbox/following.ts* after accepting a follower. Fedify then creates a Linked Data signature in *packages/fedify/src/federation/middleware.ts* by calling `signJsonLd()`. The stack reaches `hashJsonLd()` in *packages/fedify/src/sig/ld.ts*, where `jsonld.canonize()` throws `Error: Maximum deep iterations exceeded (6)`.

Because this exception is not caught, one problematic outbound activity aborts the whole follow-acceptance handler after the database rows and notification have already been created.

Expected behavior
-----------------

If JSON-LD canonization fails while adding the optional Linked Data signature, Fedify should either send the activity without that signature and log the signing failure, or expose a documented way for applications to opt into that behavior. The send should not fail after the application has already committed local follow state.

Notes
-----

The failure is in the Fedify signing pipeline rather than Hackers' Pub's model or GraphQL code. Hackers' Pub can avoid some damage with app-level retries, but it cannot safely recover from a canonization exception thrown inside `sendActivity()` without duplicating Fedify delivery behavior.

notJoon:
## Description

Fixes #824

This changes outbound activity delivery to continue without an optional Linked Data Signature when JSON-LD canonicalization/processing fails during signing. These failures are logged as warnings.

Other signing failures, such as key, configuration, or programming errors, still fail.

## AI Usage

Codex (GPT-5.5) was used after the implementation for self-review.

### DIFF
--- a/packages/fedify/src/federation/middleware.ts
+++ b/packages/fedify/src/federation/middleware.ts
@@ -184,24 +184,34 @@ function isLinkedDataSignatureJsonLdProcessingError(
   error: unknown,
 ): error is Error {
   if (!(error instanceof Error)) return false;
-  if (error.message.startsWith("Maximum deep iterations exceeded")) {
+  const errorName = (error as Error & { name?: unknown }).name;
+  const errorMessage = (error as Error & { message?: unknown }).message;
+  if (
+    typeof errorMessage === "string" &&
+    errorMessage.startsWith("Maximum deep iterations exceeded")
+  ) {
     return true;
   }
   const details = (error as Error & {
     details?: { code?: unknown; cause?: unknown };
   }).details;
+  const errorCause = (error as Error & { cause?: unknown }).cause;
+  const cause = errorCause instanceof Error ? errorCause : details?.cause;
   if (
-    error.name === "jsonld.InvalidUrl" &&
+    errorName === "jsonld.InvalidUrl" &&
     details?.code === "loading remote context failed" &&
-    details.cause instanceof Error
+    cause instanceof Error
   ) {
-    return details.cause.message.startsWith(
-      "Maximum deep iterations exceeded",
-    ) || details.cause.name.startsWith("jsonld.") ||
-      details.cause.name === "UnsafeJsonLdError";
+    const causeName = (cause as Error & { name?: unknown }).name;
+    const causeMessage = (cause as Error & { message?: unknown }).message;
+    return (
+      typeof causeMessage === "string" &&
+      causeMessage.startsWith("Maximum deep iterations exceeded")
+    ) || (typeof causeName === "string" && causeName.startsWith("jsonld."));
   }
-  return error.name.startsWith("jsonld.") ||
-    error.name === "UnsafeJsonLdError";
+  return errorName !== "jsonld.InvalidUrl" &&
+    typeof errorName === "string" &&
+    errorName.startsWith("jsonld.");
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/fedify/src/federation/middleware.ts
```diff
@@ -415,6 +415,30 @@ function isPermanentInboxParseError(error: unknown): error is Error {
...
+    details?.code === "loading remote context failed" &&
+    details.cause instanceof Error
+  ) {
+    return details.cause.message.startsWith(
+      "Maximum deep iterations exceeded",
+    ) || details.cause.name.startsWith("jsonld.") ||
+      details.cause.name === "UnsafeJsonLdError";
+  }
+  return error.name.startsWith("jsonld.") ||
+    error.name === "UnsafeJsonLdError";
```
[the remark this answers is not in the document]
notJoon: https://github.com/fedify-dev/fedify/pull/842/commits/aa439f642417537414b0ae7e75987781f910d605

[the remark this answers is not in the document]
notJoon: https://github.com/fedify-dev/fedify/pull/842/commits/aa439f642417537414b0ae7e75987781f910d605

packages/fedify/src/federation/middleware.ts
```diff
@@ -415,6 +415,40 @@ function isPermanentInboxParseError(error: unknown): error is Error {
         isInvalidUrlTypeError(error)));
 }
 
+function isLinkedDataSignatureJsonLdProcessingError(
```
notJoon: I think this could be simplified by avoiding access to `unknown` fields throughout the body, and handling it with a single structural type. just a thought

dahlia (changes requested):
Could you change the target of this pull request to `2.0-maintenance` rather than `main`?

packages/fedify/src/federation/middleware.ts
dahlia: It might be better to split it into two functions: a [user-defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) which takes an `unknown` value, and an actual predicate function which takes a proper structural type.

dahlia:
> Codex (GPT-5.5) was used after the implementation for self-review.

So, could you put `Assisted-by: Codex:gpt-5.5` trailers in your commit messages? See also our [AI usage policy](https://github.com/fedify-dev/fedify/blob/main/AI_POLICY.md)!

packages/fedify/src/federation/middleware.ts
notJoon: applied the suggestion. thanks! https://github.com/fedify-dev/fedify/pull/842/commits/cf43e9629d8dd0c17bf22ee74a00d0304e9c0fb5

notJoon:
@dahlia I’ve updated all the points you mentioned. Could you please check them? Thank you!

dahlia (approved):
Thanks for your contribution!

PR #844, opened 2026-06-26
Treat malformed actor keys as lookup failures

### LINKED ISSUE #825: Malformed actor public keys should not crash inbox verification
Hackers' Pub is seeing repeated production inbox failures from Fedify 2.2.4 when a remote actor document contains a malformed `publicKey` entry. The Sentry issue is `GRAPHQL-4F` and the event count was 23 when triaged.

Observed behavior
-----------------

The request fails while verifying an incoming ActivityPub inbox request. The stack reaches Fedify's key lookup path in *packages/fedify/src/sig/key.ts*, then `Person.#publicKey_fromJsonLd()` from *@fedify/vocab* throws `TypeError: Expected an object of any type of: https://w3id.org/security#Key`.

In the current implementation, `resolveFetchedKey()` catches `TypeError` while parsing the fetched document as an `Object` or as the requested key class, but once it has parsed an actor and starts iterating `object.getPublicKeys()`, a malformed key emitted by that iterator escapes the function.

Expected behavior
-----------------

A malformed remote `publicKey` should make key resolution fail as an unverifiable or invalid key, not crash the whole inbox request with a 500 response. The behavior should match the existing invalid-object branch in *packages/fedify/src/sig/key.ts*: cache the negative lookup if appropriate, return `null`, and let signature verification fail normally.

Notes
-----

This looks app-independent because the exception is raised before Hackers' Pub's inbox listener code runs. The app can reject malformed senders, but it does not have a clean boundary to convert this specific Fedify key lookup exception into an authentication failure.

notJoon:
## Description

fixes #825 

- Treat malformed actor `publicKey` entries as failed key lookups during inbox signature verification
- Cache the failed lookup as `null` so repeated verification attempts do not reprocess the same malformed key
- Add a regression test for an inline malformed actor `publicKey` object that fails while iterating `getPublicKeys()`

## AI Usage

AI assistance was used to review the fix and write the regression test.

### DIFF
--- a/packages/fedify/src/sig/key.test.ts
+++ b/packages/fedify/src/sig/key.test.ts
@@ -353,11 +353,12 @@ test("fetchKey()", async () => {
 });
 
 test("fetchKey() returns null for a malformed actor publicKey", async () => {
+  const actorId = "https://example.com/malformed-public-key";
   const keyId = "https://example.com/malformed-public-key#main-key";
   const cache: Record<string, CryptographicKey | Multikey | null> = {};
   const options: FetchKeyOptions = {
     async documentLoader(resource) {
-      if (resource === keyId) {
+      if (resource === actorId) {
         return {
           contextUrl: null,
           documentUrl: resource,
@@ -366,15 +367,22 @@ test("fetchKey() returns null for a malformed actor publicKey", async () => {
               "https://www.w3.org/ns/activitystreams",
               "https://w3id.org/security/v1",
             ],
-            id: "https://example.com/malformed-public-key",
+            id: actorId,
             type: "Person",
-            publicKey: {
-              "@context": "https://w3id.org/security/v1",
-              id: keyId,
-              type: "Key",
-              owner: "https://example.com/malformed-public-key",
-              publicKeyPem: "not a public key",
-            },
+            publicKey: keyId,
+          },
+        };
+      }
+      if (resource === keyId) {
+        return {
+          contextUrl: null,
+          documentUrl: resource,
+          document: {
+            "@context": "https://w3id.org/security/v1",
+            id: keyId,
+            type: "Key",
+            owner: actorId,
+            publicKeyPem: "not a public key",
           },
         };
       }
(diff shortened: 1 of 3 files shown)


packages/fedify/src/sig/key.test.ts
```diff
@@ -351,3 +351,54 @@ test("fetchKey()", async () => {
...
+  assertEquals(await fetchKey(keyId, CryptographicKey, options), {
+    key: null,
+    cached: false,
+  });
+  assertEquals(cache, { [keyId]: null });
+  assertEquals(await fetchKey(keyId, CryptographicKey, options), {
+    key: null,
+    cached: true,
+  });
+});
```
dahlia, line 412: The fix in *key.ts* looks correct, but the regression test does not exercise the fixed path.

I checked by reverting the *key.ts* change and running the test again. It still passes. A regression test should fail when the fix is removed; this one has no falsifying power.

The cause is the inline `publicKey`. An inline malformed key throws inside `Object.fromJsonLd()` (`TypeError: Invalid PEM-SPKI format`) before `getPublicKeys()` ever runs. That throw is caught by the existing invalid-object branch, not the `getPublicKeys()` iteration this PR changes. The assertion passes, but through the wrong path.

The issue's case needs the actor to parse cleanly and then fail while iterating keys. That happens when `publicKey` is a URL reference, fetched and parsed lazily. With that fixture, the pre-fix code throws the reported error:

```
TypeError: Expected an object of any type of: https://w3id.org/security#Key
  at Person.#publicKey_fromJsonLd → #fetchPublicKey → getPublicKeys
  at fetchKeyInternal (src/sig/key.ts:338)
```

and the post-fix code returns `{ key: null, cached: false }`.

Could you change the fixture so `publicKey` points to a malformed remote key document instead of an inline object? Then the test fails before the fix and passes after.

dahlia (changes requested):

packages/fedify/src/sig/key.test.ts
notJoon: https://github.com/fedify-dev/fedify/commit/60de00e32ef215ac3b218fc2a6c0e1ffe561238e

I've updated the test as suggested. For self confirmation, I divided the cases by whether suppressError is present and try-catch is applied, and checked that the revised tests cover the problematic paths. Could you take a look to confirm the changes are appropriate? Thank you!

dahlia (approved):
Thanks for your contribution!

Issue #908, opened 2026-06-30
`test:bun` scripts in `cli`/`debugger`/`vocab-tools` lack an explicit `--timeout`, causing flaky failures under mise run test on `2.0-maintenance`


2chanhaeng:
`mise run test` runs the Deno/Node.js/Bun suites in parallel and saturates the CPU. Under that load, `packages/cli/src/startup.test.ts` (which spawns a `node dist/mod.js --help` subprocess) exceeds Bun's default 5000ms per-test timeout and is killed, failing with `null !== 0` "Unhandled error between tests". It never fails under `mise run test:bun` alone.

The cause is that `cli`'s `test:bun` script has no explicit `--timeout`, unlike most packages (`fedify`, `postgres`, `redis`, `sqlite`, … all pass one). The same is missing in `debugger` and `vocab-tools`.

## Proposed Fix

Add an explicit `--timeout` to the `test:bun` scripts of `cli`, `debugger`, and `vocab-tools`, consistent with the rest of the monorepo:

```diff
- "test:bun": "bun test"
+ "test:bun": "bun test --timeout 60000"
```

Test-infrastructure only; no product code changes.

## Note

Observed on a branch based on `2.0-maintenance`, so the missing `--timeout` may already be addressed on `main` or other branches.

PR #909, opened 2026-06-30
Fix collection dispatchers 404 after `build()`

### LINKED ISSUE #849: `setCollectionDispatcher()` routes return 404 after `FederationBuilder.build()`
## Description

A custom collection dispatcher registered with `FederationBuilder.setCollectionDispatcher()` appears not to work on the built `Federation` instance. The route is accepted by the builder, but fetching the registered collection URL returns `404 Not Found`.

This looks like the collection dispatcher state is not copied from the builder into the built federation instance.

## Reproduction

Tested with `@fedify/fedify@2.3.1`:

```ts
import { createFederationBuilder } from "jsr:@fedify/fedify@2.3.1";
import { Collection, Note } from "jsr:@fedify/fedify@2.3.1/vocab";

const builder = createFederationBuilder<void>();

builder.setCollectionDispatcher(
  "test",
  Note,
  "/c/{id}",
  () =>
    new Collection({
      id: new URL("https://example.com/c/x"),
    }),
);

const federation = await builder.build({
  kv: undefined,
  origin: "https://example.com",
});

const request = new Request("https://example.com/c/x", {
  headers: { accept: "application/activity+json" },
});

const response = await federation.fetch(request);

console.log(response.status);
console.log(await response.text());
```

Actual output:

```txt
404
Not Found
```

## Expected behavior

The request to `https://example.com/c/x` should be handled by the custom collection dispatcher registered with `setCollectionDispatcher()`, and should return an ActivityPub `Collection`.

## Actual behavior

The built federation returns `404 Not Found`.

## Suspected cause

`FederationBuilder.build()` clones/copies several mutable fields into the new `FederationImpl`, including the router, object callbacks, actor callbacks, inbox/outbox callbacks, and built-in collection callbacks such as `followers`, `following`, `liked`, and `featured`.

However, it appears not to copy these custom collection dispatcher fields:

- `collectionCallbacks`
- `collectionTypeIds`

The route itself is added to the builder router, but the corresponding collection callback/type maps are missing on the built federation instance, so the route cannot be dispatched.

## Context

I ran into this while implementing a FEP-c0e0 `fedibird:emojiReactions` collection in Hackers' Pub. Registering the collection with `setCollectionDispatcher()` produced a 404 after `build()`. As a workaround, I registered equivalent `Collection`/`CollectionPage` object dispatchers instead.

This does not seem to be a URI Template matching issue. For example, `/ap/notes/{id}` does not match `/ap/notes/123/emoji-reactions`, and `/ap/notes/{id}/emoji-reactions` correctly extracts `id = "123"`. The failure appears specific to custom collection dispatchers not being carried over into the built federation.

2chanhaeng:
Summary
-------

Fixes custom collection dispatchers registered through `FederationBuilder` returning 404 after `build()` (#849).  Also includes a small test-only mitigation for a flaky CLI startup test that fails under `mise run test` (#908).

Related issue
-------------

 -  fixes https://github.com/fedify-dev/fedify/issues/849
 -  relates to https://github.com/fedify-dev/fedify/issues/908 (test-only mitigation; the broader fix is tracked there)

Changes
-------

 -  `FederationBuilderImpl.build()` now copies `collectionCallbacks` and `collectionTypeIds` onto the built federation, alongside the object callbacks and type IDs it already copied.
 -  Added a regression test in *builder.test.ts* that builds a federation with a custom collection dispatcher and asserts both that the callbacks/item types reach the built federation and that the registered route dispatches with 200 instead of 404.
 -  Gave the *startup.test.ts* "CLI starts successfully with --help" test an explicit 60s timeout and made its subprocess `close` handler capture the termination signal for clearer failure messages.

Benefits
--------

 -  Custom collection dispatchers (`setCollectionDispatcher()` / `setOrderedCollectionDispatcher()`) configured via the builder now work end-to-end after `build()`, instead of silently 404ing.
 -  The CLI startup test no longer fails intermittently under the parallel `mise run test` matrix, where Bun's default 5000ms per-test timeout would kill the slow-to-start subprocess and surface a confusing `null !== 0` error.  Capturing the signal also makes any future failure diagnosable.

Additional notes
----------------

 -  Targets `2.0-maintenance` as the oldest branch containing the bug.
 -  The *startup.test.ts* change is a local mitigation only.  The agreed broader fix — adding an explicit `--timeout` to the `test:bun` scripts of `cli`, `debugger`, and `vocab-tools` — is tracked in #908 and intentionally not bundled here.
 -  AI disclosure: Assisted-by: Claude Code:claude-opus-4.8

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,19 @@ Version 2.0.22
 
 To be released.
 
+### @fedify/fedify
+
+ -  Fixed custom collection dispatchers registered through
+    `FederationBuilder.setCollectionDispatcher()` and
+    `setOrderedCollectionDispatcher()` returning 404 Not Found after
+    `build()`.  `build()` now copies the collection callbacks and item types
+    onto the built federation, so the registered routes dispatch their
+    collections instead of being treated as unknown routes.
+    [[#849], [#851] by ChanHaeng Lee]
+
+[#849]: https://github.com/fedify-dev/fedify/issues/849
+[#851]: https://github.com/fedify-dev/fedify/pull/851
+
 ### @fedify/vocab
 
  -  Fixed Activity Vocabulary parsing so malformed language tags in remote
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,7 +12,7 @@ To be released.
 
  -  Fixed custom collection dispatchers registered through
     `FederationBuilder.setCollectionDispatcher()` and
-    `setOrderedCollectionDispatcher()` returning 404 Not Found after
+    `setOrderedCollectionDispatcher()` returning `404 Not Found` after
     `build()`.  `build()` now copies the collection callbacks and item types
     onto the built federation, so the registered routes dispatch their
     collections instead of being treated as unknown routes.
--- a/packages/cli/src/startup.test.ts
+++ b/packages/cli/src/startup.test.ts
@@ -10,7 +10,14 @@ const packageDir = resolve(dirname(fileURLToPath(import.meta.url)), "..");
 async function runCommand(
   command: string,
   args: string[],
-): Promise<{ code: number | null; stdout: string; stderr: string }> {
+): Promise<
+  {
+    code: number | null;
+    signal: NodeJS.Signals | null;
+    stdout: string;
+    stderr: string;
+  }
+> {
   return await new Promise((resolvePromise, reject) => {
     const child = spawn(command, args, {
       cwd: packageDir,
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -8,6 +8,19 @@ Version 2.0.22
 
 To be released.
 
+### @fedify/fedify
+
+ -  Fixed custom collection dispatchers registered through
+    `FederationBuilder.setCollectionDispatcher()` and
+    `setOrderedCollectionDispatcher()` returning 404 Not Found after
```
```suggestion
    `setOrderedCollectionDispatcher()` returning `404 Not Found` after
```

2chanhaeng: Addressed in b8a396a63139d7c16a811cc58538d395015f0471.

dahlia (approved):

PR #913, opened 2026-06-30
Report HTML document-loader failures clearly

### LINKED ISSUE #912: Report HTML document-loader failures without JSON parser crashes
Summary
-------

Hackers' Pub is seeing repeated production failures from Fedify 2.3.1 when a remote document request returns an HTML page without a usable ActivityPub alternate link. The Sentry issues are `GRAPHQL-3E` and `GRAPHQL-2F`; together they account for more than 20 events.

Observed behavior
-----------------

The stack reaches `getRemoteDocument()` in *@fedify/vocab-runtime/src/docloader.ts*. When the response has `Content-Type: text/html`, Fedify scans `<a>` and `<link>` tags for an ActivityPub alternate. If no usable alternate is found, it falls through to `JSON.parse(html)`. That produces a generic `SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON`.

This looks like remote invalid input or a remote server error page, but it surfaces as an application exception in the caller's Sentry project. It also loses useful response context such as status, content type, and document URL from the thrown error message.

Expected behavior
-----------------

If an HTML response has no ActivityPub alternate, Fedify should fail document loading with a typed fetch or document-loader error that says the response was HTML and no ActivityPub alternate was found. That would let applications classify the failure as remote input or remote transport failure instead of a JSON parser crash.

Notes
-----

The relevant code path is in *@fedify/vocab-runtime/src/docloader.ts*. The same path intentionally parses small HTML pages to discover alternate links, so the problematic behavior is specifically the fallback after alternate discovery fails.

Related Sentry issues in Hackers' Pub:

 -  `GRAPHQL-3E`: `getRemoteDocument(@fedify/vocab-runtime/2.3.1/src/docloader)`, 11 events when triaged
 -  `GRAPHQL-2F`: `getRemoteDocument(@fedify/vocab-runtime/2.3.1/src/docloader)`, 10 events when triaged

dahlia:
Fedify already treats HTML as a possible discovery page, not as an immediate failure. That is useful for servers that expose an ActivityPub representation through an alternate link, so this PR keeps that behavior. The problem is the fallback after discovery fails: once the loader has seen an HTML/XHTML response and found no usable ActivityPub alternate, passing the same body to the JSON parser turns a remote HTML page into a generic `SyntaxError`.

This moves the failure to the point where the document loader still has the right context. In *packages/vocab-runtime/src/docloader.ts*, the HTML/XHTML branch now clones the response before reading the body, keeps the existing header and in-document alternate discovery, then throws `FetchError` if no distinct ActivityPub or JSON-LD alternate can be found. Oversized HTML follows the same typed error path because Fedify cannot scan it safely. In both cases the error keeps the document URL, the response, and the Content-Type in the message.

The tests in *packages/vocab-runtime/src/docloader.test.ts* cover a real HTML page with no ActivityPub alternate and assert that the error is a `FetchError` with the original response metadata. The ReDoS regression test still checks that hostile HTML completes quickly, but now expects the same document-loader error instead of relying on the JSON parser. I also narrowed the malformed Link-header test to a JSON-LD response, so it continues to cover Link parser recovery without depending on the old HTML fallback.

*CHANGES.md* records this under `@fedify/vocab-runtime` for the next minor release.

Fixes #912.

### DIFF
--- a/packages/vocab-runtime/src/docloader.ts
+++ b/packages/vocab-runtime/src/docloader.ts
@@ -121,6 +121,13 @@ function createResponseMetadata(response: Response): Response {
   });
 }
 
+async function cancelResponseBody(response: Response): Promise<void> {
+  const body = response.body as { cancel?: unknown } | null;
+  if (body != null && typeof body.cancel === "function") {
+    await body.cancel();
+  }
+}
+
 async function readBoundedText(
   response: Response,
   maxBytes: number,
@@ -129,13 +136,27 @@ async function readBoundedText(
   if (contentLength != null) {
     const size = Number(contentLength);
     if (Number.isFinite(size) && size > maxBytes) {
+      await cancelResponseBody(response);
       return { text: "", size, tooLarge: true };
     }
   }
 
   if (response.body == null) return { text: "", size: 0, tooLarge: false };
 
-  const reader = response.body.getReader();
+  const body = response.body as ReadableStream<Uint8Array> & {
+    getReader?: unknown;
+  };
+  if (typeof body.getReader !== "function") {
+    const text = await response.text();
+    const size = new TextEncoder().encode(text).byteLength;
+    return {
+      text: size <= maxBytes ? text : "",
+      size,
+      tooLarge: size > maxBytes,
+    };
+  }
+
+  const reader = body.getReader();
   const decoder = new TextDecoder();
   let text = "";
   let size = 0;
--- a/packages/vocab-runtime/src/docloader.ts
+++ b/packages/vocab-runtime/src/docloader.ts
@@ -202,13 +202,19 @@ export async function getRemoteDocument(
   ) {
     // Security: Limit HTML response size to mitigate ReDoS attacks
     const MAX_HTML_SIZE = 1024 * 1024; // 1MB
+    const errorResponse = response.clone();
     const html = await response.text();
     if (html.length > MAX_HTML_SIZE) {
       logger.warn(
         "HTML response too large, skipping alternate link discovery: {url}",
         { url: documentUrl, size: html.length },
       );
-      document = JSON.parse(html);
+      throw new FetchError(
+        documentUrl,
+        `HTML document is too large to scan for an ActivityPub alternate link ` +
+          `(Content-Type: ${contentType})`,
+        errorResponse,
+      );
     } else {
       // Safe regex patterns without nested quantifiers to prevent ReDoS
       // (CVE-2025-68475)
@@ -247,7 +253,12 @@ export async function getRemoteDocument(
           return await fetch(new URL(attribs.href, docUrl).href);
         }
       }
-      document = JSON.parse(html);
+      throw new FetchError(
+        documentUrl,
+        `HTML document has no ActivityPub alternate link ` +
+          `(Content-Type: ${contentType})`,
+        errorResponse,
+      );
     }
   } else {
     document = await response.json();
--- a/packages/vocab-runtime/src/docloader.ts
+++ b/packages/vocab-runtime/src/docloader.ts
@@ -122,9 +122,8 @@ function createResponseMetadata(response: Response): Response {
 }
 
 async function cancelResponseBody(response: Response): Promise<void> {
-  const body = response.body as { cancel?: unknown } | null;
-  if (body != null && typeof body.cancel === "function") {
-    await body.cancel();
+  if (response.body != null) {
+    await response.body.cancel();
   }
 }
 
@@ -135,7 +134,7 @@ async function readBoundedText(
   const contentLength = response.headers.get("Content-Length");
   if (contentLength != null) {
     const size = Number(contentLength);
-    if (Number.isFinite(size) && size > maxBytes) {
+    if (size > maxBytes) {
       await cancelResponseBody(response);
       return { text: "", size, tooLarge: true };
     }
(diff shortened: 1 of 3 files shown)


dahlia:
@codex review

packages/vocab-runtime/src/docloader.ts
```diff
@@ -247,7 +253,12 @@ export async function getRemoteDocument(
           return await fetch(new URL(attribs.href, docUrl).href);
         }
       }
-      document = JSON.parse(html);
+      throw new FetchError(
+        documentUrl,
+        `HTML document has no ActivityPub alternate link ` +
+          `(Content-Type: ${contentType})`,
+        errorResponse,
```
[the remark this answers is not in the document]
dahlia: Addressed in b00ea874: HTML-labelled JSON responses now keep the old compatibility path when the bounded body starts like a JSON object or array.

packages/vocab-runtime/src/docloader.ts
```diff
@@ -202,13 +202,19 @@ export async function getRemoteDocument(
...
         "HTML response too large, skipping alternate link discovery: {url}",
         { url: documentUrl, size: html.length },
       );
-      document = JSON.parse(html);
+      throw new FetchError(
+        documentUrl,
+        `HTML document is too large to scan for an ActivityPub alternate link ` +
+          `(Content-Type: ${contentType})`,
+        errorResponse,
+      );
```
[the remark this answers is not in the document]
dahlia: Addressed in b00ea874: HTML responses are now read with a byte limit, and oversized Content-Length values short-circuit before buffering the body.

dahlia:
@codex review

packages/vocab-runtime/src/docloader.ts
```diff
@@ -112,6 +113,52 @@ export type AuthenticatedDocumentLoaderFactory = (
...
+
+async function readBoundedText(
+  response: Response,
+  maxBytes: number,
+): Promise<{ text: string; size: number; tooLarge: boolean }> {
+  const contentLength = response.headers.get("Content-Length");
+  if (contentLength != null) {
+    const size = Number(contentLength);
+    if (Number.isFinite(size) && size > maxBytes) {
+      return { text: "", size, tooLarge: true };
```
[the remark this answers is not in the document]
dahlia: Addressed in c4db929f: oversized Content-Length responses now cancel the response body before returning the too-large result.

dahlia:
@codex review

sij411 (dismissed):
I approve this since compatibility concern was already handled.

packages/vocab-runtime/src/docloader.ts
```diff
@@ -112,6 +113,73 @@ export type AuthenticatedDocumentLoaderFactory = (
...
+function createResponseMetadata(response: Response): Response {
+  return new Response(null, {
+    headers: response.headers,
+    status: response.status,
+    statusText: response.statusText,
+  });
+}
+
+async function cancelResponseBody(response: Response): Promise<void> {
+  const body = response.body as { cancel?: unknown } | null;
```
2chanhaeng: I think the `as { cancel?: unknown } | null` assertion is unnecessary because DLS infers `body` as `ReadableStream<Uint8Array<ArrayBuffer>> | null` and the `ReadableStream` interface has a `cancel(reason?: any): Promise<void>` method.

packages/vocab-runtime/src/docloader.ts
```diff
@@ -112,6 +113,73 @@ export type AuthenticatedDocumentLoaderFactory = (
...
+      await cancelResponseBody(response);
+      return { text: "", size, tooLarge: true };
+    }
+  }
+
+  if (response.body == null) return { text: "", size: 0, tooLarge: false };
+
+  const body = response.body as ReadableStream<Uint8Array> & {
+    getReader?: unknown;
+  };
```
2chanhaeng: Also this assertion is unnecessary, too.

packages/vocab-runtime/src/docloader.ts
```diff
@@ -112,6 +113,73 @@ export type AuthenticatedDocumentLoaderFactory = (
...
+  if (typeof body.getReader !== "function") {
+    const text = await response.text();
+    const size = new TextEncoder().encode(text).byteLength;
+    return {
+      text: size <= maxBytes ? text : "",
+      size,
+      tooLarge: size > maxBytes,
+    };
+  }
```
2chanhaeng: If `body` is not `null`, it's `ReadableStream<Uint8Array<ArrayBuffer>>` and `ReadableStream` has `getReader` method. Therefore, this block is unreachable.

packages/vocab-runtime/src/docloader.ts
```diff
@@ -247,7 +319,17 @@ export async function getRemoteDocument(
...
+      if (trimmed.startsWith("{") || trimmed.startsWith("[")) {
+        document = JSON.parse(html.text);
+      } else {
+        throw new FetchError(
+          documentUrl,
+          `HTML document has no ActivityPub alternate link ` +
+            `(Content-Type: ${contentType})`,
+          errorResponse,
+        );
+      }
```
2chanhaeng: How about just:
```ts
try {
  document = JSON.parse(html.text);
} catch (error) {
  if (error instanceof SyntaxError) {
    throw new FetchError(/* ... */);
  }
  throw error;
}

File: packages/vocab-runtime/src/docloader.ts [hunk 8]
```diff
@@ -112,6 +113,73 @@ export type AuthenticatedDocumentLoaderFactory = (
...
+}
+
+async function readBoundedText(
+  response: Response,
+  maxBytes: number,
+): Promise<{ text: string; size: number; tooLarge: boolean }> {
+  const contentLength = response.headers.get("Content-Length");
+  if (contentLength != null) {
+    const size = Number(contentLength);
+    if (Number.isFinite(size) && size > maxBytes) {
```
2chanhaeng: If `contentLength = Infinity` is possible, this filter would ignore it. Or else, `Number.isFinite` is unnecessary. I think `size < 0` would be more useful instead.

2chanhaeng (commented):
Some suggestions?

packages/vocab-runtime/src/docloader.ts
dahlia: Addressed in 359f2707: the body cancellation path now uses the standard Response body type directly.

dahlia:
@codex review

sij411 (approved):

PR #952, opened 2026-07-18
Run `format` before `format:check` in init hydration test validation

### LINKED ISSUE #950: Run `format` before `format:check` in `fedify init`'s hydration test validation
When running `mise run test:init` for fedify init, the hydration test only runs `format:check` (and `lint`) without running `format` first. When the package manager is `npm` or `pnpm`, `format:check` fails, which causes the entire test suite to fail.

## Proposed Fix

Run `format` (auto-fix) before `format:check` in the hydration test validation, so that freshly scaffolded projects are already compliant with the formatter before being checked.

Palcimer:
Summary
-------
Current `validateDevToolScripts` only runs `format:check` and `lint` after scaffolding a project, without auto-formatting it first.  Since neither Fedify's own generated files nor most scaffolding CLI output exactly match the project's chosen formatter's style, `format:check` almost always failed for `npm`/`pnpm` package managers, causing the entire hydration test suite to fail regardless of the actual generated code. Fixed by running `format` script before `format:check` so that scaffolded projects are already compliant with the formatter before being checked.

Assisted-by: Claude Code:claude-sonnet-5

Related issue
-------------

 -  closes #950 

Changes
-------

 -  Added `format` script before `format:check` in `validateDevToolScripts` function.

Benefits
--------

 -  Test suites for package managers `npm`/`pnpm` will not fail because of format issue.

Checklist
---------

 -  [] ~~Did you write some relevant docs about this change (if it's a new feature)?~~
 -  [] Did you write a regression test to reproduce the bug (if it's a bug fix)?
 -  [] ~~Did you write some tests for this change (if it's a new feature)?~~

### DIFF
--- a/packages/init/src/test/create.ts
+++ b/packages/init/src/test/create.ts
@@ -155,7 +155,7 @@ async function validateDevToolScripts(
     if (format.code !== 0) return false;
   }
 
-  for (const script of ["format:check", "lint"]) {
+  for (const script of ["format", "format:check", "lint"]) {
     const result = await $`${[packageManager, "run", script]}`
       .cwd(dir)
       .stdin("null")
--- a/packages/init/src/test/create.ts
+++ b/packages/init/src/test/create.ts
@@ -134,7 +134,7 @@ async function validateDevToolScripts(
   dir: string,
   options: GeneratedType<ReturnType<typeof generateTestCases>>,
 ): Promise<boolean> {
-  const [webFramework, packageManager] = options as [
+  const [_, packageManager] = options as [
     WebFramework,
     PackageManager,
     KvStore,
@@ -142,18 +142,6 @@ async function validateDevToolScripts(
   ];
   if (packageManager === "deno") return true;
   if (!(await hasInstalledNodeDependencies(dir))) return true;
-
-  if (webFramework === "astro") {
-    const format = await $`${[packageManager, "run", "format"]}`
-      .cwd(dir)
-      .stdin("null")
-      .stdout("piped")
-      .stderr("piped")
-      .noThrow()
-      .spawn();
-    await saveOutputs(dir, format);
-    if (format.code !== 0) return false;
-  }
 
   for (const script of ["format", "format:check", "lint"]) {
     const result = await $`${[packageManager, "run", script]}`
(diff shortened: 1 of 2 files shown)


packages/init/src/test/create.ts
```diff
@@ -143,19 +143,7 @@ async function validateDevToolScripts(
...
-      .stdout("piped")
-      .stderr("piped")
-      .noThrow()
-      .spawn();
-    await saveOutputs(dir, format);
-    if (format.code !== 0) return false;
-  }
-
-  for (const script of ["format:check", "lint"]) {
+  for (const script of ["format", "format:check", "lint"]) {
```
2chanhaeng, line 146: The commands are not necessary, so wrapping the code with `try { /* ... */ } catch { /* pass */ }` can prevent unnecessary exits.

2chanhaeng (changes requested):

packages/init/src/test/create.ts
Palcimer: I want to make sure I understand what you mean by "unnecessary exits". Do you mean only createTestApp being marked as failed failing, or does it also include the runServerAndLookup test not proceeding properly(because the `testDir` isn't passed to the function)?

2chanhaeng: Both — and the second one is the part I'm actually worried about.

When `validateDevToolScripts()` returns `false`, `createTestApp()` returns `""`, and `runServerAndLookupUser()` drops falsy entries with `dirs.filter(Boolean)`. So a failing `format`, `format:check`, or `lint` doesn't just mark the case as `Fail`; it also silently removes that case from the lookup test.

Those three scripts pass today, but they can start failing later for reasons that have nothing to do with the generated app being correct — a formatter or linter version bump, a new default rule, or template output that no longer matches the formatter's preference. In that situation the generated app is still fine, and we would lose lookup coverage for the case because of a cosmetic check.

Looking at the code again while re-reviewing, I think instead of using `try ... catch`, just change the `if (result.code !== 0) return false` on line 155 to something like `if (result.code !== 0) printMessage\`    Warning: \\\`${script}\\\` failed.\`;`. Could you fix this part?

2chanhaeng (changes requested):

packages/init/src/test/create.ts
Palcimer: Sure, thanks!

2chanhaeng (approved):

2chanhaeng:
Great! Before merging this PR, please rebase from the main branch.

2chanhaeng (approved):

dahlia (approved):

PR #957, opened 2026-07-19
Handle non-IP DNS lookup results

### LINKED ISSUE #956: Cloudflare Workers: document loader rejects public CNAME-backed hosts as private addresses
## Summary

Fedify's document loader rejects valid public ActivityPub URLs hosted on CNAME-backed custom domains when running on Cloudflare Workers with `nodejs_compat`.

For example, this public ActivityPub Article cannot be loaded:

https://writings.hongminhee.org/ap/articles/2026/07/fedified-blog

Fedify throws:

```text
UrlError: Invalid or private address: hongminhee-writings.netlify.app.
```

`writings.hongminhee.org` is a public hostname whose DNS record points to `hongminhee-writings.netlify.app` through a CNAME.

## Minimal reproduction

```ts
import { getDocumentLoader } from "@fedify/vocab-runtime";

const loader = getDocumentLoader();

await loader(
  "https://writings.hongminhee.org/ap/articles/2026/07/fedified-blog",
);
```

Equivalent Wrangler configuration:

```jsonc
{
  "compatibility_date": "2026-06-16",
  "compatibility_flags": ["nodejs_compat"]
}
```

## Actual behavior

The call fails before the document is fetched:

```text
UrlError: Invalid or private address: hongminhee-writings.netlify.app.
```

## Runtime scope and likely root cause

Fedify's `validatePublicUrl()` calls:

```ts
addresses = await lookup(hostname, { all: true });
```

and then assumes that every `LookupAddress.address` is an IPv4 or IPv6 literal:

https://github.com/fedify-dev/fedify/blob/main/packages/vocab-runtime/src/url.ts

Under Cloudflare Workers' `node:dns` compatibility implementation, the lookup can return the CNAME target hostname—`hongminhee-writings.netlify.app.`—in the `address` field instead of a final IP address.

That hostname is passed to `validatePublicIpAddress()`, which rejects it because it is neither an IPv4 nor an IPv6 literal.

The same lookup under regular Node.js resolves to final IP addresses and does not exhibit this behavior.

## Expected behavior

Fedify should successfully load the public ActivityPub document after securely resolving the CNAME target to its final A/AAAA records.

All resolved addresses should still be checked against Fedify's private and reserved address restrictions.

## AI disclosure

This issue draft was created using codex:5.6-sol and fixed by @SJang1

SJang1:
Fixes #956

Ignore non-IP aliases emitted by Node-compatible DNS implementations when a lookup also returns actual IP addresses. Fail closed for alias-only results and continue checking every returned IP.

Assisted-by: Codex:gpt-5

### DIFF
--- a/packages/vocab-runtime/src/url.ts
+++ b/packages/vocab-runtime/src/url.ts
@@ -60,10 +60,17 @@ export async function validatePublicUrl(url: string): Promise<void> {
 /**
  * Validates the IP addresses returned by `node:dns.lookup()`.
  *
- * Some Node.js-compatible runtimes include CNAME records in lookup results,
- * even though Node.js specifies that the `address` field contains an IPv4 or
- * IPv6 address. Ignore those aliases as long as the lookup also returned an
- * actual IP address, while continuing to reject every non-public IP address.
+ * Cloudflare Workers' `node:dns` implementation currently maps every record
+ * in a DNS-over-HTTPS `Answer` array—including CNAME records—to a
+ * `LookupAddress`, even though Node.js specifies that the `address` field must
+ * contain an IPv4 or IPv6 literal.  See:
+ * https://github.com/cloudflare/workerd/issues/6886
+ *
+ * Work around that bug by ignoring non-IP entries only when the lookup also
+ * returns at least one actual IP address.  This remains fail-closed: a result
+ * containing no IP addresses is rejected, and every returned IP address is
+ * still validated and must be public.  This workaround can be revisited once
+ * the Workerd issue is fixed in supported Cloudflare Workers runtimes.
  *
  * @internal
  */
--- a/packages/vocab-runtime/src/url.ts
+++ b/packages/vocab-runtime/src/url.ts
@@ -54,8 +54,31 @@ export async function validatePublicUrl(url: string): Promise<void> {
   } catch {
     addresses = [];
   }
-  for (const { address, family } of addresses) {
+  validateLookupAddresses(addresses);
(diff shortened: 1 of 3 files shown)


packages/vocab-runtime/src/url.ts
```diff
@@ -51,12 +51,35 @@ export async function validatePublicUrl(url: string): Promise<void> {
...
+    throw new UrlError("DNS lookup failed", { cause: error });
   }
-  for (const { address, family } of addresses) {
+  validateLookupAddresses(addresses);
+}
+
+/**
+ * Validates the IP addresses returned by `node:dns.lookup()`.
+ *
+ * Some Node.js-compatible runtimes include CNAME records in lookup results,
```
dahlia: Could you make it explicit here that this is a workaround for a bug in Cloudflare Workers' `node:dns` implementation, rather than a general variation among Node-compatible runtimes? Workerd currently maps every record in the DNS-over-HTTPS `Answer` array—including CNAME records—to `LookupAddress`, even though Node.js specifies that `address` is an IPv4 or IPv6 literal.

Please link the upstream Workerd issue once it is filed, and explain why this remains fail-closed: non-IP entries may be ignored only when at least one actual IP address is present, and every returned IP must still be validated. That context will help future readers understand why the spec-violating input is tolerated and when the workaround can be revisited. It would also be helpful to rename the test to mention Cloudflare Workers so the regression's origin is visible there as well.

dahlia (approved):

PR #972, opened 2026-07-27
Avoid duplicate response headers on Elysia 1.4.18 and earlier

### LINKED ISSUE #970: @fedify/elysia duplicates response headers on Elysia 1.4.18 and earlier
The `@fedify/elysia` `fedify()` plugin duplicates every HTTP response header that Fedify produces when the app runs on Elysia 1.4.18 or earlier.  A federation response such as an actor document is sent with a duplicated `Content-Type`, so a strict ActivityPub peer can reject it.

Environment
-----------

 -  `@fedify/elysia`: the plugin request handling that causes this is present in every published version and on every branch from *1.8-maintenance* through *main*.
 -  Elysia: 1.4.18 and earlier.  Elysia 1.4.19 and later mask the bug (see below).
 -  Runtime: reproduced on Deno, Bun, and Node.
 -  Operating system: not OS-specific (reproduced on macOS).

Steps to reproduce
------------------

Register a real actor dispatcher, mount the plugin, and read the response headers.  Pin `elysia` to 1.4.18 or earlier (for example `elysia@1.3.8`); `app.handle()` is enough, so no HTTP server is needed:

~~~~ ts
import { createFederation, MemoryKvStore } from "@fedify/fedify";
import { Person } from "@fedify/vocab";
import { Elysia } from "elysia";
import { fedify } from "@fedify/elysia";

const federation = createFederation<void>({ kv: new MemoryKvStore() });
federation.setActorDispatcher(
  "/users/{identifier}",
  (ctx, id) =>
    id === "alice"
      ? new Person({ id: ctx.getActorUri(id), preferredUsername: id })
      : null,
);

const app = new Elysia().use(fedify(federation, () => undefined));
const res = await app.handle(
  new Request("http://localhost/users/alice", {
    headers: { Accept: "application/activity+json" },
  }),
);
console.log(res.headers.get("content-type"));
// => "application/activity+json, application/activity+json"
~~~~

For a fuller reproduction that serves this over Bun, Deno, and Node and shows the raw wire headers, see the [reproduction repository].

[reproduction repository]: https://github.com/dktsudgg/elysia-header-duplication

Expected behavior
-----------------

The federation response is forwarded as-is, with each header appearing once:

~~~~ http
HTTP/1.1 200 OK
Content-Type: application/activity+json
Vary: Accept
~~~~

Actual behavior
---------------

Every header Fedify sets is duplicated.  `Content-Type` and `Vary` each come back twice:

~~~~ http
HTTP/1.1 200 OK
Content-Type: application/activity+json, application/activity+json
Vary: Accept, Accept
~~~~

Cause
-----

In *packages/elysia/src/index.ts*, the `onRequest` handler copies every header of the federation response into `set.headers` and then returns the same `response`:

~~~~ ts
response.headers.forEach((value, key) => {
  set.headers[key] = value;
});

if (response.body) {
  return response;
}
~~~~

When the response has a body it is returned directly, but it already carries its own headers, so each header now lives both on the response and in `set.headers`.  Elysia merges `set.headers` into the returned response, and on Elysia 1.4.18 and earlier that merge appends unconditionally, so every header is duplicated.

Why Elysia 1.4.18 and earlier
-----------------------------

Elysia 1.4.19 changed its header merge to skip headers the response already carries.  Its release notes list the fix under the bug-fix section: “#1591, #1590 merge set.headers without duplicating Response by @truehazker” ([Elysia 1.4.19 release notes]).  On 1.4.19 and later the same plugin code no longer duplicates, which is why the steps above pin an Elysia at or below 1.4.18.  1.3.8 is within the range `@fedify/elysia` supports (`^1.3.6` in *package.json*, `^1.3.8` in the Deno import map and pnpm catalog), so the duplication happens on a supported configuration.

[Elysia 1.4.19 release notes]: https://github.com/elysiajs/elysia/releases/tag/1.4.19

Proposed fix
------------

Move the early return above the header copy, so the copy only runs on the bodyless path where the plugin builds a fresh `new Response(null, …)` that has no headers of its own:

~~~~ diff
diff --git a/packages/elysia/src/index.ts b/packages/elysia/src/index.ts
--- a/packages/elysia/src/index.ts
+++ b/packages/elysia/src/index.ts
@@ -42,15 +42,15 @@ export const fedify = <TContextData = unknown>(
         if (!notFound && !notAcceptable) {
           set.status = response.status;
 
-          response.headers.forEach((value, key) => {
-            set.headers[key] = value;
-          });
-
           // Return response body if it exists
           if (response.body) {
             return response;
           }
 
+          response.headers.forEach((value, key) => {
+            set.headers[key] = value;
+          });
+
           // Return empty response for successful requests without body
           return new Response(null, { status: response.status });
         }
~~~~

With this change the response with a body is returned untouched, so there is nothing for Elysia to merge back in, and the headers are no longer duplicated on any Elysia version.

Notes
-----

Per the branch policy this bug fix should target the oldest maintenance branch that contains it; the plugin logic is present at least back to *1.8-maintenance*, so I would like to confirm the target branch before opening the pull request.

dktsudgg:
Fixes #970.

`@fedify/elysia`'s `fedify()` plugin copies the federation response headers into `set.headers` and then returns the same `Response` object.

When the response has a body, that `Response` already carries its own headers,
so on Elysia 1.4.18 and earlier(where Elysia merges `set.headers` into the returned response by appending unconditionally) every response header ends up duplicated.
(for example : `Content-Type: application/activity+json, application/activity+json`)

Elysia 1.4.19 changed the merge to skip headers the response already has, which masks the bug, but the package still supports Elysia `^1.3.6`, so users on 1.4.18 and earlier receive malformed federation responses that a strict ActivityPub peer can reject.

This change moves the early `return response` (the case where the response has a body) above the header copy,
so the copy only runs on the bodyless path where a fresh `new Response(null, …)` needs the headers.
When there is a body the `Response` is returned untouched, so there is nothing for Elysia to merge back in, and the headers are no longer duplicated on any Elysia version.

See #970 for the details.

Assisted-by: Claude Code:claude-fable-5

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@ To be released.
  -  Fixed duplicate response headers on Elysia 1.4.18 and earlier, which
     append both `set.headers` and the returned `Response`'s own headers
     without deduplication.  The `fedify()` plugin no longer sets the headers
-    in both places.  [[#970], [#972]]
+    in both places.  [[#970], [#972] by dktsudgg]
 
 [#970]: https://github.com/fedify-dev/fedify/issues/970
 [#972]: https://github.com/fedify-dev/fedify/pull/972
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,16 @@ Version 1.8.16
 
 To be released.
 
+### @fedify/elysia
+
+ -  Fixed duplicate response headers on Elysia 1.4.18 and earlier, which
+    append both `set.headers` and the returned `Response`'s own headers
+    without deduplication.  The `fedify()` plugin no longer sets the headers
+    in both places.  [[#970], [#972]]
+
+[#970]: https://github.com/fedify-dev/fedify/issues/970
+[#972]: https://github.com/fedify-dev/fedify/pull/972
+
 
 Version 1.8.15
 --------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@ To be released.
  -  Fixed duplicate response headers on Elysia 1.4.18 and earlier, which
     append both `set.headers` and the returned `Response`'s own headers
     without deduplication.  The `fedify()` plugin no longer sets the headers
-    in both places.  [[#970], [#972] by dktsudgg\]
+    in both places.  [[#970], [#972] by Kyujin Lim\]
 
 [#970]: https://github.com/fedify-dev/fedify/issues/970
 [#972]: https://github.com/fedify-dev/fedify/pull/972
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,7 +16,7 @@ To be released.
     in both places.  [[#970], [#972] by dktsudgg\]
 
 [#970]: https://github.com/fedify-dev/fedify/issues/970
-[#972]: https://github.com/fedify-dev/fedify/issues/972
+[#972]: https://github.com/fedify-dev/fedify/pull/972
 
 
 Version 2.0.24
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,16 @@ Version 2.0.25
 
 To be released.
 
+### @fedify/elysia
+
+ -  Fixed duplicate response headers on Elysia 1.4.18 and earlier, which
+    append both `set.headers` and the returned `Response`'s own headers
+    without deduplication.  The `fedify()` plugin no longer sets the headers
+    in both places.  [[#970], [#972] by dktsudgg\]
+
+[#970]: https://github.com/fedify-dev/fedify/issues/970
+[#972]: https://github.com/fedify-dev/fedify/issues/972
+
 
 Version 2.0.24
 --------------
--- a/changes.d/elysia/duplicate-response-headers.md
+++ b/changes.d/elysia/duplicate-response-headers.md
@@ -1,4 +1,4 @@
  -  Fixed duplicate response headers on Elysia 1.4.18 and earlier, which
     append both `set.headers` and the returned `Response`'s own headers
     without deduplication.  The `fedify()` plugin no longer sets the headers
-    in both places.  [[#970], [#972] by dktsudgg]
+    in both places.  [[#970], [#972] by Kyujin Lim]
--- a/changes.d/elysia/duplicate-response-headers.md
+++ b/changes.d/elysia/duplicate-response-headers.md
@@ -1,0 +1,4 @@
+ -  Fixed duplicate response headers on Elysia 1.4.18 and earlier, which
+    append both `set.headers` and the returned `Response`'s own headers
+    without deduplication.  The `fedify()` plugin no longer sets the headers
+    in both places.  [[#970], [#972] by dktsudgg]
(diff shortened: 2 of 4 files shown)


dahlia (changes requested):
Could you add a changelog entry to *CHANGES.md*? Thanks!

dktsudgg:
@dahlia I added a changelog entry!

CHANGES.md
```diff
@@ -8,6 +8,16 @@ Version 1.8.16
 
 To be released.
 
+### @fedify/elysia
+
+ -  Fixed duplicate response headers on Elysia 1.4.18 and earlier, which
+    append both `set.headers` and the returned `Response`'s own headers
+    without deduplication.  The `fedify()` plugin no longer sets the headers
+    in both places.  [[#970], [#972]]
```
dahlia: Add your name as well for the credit!

dktsudgg: @dahlia I added it. Thank you!

dahlia:
@dktsudgg Could you change the target of this pull request to `2.0-maintenance`?  1.8.x is now end-of-life.

dktsudgg:
@dahlia I changed target branch to `2.0-maintenance`

changes.d/elysia/duplicate-response-headers.md
```diff
@@ -0,0 +1,4 @@
+ -  Fixed duplicate response headers on Elysia 1.4.18 and earlier, which
+    append both `set.headers` and the returned `Response`'s own headers
+    without deduplication.  The `fedify()` plugin no longer sets the headers
+    in both places.  [[#970], [#972] by dktsudgg]
```
dahlia: We prefer using real names here rather than pseudonyms. Unless you have a strong preference for a pseudonym, could you please use your real name instead?

dktsudgg: @dahlia I changed it to my real name!
(Addressed in 9d0061a)

dahlia:
Could you rebase your commits on the latest `2.0-maintenance`?

dahlia:
Sorry, the fix I made on `2.0-maintenance` was half-baked.  Could you make a rebase once again?

dktsudgg:
Sorry, and I found that I mistyped like `https://github.com/fedify-dev/fedify/issues/972`,
I fixed it to `https://github.com/fedify-dev/fedify/pull/972`
(Addressed in 655f2dc)

dktsudgg:
I added two commits with AI assistance to make `mise run check` pass.
please check e233750 and 9cc1d9e.
I confirmed it works locally, but the commit messages are AI-generated, so please give them a read.

It looks like there is a real incompatibility between sacho and hongdown rather than a one-off formatting issue.
hongdown 0.5.3 requires a blank line after the front matter's closing `---`, while sacho's normal form forbids one,
so a changelog fragment with a links block cannot satisfy both checkers at once.
`main` branch works around this by excluding `changes.d/` from Hongdown in *.hongdown.toml*, which is what 9cc1d9e backports.

image

dahlia (approved):
Sorry for the late response.  Looks good to me.  Thanks!

Issue #870, opened 2026-06-30
Test custom error handlers in `@fedify/next`


dahlia:
*@fedify/next* accepts custom federation fetch handlers so apps can override the default 404 and 406 responses. That extension point should have a direct regression test.

Add a test in *packages/next/src/index.test.ts* that verifies `integrateFederation()` passes custom `onNotFound` or `onNotAcceptable` handlers to `federation.fetch()`.

Suggested check: `mise run check-each next`.

Bananamilk452:
Can i work with this issue?

dahlia:
@Bananamilk452 Assigned! Thanks.

Issue #869, opened 2026-06-30
Test non-federation delegation in `@fedify/next`


dahlia:
*@fedify/next* should leave ordinary Next.js requests to the user middleware. That behavior keeps the integration from taking over the whole application.

Add a test in *packages/next/src/index.test.ts* that verifies `fedifyWith()` calls the user middleware when `isFederationRequest()` is false.

Suggested check: `mise run check-each next`.

hugosmoreira:
Hi! I’d like to work on this issue. I’ll add the focused regression test in `packages/next/src/index.test.ts` and run `mise run check-each next`. I’ll use Codex as an implementation assistant and disclose that assistance in the commit and PR. Please assign me if this scope looks good.

dahlia:
@hugosmoreira Sorry, this issue is reserved for OSSCA 2026 mentees.

Bananamilk452:
Can i work with this issue?

dahlia:
@Bananamilk452 Assigned, thanks!

Issue #868, opened 2026-06-30
Test NodeInfo route detection in `@fedify/next`


dahlia:
*@fedify/next* treats NodeInfo discovery paths as federation-related even without ActivityPub headers. A small test would keep those well-known endpoints from being routed to user middleware by mistake.

Add a test in *packages/next/src/index.test.ts* that verifies `isNodeInfoRequest()` matches `/.well-known/nodeinfo` and `/.well-known/x-nodeinfo2`.

Suggested check: `mise run check-each next`.

Bananamilk452:
Can i work with this issue?

dahlia:
@Bananamilk452 Assigned, thanks!

Issue #867, opened 2026-06-30
Test ActivityPub `Content-Type` detection in `@fedify/next`


dahlia:
*@fedify/next* also routes inbound ActivityPub requests by checking `Content-Type`. That matters for inbox POST requests, not just object fetches.

Add a test in *packages/next/src/index.test.ts* that verifies `isFederationRequest()` returns true for `Content-Type: application/activity+json`.

Suggested check: `mise run check-each next`.

Bananamilk452:
Can i work with this issue?

dahlia:
@Bananamilk452 Sure, just assigned.  Thanks!

PR #978, opened 2026-08-01
Fix fedify/next test-each failure, add fedify/next tests

### LINKED ISSUE #976: fedify/next mise test-each doesn't works well
Image

It also checks other modules file not only source files.

If it's ok, i will work this with my @fedify/next test issues.

Bananamilk452:
Closes #976, #869, #870, #868, #866, #867

## Description 

- Added a comprehensive test suite in `packages/next/src/index.test.ts` to cover federation request detection, NodeInfo route handling, middleware delegation, and custom error handling logic.
- Modified the `mise.toml` test runner to only run Deno tests if a `deno.json` file exists for the package, and to always run Node.js and Bun tests with clear output messages.

Github Copilot helped me making title of this PR, gpt-5.6-terra helped checking code quality and code creation.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -304,11 +304,15 @@ shell = "nu -c"
 run = '''
 let pkgs = ($env.usage_packages | split row " " | where ($it | is-not-empty))
 mise run prepare-each ...$pkgs
-mise run check-each ...$pkgs
 for pkg in $pkgs {
-  print $"Running tests for package: ($pkg)"
-  deno task --filter $"@fedify/($pkg)" test
+  if ($"packages/($pkg)/deno.json" | path exists) {
+    mise run check-each $pkg
+    print $"Running Deno tests for package: ($pkg)"
+    deno task --filter $"@fedify/($pkg)" test
+  }
+  print $"Running Node.js tests for package: ($pkg)"
   pnpm --filter $"@fedify/($pkg)" test
+  print $"Running Bun tests for package: ($pkg)"
   pnpm --filter $"@fedify/($pkg)" test:bun
 }
 '''
--- a/mise.toml
+++ b/mise.toml
@@ -310,10 +310,12 @@ for pkg in $pkgs {
     print $"Running Deno tests for package: ($pkg)"
     deno task --filter $"@fedify/($pkg)" test
   }
-  print $"Running Node.js tests for package: ($pkg)"
-  pnpm --filter $"@fedify/($pkg)" test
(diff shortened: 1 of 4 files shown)


2chanhaeng:
Next.js does not work well with Deno. **Seriously.** Instead of adding *deno.json* to `@fedify/next` to enable Deno support, could you modify it to exclude cases where *deno.json* is missing within the package path received during `test-each`/`check-each` tasks, or when the package path is not included in the `workspace` of the root *deno.json*?

mise.toml
```diff
@@ -304,11 +304,15 @@ shell = "nu -c"
...
-  deno task --filter $"@fedify/($pkg)" test
+  if ($"packages/($pkg)/deno.json" | path exists) {
+    mise run check-each $pkg
+    print $"Running Deno tests for package: ($pkg)"
+    deno task --filter $"@fedify/($pkg)" test
+  }
+  print $"Running Node.js tests for package: ($pkg)"
   pnpm --filter $"@fedify/($pkg)" test
+  print $"Running Bun tests for package: ($pkg)"
   pnpm --filter $"@fedify/($pkg)" test:bun
```
2chanhaeng: Could you add checking *package.json* for Node and Bun? For example, `@fedify/fresh` doesn't have *package.json* because [Fresh](https://usefresh.dev/) is made for Deno and JSR.

2chanhaeng (changes requested):
Great work! Please check the point that I commented on.

packages/next/package.json
```diff
@@ -58,6 +58,8 @@
     "build:self": "tsdown",
     "build": "pnpm --filter @fedify/next... run build:self",
     "prepack": "pnpm build",
-    "prepublish": "pnpm build"
+    "prepublish": "pnpm build",
+    "test": "node --experimental-transform-types --test",
```
2chanhaeng: Testing TS files directly doesn't work well with Next.js, I think. How about testing built files only? Like `node --test 'dist/'`.

2chanhaeng (changes requested):
Please add `packages/next` to `test.exclude` in root *deno.json*

Bananamilk452:
I made a change! Please take an another look!

2chanhaeng (dismissed):
Great work! Thanks for your contribution!

2chanhaeng:
Oh, before merging this PR, please rebase from the main branch. Thanks!

2chanhaeng (approved):

dahlia (approved):

PR #979, opened 2026-08-01
Align `Follow` tutorials with Fedify's origin-based security model

dahlia:
Fedify's `Follow` examples were written before the origin-based security model from [FEP-fe34] was introduced. Several tutorials therefore still treated an activity embedded by another origin as authoritative. The most visible failure occurred while handling `Accept(Follow)`: calling `accept.getObject()` across the origin boundary caused Fedify to discard the inline `Follow` and dereference its ID instead. Because the tutorial had assigned an arbitrary ID without an object dispatcher, that request reached an HTML application route and failed while being parsed as ActivityStreams JSON.

Setting `crossOrigin: "trust"` would hide the immediate error, but it would also teach readers to bypass the boundary that Fedify now uses to distinguish an authoritative object from an untrusted copy. That is especially undesirable in introductory material, where examples tend to become production code. The examples should make the secure path the natural path.

The revised flow uses an activity ID as the reference carried by `Accept`, rather than embedding the whole `Follow`. For outbound requests, the tutorials record the request before sending it and use that local state to correlate the reply. *docs/tutorial/microblog.md* also gives each outgoing `Follow` a dispatcher-backed ID, so a remote server can dereference the authoritative activity at its origin, while the local inbox can verify the reply without making an HTTP request back to itself.

The matching rules preserve interoperability without weakening the origin check. An exact match against the stored activity ID is preferred. A peer-minted ID may fall back to an authenticated remote actor only when the ID belongs to a remote origin and the pending request is unambiguous; a mismatched ID on the local origin is rejected. The same reasoning covers shared-inbox delivery, where no local recipient is available in the request context, and retries reuse the stored request identity instead of minting an ID that a later `Accept` cannot match.

The simpler inbound examples now put `follow.id` in `Accept.object`, and the outbound examples no longer call `getObject()` merely to inspect a cross-origin inline `Follow`. The related guidance in *docs/manual/context.md*, *docs/manual/context-advanced.md*, *docs/manual/inbox.md*, and *docs/manual/migrate.md* was updated for the same reason. The framework examples and *.agents/skills/create-example-app-with-integration/example/src/federation.ts* follow the same pattern so newly generated integrations do not reintroduce the unsafe example.

[FEP-fe34]: https://w3id.org/fep/fe34

### DIFF
--- a/docs/tutorial/microblog.md
+++ b/docs/tutorial/microblog.md
@@ -3935,11 +3935,14 @@ The handler first tries to match `accept.objectId` against the exact activity ID
 we stored.  Some implementations mint a different ID on their own origin for
 the nested `Follow`, so it falls back to the pending request for the
 authenticated remote actor.  It never falls back for a mismatched ID on our own
-origin, and it never reads fields from the cross-origin inline `Follow`.  This
-also works when the `Accept` is delivered through the shared inbox.  After the
-match succeeds, it inserts the relationship into `follows` and marks the stored
-activity as accepted in one transaction.  The stored activity remains
-available through its object dispatcher.
+origin, and it never reads fields from the cross-origin inline `Follow`.
+Because this tutorial permits only one local account, the fallback can match at
+most one pending request, even when the `Accept` arrives through the shared
+inbox.  A multi-user application would need to narrow the match by the local
+recipient and reject ambiguous shared-inbox deliveries.  After the match
+succeeds, it inserts the relationship into `follows` and marks the stored
+activity as accepted in one transaction.  The stored activity remains available
+through its object dispatcher.
 
 ### Testing
 
--- a/docs/tutorial/microblog.md
+++ b/docs/tutorial/microblog.md
@@ -1915,7 +1915,7 @@ interface Actor { id: number; }
 federation
   .setInboxListeners("/users/{identifier}/inbox", "/inbox")
   .on(Follow, async (ctx, follow) => {
-    if (follow.objectId == null) {
+    if (follow.id == null || follow.objectId == null) {
       logger.debug("The Follow object does not have an object: {follow}", {
         follow,
       });
@@ -1928,7 +1928,7 @@ federation
       });
       return;
     }
-    const follower = await follow.getActor();
+    const follower = await follow.getActor(ctx);
     if (follower?.id == null || follower.inboxId == null) {
       logger.debug("The Follow object does not have an actor: {follow}", {
         follow,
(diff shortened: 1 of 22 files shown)


docs/tutorial/microblog.md
```diff
@@ -3744,51 +3875,71 @@ import { type Federation } from "@fedify/fedify";
...
+      .get(accept.objectId.href, accept.actorId.href);
+    const referencesLocalOrigin =
+      accept.objectId.origin === ctx.canonicalOrigin;
+    const request = exactRequest ?? (referencesLocalOrigin ? null : db
+      .prepare<unknown[], FollowRequest>(
+        `
+        SELECT follow_requests.*
+        FROM follow_requests
+        WHERE follow_requests.following_uri = ?
+          AND follow_requests.accepted = 0
```
sij411, line 3912: what if two actors A, B sends Follow to C, C sends Accept to A. and if C remints the referenced Follow id the fallback may select B's Follow pending row then B may be marked as accepted whilst A remains pending

dahlia: This would be ambiguous in a multi-user application, but this tutorial deliberately builds a single-user microblog. The `users` table enforces `CHECK (id = 1)`, and outgoing follow requests obtain `follower_id` by joining the sole local user to its actor. Together with `UNIQUE (follower_id, following_uri)`, this means the tutorial can have at most one pending request for C, even if C remints the referenced `Follow` ID.

If the application were extended to support multiple local actors, the fallback would need to be narrowed using `ctx.recipient`. For shared-inbox delivery, it should decline to guess whenever more than one candidate remains. I think the single-user tutorial should keep the simpler query, though I can add a sentence making that assumption explicit.

sij411: Understood.

dahlia: Addressed in 7057839a2970299c723c2f71351f84b29d687b6b.

docs/tutorial/threadiverse.md
```diff
@@ -2468,23 +2512,40 @@ export async function followCommunity(formData: FormData): Promise<void> {
...
     .onConflictDoNothing()
     .run();
+  const row = db
+    .select({ followActivityId: follows.followActivityId })
+    .from(follows)
+    .where(
+      and(
+        eq(follows.followerUri, followerUri.href),
+        eq(follows.followedUri, actor.id.href),
+      ),
```
2chanhaeng, line 2539: No imports for `and` and `eq` from `drizzle-orm`.

sij411 (dismissed):

2chanhaeng (changes requested):

docs/tutorial/threadiverse.md
dahlia: Addressed in 36211315: the standalone follow action now imports both Drizzle helpers it uses.

2chanhaeng (approved):

PR #980, opened 2026-08-01
Reject standalone keys with mismatched IDs

### LINKED ISSUE #963: Handle mismatched `id` values in standalone key documents
When `resolveFetchedKey` in *packages/fedify/src/sig/key.ts* fetches the document at a `keyId` URL and parses it as a standalone `CryptographicKey` or `Multikey`, it does not check whether the object's `id` matches the requested URL:

~~~~ typescript
let key: T | null = null;
if (object instanceof cls) key = object;
else if (isActor(object)) {
  // …iterates keys and matches k.id?.href === keyId
}
~~~~

The actor branch compares each candidate with `k.id?.href === keyId` before accepting it. The standalone branch skips that check. A document served at one URL can therefore declare a different `id` and still be cached under the requested URL.

This came from a security report, but we closed the report because the mismatch does not provide a way to impersonate another actor. The verification path does not use the key document's `id` to establish the actor's identity. It checks the key's owner or controller against the activity's actor.

The current behavior is nevertheless confusing: `key.id` may differ from the `keyId` used for the lookup and cache entry. For a standalone key with a present but mismatched `id`, we could reject the document or emit a warning.

The actor branch intentionally allows one related case. When the requested `keyId` has no fragment and the actor exposes exactly one key, it accepts that key without an exact `id` match:

~~~~ typescript
const keyIdUrl = new URL(keyId);
if (key == null && keyIdUrl.hash === "" && length === 1) {
  key = lastKey;
}
~~~~

A change to the standalone path should account for this existing tolerance rather than making the result depend unexpectedly on whether the response is an actor or a key document. I am not yet sure whether a mismatch should be an error. Logging it is less likely to break interoperability, but rejecting it keeps the fetched key's identity aligned with its cache key.

Acceptance criteria:

 -  A standalone key document with a present `id` that differs from the requested `keyId` is rejected or handled according to a documented and tested rule.
 -  The fragmentless single-key tolerance in the actor branch continues to work.
 -  A regression test covers a standalone key document with a mismatched `id`.

ojspp41:
## Background

`fetchKey()` accepts either an actor document containing keys or a standalone `CryptographicKey`/`Multikey` document.

The actor branch already checks candidate key IDs against the requested `keyId`. The standalone branch previously accepted any object of the requested key class without checking its `id`. As a result, a document fetched from one URL could declare a different ID and still be returned and cached under the requested URL.

Although this does not let the key impersonate another actor—the owner/controller checks remain responsible for binding a key to an actor—it leaves the fetched key identity inconsistent with its lookup and cache key.

## Changes

- Accept a standalone key when it has no `id`, preserving the existing compatibility behavior.
- Accept a standalone key with an `id` only when its URL exactly matches the requested `keyId`.
- Reject and negatively cache standalone key documents whose present `id` differs from the requested URL.
- Keep actor key resolution unchanged, including the existing fragmentless single-key tolerance.
- Add a changelog fragment for `@fedify/fedify`.

## Regression coverage

The regression test supplies standalone key documents whose IDs differ from their requested URLs and verifies that:

- both `CryptographicKey` and `Multikey` documents are rejected;
- the first lookup returns `{ key: null, cached: false }`;
- the unavailable result is cached under the requested URL; and
- a repeated lookup returns the negative cache entry with `cached: true`.

I also confirmed that the new regression test fails against the previous standalone-key behavior and passes with this change.

## Existing cache entries

This change validates newly fetched standalone key documents. It does not retroactively validate or evict mismatched key objects that may already exist in a persistent `KeyCache` from before the upgrade.

I kept existing cache-entry handling out of scope because #963 specifically concerns resolving a fetched standalone document, and cache storage and retention are provided by the application. I am happy to add validation or stale-entry handling for cached keys in this PR if the maintainers prefer that behavior.

## Testing

The following checks passed locally:

- `mise exec -- deno test --check --doc --allow-all --unstable-kv --trace-leaks packages/fedify/src/sig/key.test.ts`
- `mise run check-each fedify`
- `mise run test-each fedify` (Deno, Node.js, and Bun)
- `mise run check`

Fixes #963.

## AI usage

Implementation, regression-test drafting, code review, and local verification were assisted by Codex (GPT-5.6 Sol; model ID: `gpt-5.6-sol`). I reviewed the resulting change and exercised it locally using the Deno, Node.js, and Bun workflows listed above.

### DIFF
--- a/changes.d/fedify/mismatched-key-id.md
+++ b/changes.d/fedify/mismatched-key-id.md
@@ -1,5 +1,6 @@
  -  Standalone key documents whose `id` differs from the requested key URL are
-    now rejected instead of being cached under the wrong URL. [[#963]] by
-    ojspp41
+    now rejected instead of being cached under the wrong URL.
+    ([#963], [#980] by Junseok Oh)
 
 [#963]: https://github.com/fedify-dev/fedify/issues/963
+[#980]: https://github.com/fedify-dev/fedify/issues/980
--- a/changes.d/fedify/mismatched-key-id.md
+++ b/changes.d/fedify/mismatched-key-id.md
@@ -1,0 +1,5 @@
+ -  Standalone key documents whose `id` differs from the requested key URL are
+    now rejected instead of being cached under the wrong URL. [[#963]] by
+    ojspp41
+
+[#963]: https://github.com/fedify-dev/fedify/issues/963
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,14 @@ Version 2.0.25
 
 To be released.
 
+### @fedify/fedify
+
+ -  Standalone key documents whose `id` differs from the requested key URL are
+    now rejected instead of being cached under the wrong URL. [[#963]] by
+    ojspp41
+
+[#963]: https://github.com/fedify-dev/fedify/issues/963
+
 
 Version 2.0.24
 --------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,10 +11,11 @@ To be released.
 ### @fedify/fedify
 
  -  Standalone key documents whose `id` differs from the requested key URL are
-    now rejected instead of being cached under the wrong URL. [[#963]] by
-    ojspp41
+    now rejected instead of being cached under the wrong URL.
+    ([#963], [#980] by Junseok Oh)
 
 [#963]: https://github.com/fedify-dev/fedify/issues/963
+[#980]: https://github.com/fedify-dev/fedify/issues/980
 
 
 Version 2.0.24
--- a/packages/fedify/src/sig/key.test.ts
+++ b/packages/fedify/src/sig/key.test.ts
@@ -352,6 +352,65 @@ test("fetchKey()", async () => {
   );
 });
 
+test("fetchKey() rejects standalone keys with a mismatched id", async () => {
+  for (
+    const { keyId, standaloneKey, fetch } of [
+      {
+        keyId: "https://example.com/key",
+        standaloneKey: rsaPublicKey1,
+        fetch: (keyId: string, options: FetchKeyOptions) =>
+          fetchKey(keyId, CryptographicKey, options),
+      },
+      {
+        keyId: "https://example.com/multikey",
+        standaloneKey: ed25519Multikey,
+        fetch: (keyId: string, options: FetchKeyOptions) =>
+          fetchKey(keyId, Multikey, options),
+      },
+    ]
+  ) {
+    const cache: Record<string, CryptographicKey | Multikey | null> = {};
+    const options: FetchKeyOptions = {
+      async documentLoader(resource) {
+        if (resource === keyId) {
+          const document = await standaloneKey.toJsonLd({
+            contextLoader: mockDocumentLoader,
+          });
+          return {
+            contextUrl: null,
+            documentUrl: resource,
+            document: {
+              ...document as Record<string, unknown>,
+              id: "https://example.com/different-key",
+            },
+          };
+        }
+        return await mockDocumentLoader(resource);
+      },
+      contextLoader: mockDocumentLoader,
+      keyCache: {
+        get(keyId) {
+          return Promise.resolve(cache[keyId.href]);
+        },
+        set(keyId, key) {
+          cache[keyId.href] = key;
+          return Promise.resolve();
+        },
+      } satisfies KeyCache,
+    };
+
+    assertEquals(await fetch(keyId, options), {
+      key: null,
+      cached: false,
+    });
+    assertEquals(cache, { [keyId]: null });
+    assertEquals(await fetch(keyId, options), {
+      key: null,
+      cached: true,
+    });
+  }
+});
+
 test("fetchKey() returns null for a malformed actor publicKey", async () => {
   const actorId = "https://example.com/malformed-public-key";
   const keyId = "https://example.com/malformed-public-key#main-key";
--- a/packages/fedify/src/sig/key.ts
+++ b/packages/fedify/src/sig/key.ts
@@ -323,7 +323,10 @@ async function fetchKeyInternal<T extends CryptographicKey | Multikey>(
     }
   }
   let key: T | null = null;
-  if (object instanceof cls) key = object;
+  if (
+    object instanceof cls &&
+    (object.id == null || object.id.href === keyId)
+  ) key = object;
   else if (isActor(object)) {
     // Treat malformed remote actor keys as missing keys.
     // @ts-ignore: cls is either CryptographicKey or Multikey


dahlia (changes requested):
Before I continue reviewing the implementation, I need a few points clarified and corrected.

The PR description identifies the model as “Codex (GPT-5),” and the commit uses `Assisted-by: Codex:gpt-5`. GPT-5 is a real model identifier, but it was released nearly a year ago, and the current model family is GPT-5.6. Please check the exact model identifier shown in the Codex session or configuration. If `gpt-5` is accurate, please confirm that. If it is not, update the PR description and amend the commit so that both identify the model that was actually used. Please do not replace it with the latest model merely because it is current.

The commit also records both its author and committer as `junseok <junseok@junseoks-MacBook-Air.local>`, which GitHub cannot associate with your account. Please configure your preferred name and a GitHub-verified email address, or your GitHub-provided `noreply` address, then amend the commit and force-push it.

Fedify's [*AI_POLICY.md*](https://github.com/fedify-dev/fedify/blob/main/AI_POLICY.md) requires an AI-assisted pull request to be fully verified through human use. The PR description says that you reviewed the change and ran the Deno, Node.js, and Bun checks locally. Before proceeding, I need to confirm both your understanding of the patch and what you personally verified.

Please answer these questions in your own words, without using Codex or any other AI tool (영어가 불편하시면 한국어로 답변하셔도 됩니다).

 -  Why does accepting a standalone key with a mismatched `id` not allow that key to impersonate another actor?
 -  Why does the patch continue accepting a standalone key without an `id`?
 -  Why is the actor branch’s fragmentless single-key tolerance unaffected by this change?
 -  Why should a rejected mismatched key be negatively cached under the requested URL?
 -  How did you verify that the regression test fails without the patch?
 -  Did you personally run and review the results of each listed Deno, Node.js, and Bun check, or were those checks run autonomously by Codex?

Once these points have been addressed, I can continue reviewing the implementation.

ojspp41:
1. Why does accepting a standalone key with a mismatched id not allow that key to impersonate another actor?

  처음에는 key 조회와 actor 소유권 확인이 서로 별개라고 봤습니다. 그런데 doesActorOwnKey()도 key의 ownerId나 id를 사용하
  고 있어 사칭 가능성과 무관하다고 단정할 수 없었습니다. 그래서 요청 URL과 key의 id가 다르면 소유권을 확인하기 전에 거부
  하도록 수정했습니다.

  2. Why does the patch continue accepting a standalone key without an id?

  이번 변경에서는 id가 있고 그 값이 요청 URL과 다를 때만 거부합니다. id가 없는 key까지 막으면 기존 동작과의 호환성이 깨
  질 수 있어 이 경우는 그대로 허용했습니다.

  3. Why is the actor branch’s fragmentless single-key tolerance unaffected by this change?

  이 경우는 actor 문서 안에서 key를 찾을 때 적용되는 기존 동작입니다. 이번 검사는 standalone key를 직접 가져오는 분기에
  만 추가했기 때문에 기존 로직에는 영향을 주지 않습니다.

  4. Why should a rejected mismatched key be negatively cached under the requested URL?

  캐시는 실제로 요청한 URL을 기준으로 동작합니다. 해당 URL에서 유효한 key를 가져오지 못한 것이므로, 같은 문서를 반복해서
  요청하지 않도록 실패 결과도 요청 URL에 저장해야 합니다.

  5. How did you verify that the regression test fails without the patch?

  테스트는 그대로 두고 packages/fedify/src/sig/key.ts의 구현 조건만 패치 전 상태로 되돌려 실행했습니다. 그 결과 key:
  null이 아니라 mismatched CryptographicKey가 반환되면서 테스트가 실패했습니다. 이후 패치를 복원하자 같은 테스트가 통과
  했고, 전체 7개 테스트도 모두 통과했습니다.

  6. Did you personally run and review the results of each listed Deno, Node.js, and Bun check, or were those checks run
     autonomously by Codex?

  focused Deno 테스트, 패치 제거·복원 검증, Deno 전체 7개 테스트, check-each
  fedify는 제가 직접 실행했습니다. 마지막으로 test-each fedify의 Deno·Node.js·Bun 검사와 mise run check도 직접 실행해 모
  두 종료 코드 0으로 끝난 것을 확인했습니다.

image

changes.d/fedify/mismatched-key-id.md
```diff
@@ -0,0 +1,5 @@
+ -  Standalone key documents whose `id` differs from the requested key URL are
+    now rejected instead of being cached under the wrong URL. [[#963]] by
+    ojspp41
```
dahlia: 1. A closing parenthesis is missing.
2. You need to put `#980`, the number of this pull request as well.
3. We prefer using real names rather than pseudonyms here. Unless you have any strong preference for a pseudonym, could you please use your real name instead?

dahlia (changes requested):
Thank you for the detailed response and for updating the commit. The author and committer identity is now correctly associated with your GitHub account. Your explanation and screenshot of the regression test, together with your explicit confirmation that you personally ran and reviewed the listed checks, answer my questions about human verification.

Two points remain.

First, the commit trailer now says `Assisted-by: Codex:gpt-5.6-sol`, but the PR description still says “Codex (GPT-5).” Please confirm how you determined that the exact model used was `gpt-5.6-sol`, rather than inferring it from the current model family, and update the PR description to match.

Second, your answer to the first question conflicts with both the PR description and the reasoning in #963. You wrote that, because `doesActorOwnKey()` uses the key’s `ownerId` or `id`, you could not conclude that the mismatch was unrelated to impersonation.

In the HTTP signature path, `doesActorOwnKey()` checks ownership in one of two ways. If `ownerId` is present, it compares that value with the activity’s actor. If `ownerId` is absent, it fetches the actor and checks whether that actor lists the key’s `id`. In either case, the mismatch between the requested `keyId` URL and the returned object’s `id` is not what establishes the actor–key relationship.

My understanding is therefore that this patch keeps the fetched object’s identity consistent with its lookup and cache key, but does not close an actor impersonation path. That is also what the current PR description says.

If you agree, please clarify your first answer accordingly. If you still believe the mismatch may enable impersonation, please describe a concrete example, including the requested `keyId`, the returned key’s `id` and `ownerId`, the activity’s actor, and the path through `doesActorOwnKey()` that would accept it. That would materially change the security assessment and scope of this PR.

Please answer this remaining technical point in your own words, without using Codex or any other AI tool.

changes.d/fedify/mismatched-key-id.md
ojspp41: 이전 답변에서 보안 영향을 잘못 이해했습니다.
ownerId가 있으면 activity actor와 직접 비교하고, 없으면 actor 문서에 반환된 key 객체의 id가 포함돼 있는지 확인합니다. 그래서 keyId와  id가 다르다는 것만으로 actor 소유권 검사를 통과하는 건 아닙니다.
모델명은 현재 모델을 보고 추정한 게 아니라, 당시 로컬 Codex 세션의 model을 확인해서 gpt-5.6-sol인 것을 확인했습니다. PR 설명도 이에 맞게 수정했습니다.

changes.d/fedify/mismatched-key-id.md
```diff
@@ -0,0 +1,6 @@
+ -  Standalone key documents whose `id` differs from the requested key URL are
+    now rejected instead of being cached under the wrong URL.
+    ([#963], [#980] by Junseok Oh)
+
+[#963]: https://github.com/fedify-dev/fedify/issues/963
+[#980]: https://github.com/fedify-dev/fedify/issues/980
```
dahlia, line 6: You don't have to declare these link definitions; instead you should add resolved URLs in the front matter; use the `sacho resolve-links` command.

changes.d/fedify/mismatched-key-id.md
```diff
@@ -0,0 +1,6 @@
+ -  Standalone key documents whose `id` differs from the requested key URL are
+    now rejected instead of being cached under the wrong URL.
+    ([#963], [#980] by Junseok Oh)
```
dahlia, line 3: Use square brackets instead of parentheses:


```suggestion
    [[#963], [#980] by Junseok Oh]
```

dahlia (changes requested):
Could you rebase your commits on the latest `2.0-maintenance`?

dahlia (approved):
Okay, I'm fixing the changelog formatting by myself.

PR #985, opened 2026-08-07
Add an async context data test for @fedify/express

### LINKED ISSUE #855: Add an async context data test for `@fedify/express`
*@fedify/express* accepts both synchronous and asynchronous context data factories. The async path should be covered so applications can safely load context from sessions or databases.

Add a test in *packages/express/src/index.test.ts* that verifies `integrateFederation()` waits for an async `contextDataFactory` and passes the resolved value to `federation.fetch()`.

Suggested check: `mise run check-each express`.

Jae-Hyuk-Jang:
## Summary

`integrateFederation()`'s `contextDataFactory` may return a plain value or a `Promise`, but only the synchronous path had test coverage. 
This adds a regression test for the async path, verifying the middleware waits for the factory to resolve and passes the resolved value to `federation.fetch()`.

Also adds the missing `test` script (*package.json*) and `test` task (*deno.json*) for `@fedify/express` so the suite actually runs under `mise run test:node` / `test-each`.

Fixes #855

## Test plan

- [x] `mise run check-each express`
- [x] `deno test --allow-all` (packages/express)
- [x] `node --experimental-transform-types --test` (packages/express)

## AI disclosure

This change was written with assistance from Claude Code (`claude-sonnet-5`), reviewed and verified by me.

### DIFF
--- a/packages/express/package.json
+++ b/packages/express/package.json
@@ -58,6 +58,7 @@
     "build:self": "tsdown",
     "build": "pnpm --filter @fedify/express... run build:self",
     "prepack": "pnpm build",
-    "prepublish": "pnpm build"
+    "prepublish": "pnpm build",
+    "test": "node --experimental-transform-types --test"
   }
 }
--- a/packages/express/package.json
+++ b/packages/express/package.json
@@ -59,6 +59,7 @@
     "build": "pnpm --filter @fedify/express... run build:self",
     "prepack": "pnpm build",
     "prepublish": "pnpm build",
-    "test": "node --experimental-transform-types --test"
+    "test": "node --experimental-transform-types --test",
+    "test:bun": "bun test src/"
   }
 }
--- a/packages/express/deno.json
+++ b/packages/express/deno.json
@@ -19,6 +19,7 @@
     ]
   },
   "tasks": {
-    "check": "deno fmt --check && deno lint && deno check *.ts"
+    "check": "deno fmt --check && deno lint && deno check *.ts",
+    "test": "deno test --allow-all"
   }
 }
--- a/packages/express/src/index.test.ts
+++ b/packages/express/src/index.test.ts
@@ -36,21 +36,12 @@ function createMockResponse(): {
     setHeader() {
       return response;
     },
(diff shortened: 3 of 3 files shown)


sij411 (commented):
Since `integrateFederation()` only assigns send, json, and removeHeader and never calls their existing implementations, could these methods be omitted from the initial response mock?

packages/express/package.json
```diff
@@ -58,6 +58,7 @@
     "build:self": "tsdown",
     "build": "pnpm --filter @fedify/express... run build:self",
     "prepack": "pnpm build",
-    "prepublish": "pnpm build"
+    "prepublish": "pnpm build",
+    "test": "node --experimental-transform-types --test"
```
2chanhaeng: Please add `test:bun` script.

2chanhaeng (changes requested):
Thanks for your contribution! Please check my comment.

packages/express/package.json
Jae-Hyuk-Jang: Added, thanks!

Jae-Hyuk-Jang:
> Since `integrateFederation()` only assigns send, json, and removeHeader and never calls their existing implementations, could these methods be omitted from the initial response mock?

Good catch, removed them — thanks!

packages/express/package.json
```diff
@@ -58,6 +58,8 @@
     "build:self": "tsdown",
     "build": "pnpm --filter @fedify/express... run build:self",
     "prepack": "pnpm build",
-    "prepublish": "pnpm build"
+    "prepublish": "pnpm build",
+    "test": "node --experimental-transform-types --test",
```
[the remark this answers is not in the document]
2chanhaeng: That is the version required for actual use, and this script is the version used for development, which is pinned to the latest version of v22 according to *mise.toml*. The latest available version of `node@22` in `mise` is 22.23.2, which can use the `--experimental-transform-types` option. Do not mention this again.

2chanhaeng (approved):

dahlia (approved):

PR #990, opened 2026-08-09
Add a smoke test to generated `fedify init` apps

### LINKED ISSUE #898: Add a smoke-test task to generated `fedify init` apps
## Why

`fedify init` creates runnable apps, but generated projects do not have a standard local smoke-test command. A small generated task that starts the app and verifies a local actor lookup would help new users and would give contributors a clearer way to validate scaffold changes.

## Scope

Add a smoke-test task to generated projects from `@fedify/init`. The task should be practical across supported package managers and should check the generated actor endpoint with `fedify lookup` or an equivalent local request.

This will likely touch task generation in *packages/init/src/action/*, framework descriptions in *packages/init/src/webframeworks/*, and tests under *packages/init/src/test/*. Keep the first version focused on generated apps that can run locally without external services.

## Suggested checks

Run `mise run check-each init`. Then run focused `mise test:init` cases for at least one Deno template and one Node.js or Bun template.

Palcimer:
Summary
-------
  `fedify init` creates runnable apps, but a generated project has no standard way to check that it actually serves an actor object.  Confirming it means starting the dev server by hand and looking an actor up separately, which is awkward for new users and gives contributors no quick way to validate scaffold changes.
  This adds a `test` task to every scaffolded project.  The task runs a generated *scripts/smokeTest.ts* that starts the app, reads the port the server actually bound from its output, waits for it to answer, and resolves the local actor with `lookupObject()`.  On success, it prints the resolved actor. On failure, it prints the server's stdout and stderr output and exits the process with a failure code. 

Assisted-by: Claude Code:claude-sonnet-5

Related issue
-------------

 -  closes #898 

Changes
-------
 -  Added a required `testFile` property to the `WebFrameworkInitializer` interface, holding the path the smoke-test script is written to.
 -  Added a template for the generated smoke-test script under *templates/defaults*, shared by every framework.
 -  Added `loadTest()`, which reads that template and bakes in the dev command for the chosen package manager.
 -  Added `tsx` as a dev dependency to generated apps whose package manager runs on Node.js, so that they can launch the smoke test.
 -  Added a `test` task that starts the smoke test to the generated apps.
 -  Updated the tests for `patchFiles()` to verify that `fedify init` writes the smoke-test script.

Benefits
--------

 - Users can quickly verify that a newly generated app is running correctly and exposing an actor as expected. 

Checklist
---------

 -  [] ~~Did you write some relevant docs about this change (if it's a new feature)?~~
 -  [] ~~Did you write a regression test to reproduce the bug (if it's a bug fix)?~~

### DIFF
--- a/changes.d/init/smoke-test.md
+++ b/changes.d/init/smoke-test.md
@@ -1,0 +1,4 @@
+ -  Added a `test` task to projects scaffolded by `fedify init`.  It starts
+    the app, waits for it to become ready, and checks that it resolves a local
+    actor, giving projects a standard smoke test to run right after scaffolding
+    and whenever the app changes afterwards.  [[#898]]
--- a/changes.d/init/smoke-test.md
+++ b/changes.d/init/smoke-test.md
@@ -1,4 +1,8 @@
+---
+links:
+  '#990': https://github.com/fedify-dev/fedify/pull/990
+---
  -  Added a `test` task to projects scaffolded by `fedify init`.  It starts
     the app, waits for it to become ready, and checks that it resolves a local
     actor, giving projects a standard smoke test to run right after scaffolding
-    and whenever the app changes afterwards.  [[#898]]
+    and whenever the app changes afterwards.  [[#898], [#990] by Jang Hanarae]
--- a/packages/init/src/action/templates.ts
+++ b/packages/init/src/action/templates.ts
@@ -1,6 +1,6 @@
 import { concat, entries, join, map, pipe, when } from "@fxts/core";
 import { toMerged } from "es-toolkit";
-import { readTemplate } from "../lib.ts";
+import { getDevCommand, readTemplate } from "../lib.ts";
 import type { InitCommandData, PackageManager } from "../types.ts";
 import { replace } from "../utils.ts";
 import { needsDenoDotenv } from "./utils.ts";
@@ -55,6 +55,28 @@ export const loadLogging = async (
   pipe(
     await readTemplate(initializer.loggingTemplate ?? "defaults/logging.ts"),
     replace(/\/\* project name \*\//, JSON.stringify(projectName)),
+  );
+
+/**
+ * Loads the smoke-test script content for the initializer.
+ *
+ * Every framework shares the same *defaults/smokeTest.ts* template, so unlike
+ * {@link loadLogging} there is no per-framework template override.  The
+ * template spawns the project's own dev server, so it needs the dev command
+ * for the chosen package manager baked in at generation time.
+ *
+ * @param param0 - {@link InitCommandData} containing `packageManager`
+ * @returns The complete smoke-test script content as a string
+ */
+export const loadTest = async (
+  { packageManager }: InitCommandData,
+) =>
+  pipe(
+    await readTemplate("defaults/smokeTest.ts"),
+    replace(
+      /\/\* dev command \*\//,
+      JSON.stringify(getDevCommand(packageManager).split(" ")),
+    ),
   );
 
 /**
--- a/packages/init/src/action/templates.ts
+++ b/packages/init/src/action/templates.ts
@@ -60,7 +60,7 @@ export const loadLogging = async (
 /**
  * Loads the smoke-test script content for the initializer.
  *
- * Every framework shares the same *defaults/smokeTest.ts* template, so unlike
+ * Every framework shares the same *defaults/smoke.test.ts* template, so unlike
  * {@link loadLogging} there is no per-framework template override.  The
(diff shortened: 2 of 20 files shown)


changes.d/init/smoke-test.md
```diff
@@ -0,0 +1,4 @@
+ -  Added a `test` task to projects scaffolded by `fedify init`.  It starts
+    the app, waits for it to become ready, and checks that it resolves a local
+    actor, giving projects a standard smoke test to run right after scaffolding
+    and whenever the app changes afterwards.  [[#898]]
```
dahlia: Add this pull request number and also your name here!

dahlia (changes requested):
@2chanhaeng Could you take a look into this? Thanks!

packages/init/src/templates/defaults/smoke.test.ts.tpl
2chanhaeng, line 1: It seems like it would be better to either match the file name to kebab-case like the other file names, or name it `smoke.test.ts.tpl` like the other test files. I personally prefer the latter. Please also modify the actually generated files to follow the same naming convention.

packages/init/src/templates/defaults/smoke.test.ts.tpl
```diff
@@ -0,0 +1,141 @@
...
+const DEV_COMMAND: string[] = /* dev command */;
+const HANDLE = "john";
+const STARTUP_TIMEOUT = 15_000;
+
+async function main(): Promise<void> {
+  const [command, ...args] = DEV_COMMAND;
+  const server = spawn(command, args, {
+    stdio: ["ignore", "pipe", "pipe"],
+    detached: true,
+  });
```
2chanhaeng, line 18: I think this would be failed in Windows. Check `shell` option in [`spawn` of `node:child_process`](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options).

packages/init/src/action/templates.ts
```diff
@@ -57,6 +57,28 @@ export const loadLogging = async (
...
+ * @returns The complete smoke-test script content as a string
+ */
+export const loadTest = async (
+  { packageManager }: InitCommandData,
+) =>
+  pipe(
+    await readTemplate("defaults/smokeTest.ts"),
+    replace(
+      /\/\* dev command \*\//,
+      JSON.stringify(getDevCommand(packageManager).split(" ")),
```
2chanhaeng: If converted as like this, the formatter will show that a formatting issue occurred, causing unnecessary confusion for the user. Add a space (` `) after the comma (`,`).

packages/init/src/templates/defaults/smokeTest.ts.tpl
```diff
@@ -0,0 +1,141 @@
...
+  }
+  if (!isActor(object)) {
+    throw new Error(`Expected an actor at ${url}, but got a non-actor object.`);
+  }
+  return object;
+}
+
+function stopServer(server: ReturnType<typeof spawn>): void {
+  try {
+    if (server.pid != null) process.kill(-server.pid, "SIGKILL");
```
2chanhaeng: Process tree termination handling for Windows is required. Because Windows does not support terminating a process group using a negative PID.

packages/init/src/templates/defaults/smokeTest.ts.tpl
```diff
@@ -0,0 +1,141 @@
...
+      const text = chunk.toString("utf8");
+      for (const pattern of portPatterns) {
+        const match = text.match(pattern);
+        if (match && match[1]) {
+          const port = Number.parseInt(match[1], 10);
+          clearTimeout(timeout);
+          resolve(port);
+          return;
+        }
+      }
```
2chanhaeng: The code searches a port within individual chunks. This might cause the banner to be split, potentially preventing the port from being found even if the server is operating normally. Please refer to `src/test/server.ts` and modify it.

2chanhaeng (changes requested):
Thanks for your hard work! Before merging this PR, it seems like this PR need to consider the Windows environment a bit more. I've left comments on the issues that stood out, but there might be other compatibility problems as well. If you have a Windows machine, it would be great to run it in an actual Windows environment. If you live in Korea, another option is to try running it at a PC방. Of course, if you don't have the means to do so, I completely understand and.

dahlia (approved):

2chanhaeng (approved):
Great work! Finally, please rebase from the main branch.

dahlia:
I think you don't have to rebase.  I'll merge this now!

Palcimer:
Thanks for the review! Glad it got merged.

PR #993, opened 2026-08-10
Add delegation tests for `@fedify/elysia`

### LINKED ISSUE #854: Test not-found delegation in `@fedify/elysia`
*@fedify/elysia* relies on `onNotFound` to let normal Elysia routes handle non-federation requests. That behavior is part of the integration contract, but it is not covered by a small package-level test.

Add a test in *packages/elysia/src/index.test.ts* that verifies `fedify()` does not return a Fedify response when `onNotFound` is used.

Suggested check: `mise run check-each elysia`.

dktsudgg:
Closes #854

## Summary

Adds two tests to `packages/elysia/src/index.test.ts` that verify the `fedify()` plugin does not return a Fedify response when federation delegates the request back to the app.

### 1) Not-found delegation test

A federation with no dispatcher registered reports not-found via `onNotFound`,
and the status, headers, and body all come from the app's own Elysia route.

### 2) Not-acceptable delegation test

A request matches the actor route while `Accept: text/html` rules out JSON-LD,
so federation declines via `onNotAcceptable` and the app's own Elysia route serves the response.

## Test Plan

Tested with the commands below, and all of them passed.

- `mise run check-each elysia`
- `mise run test-each elysia`
- `deno test --allow-all ./packages/elysia`
- `bun test ./packages/elysia`
- `mise run test`

AI assistance disclosure: Claude Code (claude-fable-5) helped write the tests, verify them across runtimes, and draft the commit messages and this description.

### DIFF
--- a/packages/elysia/src/index.test.ts
+++ b/packages/elysia/src/index.test.ts
@@ -77,4 +77,37 @@ describe("[elysia] fedify() plugin", () => {
       "body must come from Elysia",
     );
   });
+
+  test("fedify() falls through to Elysia routes when federation declines the request via onNotAcceptable", async () => {
+    const elysia = new Elysia().get(
+      "/users/alice",
+      ({ set, status }) => {
+        set.headers["X-Custom-Header"] = "custom-value";
+        return status(200, "Hello Alice");
+      },
+    );
+    // Register the actor route so the request matches, but ask for text/html
+    // so federation declines via onNotAcceptable instead of handling it.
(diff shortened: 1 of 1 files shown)


dktsudgg:
`2) Not-acceptable delegation` test goes slightly beyond the literal scope of issue #854,
but I added it because it is the sibling path of the same delegation contract.

dahlia (approved):
Thanks for your contribution!

Issue #885, opened 2026-06-30
List the SolidStart example in the examples index


dahlia:
*examples/README.md* should help users find framework examples from one place. The SolidStart example should be listed there instead of relying on directory discovery.

Update *examples/README.md* to include *examples/solidstart* with a short description of what it demonstrates.

Suggested check: `hongdown --check examples/README.md`.

rewrite0w0:
Could you please assign this issue to me?

dahlia:
@rewrite0w0 Sorry for my late response. Assigned. Thanks!

rewrite0w0:
@dahlia 
I’m sorry for the delayed response as well. Safal is moving forward with the work, I think it is right for him to proceed.
I’ll find another contribution topic and contribute :)

PR #1005, opened 2026-08-20
Add a README for the SolidStart example

### LINKED ISSUE #883: Add a README for the SolidStart example
*examples/solidstart* has a full example app, but no README. A short README would make the example discoverable without reading the source tree first.

Add *examples/solidstart/README.md* with a short description, setup steps, run command, and the actor URL that can be looked up.

Suggested check: `hongdown --check examples/solidstart/README.md`.

Safal-Shrestha-SS:
Closes #883.

*examples/solidstart* had a full app but no README, so the only way to find out
which actor it serves, or on which port, was to read the source tree.

The README covers the install and run commands, the demo actor URL, the
WebFinger and NodeInfo endpoints, and the `fedify tunnel` steps for following
the actor from another server.  I kept it close in shape to
*examples/nuxt/README.md* rather than the longer *sveltekit-sample* one, since
the issue asked for something short.

## A second commit, because the tunnel steps did not work

Writing the tunnel section and then actually running it turned up a bug in the
example itself.  Vite rejects any request whose Host header it does not
recognise, so every request through `fedify tunnel` came back as:

~~~~
Blocked request. This host ("...serveousercontent.com") is not allowed.
To allow this host, add "..." to `server.allowedHosts` in vite.config.js.
~~~~

That makes step 2 of the section impossible and the follow step unreachable.
The sibling examples had already hit this — nuxt, sveltekit-sample, astro and
netlify-astro all set `allowedHosts` in their configs — and *app.config.ts*
here was simply missing it, so the second commit adds the same three lines.

I would rather widen the diff slightly than land a README whose instructions
403.  Happy to split it into its own PR against a separate issue if you would
prefer to keep this one documentation-only.

## Checks

- `hongdown --check examples/solidstart/README.md` — passes
- `deno fmt --check` on *app.config.ts* — passes
- Ran `pnpm dev` and confirmed every URL in the README responds:
  - `GET /` → `302` to `/users/demo`
  - `GET /users/demo` with `Accept: application/activity+json` → `200`,
    `content-type: application/activity+json`
  - `GET /.well-known/webfinger?resource=acct:demo@localhost:3000` → `200`
  - `GET /nodeinfo/2.1` → `200`
  - port 3000 confirmed from the running server, not assumed
- Ran the tunnel section end to end.  With the fix, the tunnelled actor
  resolves over HTTPS with its `id` and `inbox` rewritten to the public host,
  WebFinger answers for `acct:demo@<tunnel host>`, and
  `fedify lookup @demo@<tunnel host>` completes discovery and reads back the
  keys and shared inbox — the sequence a remote server runs before it delivers
  a `Follow`.
- Ran the follow itself from a throwaway ActivityPub Academy account.  The
  handle resolved in Mastodon's search, the `Follow` reached the inbox, the
  example answered with an `Accept`, and the follower count on the profile page
  went from 0 to 1 with the remote actor recorded in
  `/users/demo/followers` — so every step of that section is confirmed working,
  not just the endpoints leading up to it.

I did not touch *examples/README.md*, since #885 covers the index entry.

## AI disclosure

Per *AI_POLICY.md*: this was written with Claude Code (claude-opus-5).  It
drafted the README, made the *app.config.ts* fix, and ran the checks listed
above.  Both commits carry the `Assisted-by: Claude Code:claude-opus-5`
trailer.

### DIFF
--- a/examples/solidstart/README.md
+++ b/examples/solidstart/README.md
@@ -1,0 +1,72 @@
+<!-- deno-fmt-ignore-file -->
+
+Fedify–SolidStart integration example application
+=================================================
+
+An example of building a federated server application using [Fedify] with
+[SolidStart].  It serves a single demo actor over ActivityPub, so it can be
+looked up and followed from other federated platforms such as [Mastodon] or
+[Misskey].
+
+Federation is wired up in *src/lib/federation.ts* and mounted through
+`fedifyMiddleware()` in *src/middleware/index.ts*.  Keys and follower
+relationships are kept in memory, so they are discarded whenever the server
+restarts.
+
+[Fedify]: https://fedify.dev
+[SolidStart]: https://start.solidjs.com/
+[Mastodon]: https://mastodon.social/
+[Misskey]: https://misskey.io/
+
+
+Running the example
+-------------------
+
+~~~~ sh
+pnpm install
+pnpm dev
+~~~~
+
+The server listens on <http://localhost:3000/>, which redirects to the demo
+actor's profile page.
+
+
+Looking up the actor
+--------------------
+
+The example exposes one actor, `demo`, at
+<http://localhost:3000/users/demo>:
+
+~~~~ sh
+curl -H "Accept: application/activity+json" http://localhost:3000/users/demo
+~~~~
+
+WebFinger and NodeInfo are served too:
+
+~~~~ sh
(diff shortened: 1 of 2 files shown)


Safal-Shrestha-SS:
@coderabbitai review

dahlia (approved):

PR #1006, opened 2026-08-20
Add AdonisJS integration package

### LINKED ISSUE #139: AdonisJS Integration
Helloooo
I think I'm not the only one who want Fedify in AdonisJS ! It should be possible because AdonisJS supports custom middleware and, since the V6, the ESM packages ! 
But I don't have any ideas to solve the problem : how to implement ?
A person on the discord of AdoniJS tells me : 
`So the middleware will return a response, and works with the raw request, which I think is possible to get from Adonis' Request class`
But I tryed and, obviously it doesn't work. I think integreate Fedify in AdonisJS should be an advance because Adonis has a big community instead of the others web frameworks x)
Anyone have an idea to integreate ?

scammo:
The new @fedify/adonisjs package integrates Fedify with the AdonisJS framework (based on Issue #139).  It provides server middleware to mount a Federation within an AdonisJS application, a service provider to manage the federation's lifecycle, a `node ace configure` hook to scaffold config/fedify.ts and the federation preload files, and a `ctx.federation` request context created on demand.

The package targets Node.js 24 or later, matching the engine floor that @adonisjs/core itself declares. It is published to npm only. AdonisJS does not support Deno or Bun.  

The architecture follows the prototype that @ThisIsMissEm shared in the issue thread: https://github.com/fedify-dev/fedify/issues/139.

I performed a personal round-trip test of the application /examples/adonisjs, which worked successfully :tada: 

Assisted by Claude. Code: claude-fable-5.


packages/adonisjs/package.json
```diff
@@ -0,0 +1,126 @@
...
+    "@fedify/fedify": "workspace:^"
+  },
+  "peerDependenciesMeta": {
+    "@adonisjs/assembler": {
+      "optional": true
+    }
+  },
+  "dependencies": {
+    "@logtape/logtape": "catalog:"
+  },
```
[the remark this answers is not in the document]
scammo: Check again. This is described in the skill and consistent with other packages.

packages/adonisjs/src/fedify_middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+    const Middleware = fedifyMiddleware<ContextData>(
+      federation,
+      config.contextDataFactory,
+      {
+        ignoreRoutePrefixes: config.ignoreRoutePrefixes,
+      },
+    );
+
+    return new Middleware().handle(ctx, next);
+  }
```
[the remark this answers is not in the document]
scammo: Plese check again
The comment on handle() records this as a deliberate choice rather than an oversight. What is allocated per request is one closure and one class object, both short-lived; the two container lookups around them cost more than the allocation does.

A WeakMap keyed on the federation instance alone would also be keying on the wrong thing: the produced middleware closes over config.contextDataFactory and config.ignoreRoutePrefixes too, so the cache would only be correct as long as one federation instance never appears with two configurations. That is a latent correctness hazard in exchange for an unmeasured micro-optimisation, so I would rather leave it as it is.

packages/adonisjs/src/logging.ts
```diff
@@ -0,0 +1,112 @@
...
+          logger.info(formatter(record));
+          return;
+        }
+        // The formatter still emits an empty `[]` where the level would go.
+        write(formatter(record).replace(/^\[\]\s+/, "").trimEnd());
+      },
+    },
+    filters: {},
+    loggers: categories.map((category) => ({ category, sinks: ["adonisjs"] })),
+  });
```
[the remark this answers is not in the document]
scammo: Please check again 
There is no lowestLevel assignment in src/logging.ts. configure() is called with sinks, filters and loggers only, and logger.level is never read anywhere in the package, so there is nothing to normalise here — the finding appears to describe code that is not in this branch.

The idea behind it is worth considering separately: mapping AdonisJS's level onto LogTape's lowestLevel would stop the sink formatting records that Pino goes on to drop. That is an enhancement rather than a fix, and it needs the "silent" → null and "warn" → "warning" mapping the comment sketches. Happy to open an issue for it.

packages/adonisjs/src/middleware.ts
```diff
@@ -0,0 +1,426 @@
...
+    enumerable: false,
+    get() {
+      if (!created) {
+        cached = create();
+        created = true;
+      }
+      return cached;
+    },
+  });
+}
```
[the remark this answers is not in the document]
scammo: please check again
ctx.federation is owned by the middleware and documented as the Fedify context for the current request; assigning to it is not a supported operation. A no-op setter would swallow that mistake silently, and a memoising one would let any downstream middleware replace the request context for the remainder of the request — a considerably worse failure mode than the TypeError, which at least names the property that was written to.

The property is also the one place where the "either Fedify saw this request or accessing the context is a bug" invariant is enforced, via E_FEDIFY_CONTEXT_UNAVAILABLE; a setter would give callers a way around it. I would rather keep the assignment failing loudly

changes.d/adonisjs/adonisjs-integration.md
```diff
@@ -0,0 +1,11 @@
...
+links:
+  '#139': https://github.com/fedify-dev/fedify/issues/139
+---
+ -  Added the new *@fedify/adonisjs* package, an integration for the AdonisJS
+    framework.  It provides a server middleware that mounts a `Federation`
+    inside an AdonisJS application, a service provider that owns the
+    federation's lifecycle, a `node ace configure` hook that scaffolds
+    *config/fedify.ts* and the federation preload files, and a `ctx.federation`
+    request context.  The package targets Node.js and is published to npm only.
+    [[#139]]
```
dahlia: Please add the number of this pull request, and also your name for the credit!

packages/adonisjs/src/middleware.ts
```diff
@@ -0,0 +1,455 @@
...
+        // `ctx.federation` is then a programming error, but it is declared
+        // non-optional, so it still has to exist and say so.
+        defineFederationContext(ctx, () => {
+          throw new E_FEDIFY_CONTEXT_UNAVAILABLE(path);
+        });
+        return next();
+      }
+
+      const request = toFetchRequest(ctx);
+      const contextData = await contextDataFactory(ctx);
```
dahlia: `contextDataFactory` runs in the server middleware stack before AdonisJS router middleware. The documented example reads `ctx.auth.user`, but `ctx.auth` is still undefined here, so every non-ignored request can fail before `next()` reaches `initialize_auth_middleware`. Please make the required request state available before invoking the factory, or redesign and document the factory so it cannot depend on auth or session state.

packages/adonisjs/src/middleware.ts
```diff
@@ -0,0 +1,455 @@
...
+         * serve an HTML profile page to browsers and an ActivityPub actor
+         * document to servers.
+         *
+         * `Vary: Accept` is set before delegating so that whatever AdonisJS
+         * returns is cached per `Accept` header rather than shared across
+         * content types.
+         */
+        onNotAcceptable: async () => {
+          outcome.delegated = "notAcceptable";
+          ctx.response.header("Vary", "Accept");
```
dahlia: Setting `Vary: Accept` before `await next()` is not sufficient. A matched route or downstream middleware can replace it with its own value, such as `Vary: Origin`, and the delegated response then returns without the merge performed only for the generated `406 Not Acceptable` path. This can let a cache reuse an HTML response for an ActivityPub request. Please merge `Accept` into the final delegated response after `next()` for every `notAcceptable` outcome.

packages/adonisjs/README.md
```diff
@@ -0,0 +1,332 @@
...
+    `@fedify/postgres`, `@fedify/redis` and `@fedify/mysql` all provide both.
+    Setting `queue` also requires `queueContextData`, which says who drains it:
+    a callback has the web process do it (started in the `ready` hook, stopped
+    on shutdown, and never started in an Ace command or a test run), while
+    `false` leaves the queue to a separate worker.
+ -  **Configure `trustProxy`** in *config/app.ts* if AdonisJS needs to see the
+    original protocol and host.
+
+A complete, runnable application using all of this lives in the
+[*examples/adonisjs*](../../examples/adonisjs) directory.
```
dahlia: *examples/adonisjs* does not exist in this PR or in the repository, so this link is broken and the promised runnable application cannot be checked. Please add and validate the example with `mise run test:examples -- adonisjs`, or remove this claim and link.

dahlia (dismissed):
Great work! Thanks for your contribution.

.github/CODEOWNERS
```diff
@@ -87,6 +87,7 @@
 # Framework and platform integrations.  Hong Minhee maintains all of
 # them; ChanHaeng Lee co-maintains the ones he contributed to, and
 # leads the Nuxt integration.
+/packages/adonisjs/                                       @scammo @dahlia
```
2chanhaeng: As I known, people with write permissions for this repository can be code owners. Check out the [GitHub Docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners).

packages/adonisjs/src/logging.ts
```diff
@@ -0,0 +1,130 @@
...
+    );
+    return;
+  }
+
+  const formatter: TextFormatter = getTextFormatter({
+    timestamp: "disabled",
+    level: () => "",
+  });
+
+  await configure({
```
2chanhaeng, line 139: `contextLocalStorage: new AsyncLocalStorage()` is necessary.

packages/adonisjs/src/fedify_middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+   * be built safely either.  It does not matter: both bindings are memoised by
+   * the service provider, so resolving them is a map lookup, and
+   * `fedifyMiddleware` only closes over its three arguments.
+   *
+   * The request-scoped resolver is used rather than the application container
+   * so that resolution participates in whatever request-scoped bindings the
+   * application has registered.  Both reach the same singletons.
+   */
+  async handle(ctx: HttpContext, next: NextFn): Promise<void> {
+    const federation = await ctx.containerResolver.make("fedify");
```
2chanhaeng: I think this build federation at every request?

packages/adonisjs/src/middleware.ts
```diff
@@ -0,0 +1,487 @@
...
+ * `completeUrl()` would `decodeURI()` the path, but Fedify's URI templates
+ * percent-encode sub-delims, so `/actors/alice%21` has to reach it encoded.
+ * The raw target also keeps the bytes a peer signed, and resolving it against
+ * an origin normalises the absolute-form target RFC 9112 permits
+ * (`GET http://host/inbox HTTP/1.1`).
+ */
+function toRequestUrl(request: HttpContext["request"]): URL {
+  // Fedify rewrites the origin to `FederationOptions.origin`, so this only has
+  // to be well-formed.
+  const authority = request.host() ?? request.hostname() ?? "localhost";
```
2chanhaeng: [`X-Forwarded-Host` can contain more than one (comma-separated) value](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#x-headers) and [AdonisJS's `request.host()` doesn't split it](https://github.com/adonisjs/http-server/blob/330001317644578ec758a9a65165422bffc9a183/src/request.ts#L566). Therefore, the `new URL(...)` in the return statement throws an Error, and the server will response 500 status. Checkout [how Express split it](https://github.com/expressjs/express/blob/1faf228935aa0a13111f92c28ee795be64ce3f0f/lib/request.js#L436).

packages/adonisjs/configure.ts
```diff
@@ -0,0 +1,110 @@
...
+   * The middleware must live in the **server** stack.
+   *
+   * Fedify answers paths the AdonisJS router does not know about
+   * (`/.well-known/webfinger`, `/.well-known/nodeinfo`, actor and inbox
+   * endpoints), so it has to run before route matching.  The server stack also
+   * runs before `bodyparser_middleware` and before `@adonisjs/shield`, which
+   * matters twice over: HTTP Signature verification needs the raw, unparsed
+   * request body, and inbox `POST`s must not be rejected by CSRF protection.
+   */
+  await codemods.registerMiddleware("server", [{
```
2chanhaeng, line 129: I think this needs `position: "before"` because the federation middleware has to run before route matching.

packages/adonisjs/src/define_config.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * The `defineConfig` helper used by `config/fedify.ts`.
+ *
+ * @module
+ */
+import { configProvider } from "@adonisjs/core";
+import type { ConfigProvider } from "@adonisjs/core/types";
+import { InvalidArgumentsException } from "@adonisjs/core/exceptions";
+import { MemoryKvStore } from "@fedify/fedify";
```
2chanhaeng: I think this import should be lazily and move to the between of line 51 and 52 like:
```ts
  if (kv === undefined) {
      const { MemoryKvStore } = await import("@fedify/fedify");
      kv = new MemoryKvStore();
```

Because this single line causes `ERR_MODULE_NOT_FOUND` and moving it makes the entry loadable without the peer.

examples/adonisjs/README.md
```diff
@@ -0,0 +1,73 @@
...
+    ~~~~
+
+2.  Install dependencies and start the server:
+
+    ~~~~ sh
+    pnpm install
+    pnpm start
+    ~~~~
+
+    The application listens on <http://localhost:3333> and the actor is `demo`.
```
2chanhaeng, line 57: The example app binds to port 0 (assigned as a random port).

examples/adonisjs/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+    ~~~~ sh
+    curl -H 'Accept: text/html' http://localhost:3333/users/demo
+    ~~~~
+
+    Servers asking for ActivityPub get the actor document instead:
+
+    ~~~~ sh
+    curl -H 'Accept: application/activity+json' \
+      http://localhost:3333/users/demo | jq
```
2chanhaeng, line 76: It's working but not on Fedify CLI(`mise cli -- lookup http://localhost:42817/users/demo -a`).

2chanhaeng (changes requested):
Since I'm not familiar with AdonisJS, I used the Claude Code:claude-fable-5 package for the review, and for the example review, I ran it myself and commented on the parts that didn't work.
AI disclosure is included the message title of 3c043cbd53d78b970b6051e81b15baae792f7134.

packages/adonisjs/src/fedify_middleware.ts
scammo: No,  the federation is built exactly once, at boot. fedify uses bind() so the container does call the factory per make(), but the factory memoises the built promise (this.#federation ??= …, provider.ts:107), and ready() triggers the single build eagerly so misconfiguration fails at startup instead of on the first request. It's a bind() + memo rather than singleton() because singleton() would also cache an early E_FEDERATION_NOT_READY rejection and prevent the app from ever starting.

I've added a test (services.test.ts, "builds the federation once, however often it is resolved") for that.

packages/adonisjs/configure.ts
scammo: position only chooses the index inside the server.use([...]) array (insertElement(0) vs addElement in the assembler codemod),  it doesn't affect whether a middleware precedes routing. That's decided by the stack, and "server" already guarantees it: the whole server stack runs before the router, in any order. I verified with a probe that registers Fedify last behind two other server middleware — a route-less actor path is still served, and on a routed path the router only runs after Fedify.

examples/adonisjs/README.md
scammo: I did add an Info in the README. If Port 3333 is available the Application will use this port. Only if this is available it will chose a random port.

packages/adonisjs/src/fedify_middleware.ts
ThisIsMissEm: Generally it's better to use service files, instead: https://docs.adonisjs.com/guides/concepts/container-services#creating-your-own-services

```
// packages/adonisjs/src/services/federation.ts
import app from '@adonisjs/core/services/app'

let federation: Federation

await app.booted(async () => {
  federation = await app.container.make('fedify')
})

export { federation as default }
```

And then the middleware imports that service, generally middleware shouldn't need to make an instance of the configuration.

packages/adonisjs/src/middleware.ts
```diff
@@ -0,0 +1,507 @@
...
+ * to first access and memoises the result.
+ */
+function defineFederationContext<TContextData>(
+  ctx: HttpContext,
+  create: () => RequestContext<TContextData>,
+): void {
+  let cached: RequestContext<TContextData> | undefined;
+  let created = false;
+
+  Object.defineProperty(ctx, "federation", {
```
ThisIsMissEm: Is there a reason to not be using macroable here?

packages/adonisjs/src/fedify_middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+    const config = await ctx.containerResolver.make("fedify.config");
+
+    const Middleware = fedifyMiddleware<ContextData>(
+      federation,
+      config.contextDataFactory,
+      {
+        ignoreRoutePrefixes: config.ignoreRoutePrefixes,
+      },
+    );
```
ThisIsMissEm: This should be in the provider — the code structure here also doesn't follow typical Adonis.js plugins. Have the provider create the middleware instance, which also has access to the configuration in scope, then just call the instance here.

ThisIsMissEm (commented):
It would be wise to use the standard package structure here: github.com/adonisjs/pkg-starter-kit/

examples/adonisjs/README.md
scammo: I had a look and I did clarify the config and readme

scammo:
> It would be wise to use the standard package structure here: github.com/adonisjs/pkg-starter-kit/

Thank you for your valuable input! I already had a look at the link, and this looks interesting. I will try to finalise my changes accordingly (hopefully tomorrow!).

scammo:
Thanks for all the input. I don't have much experience working with AdonisJS, but because I've always had a good experience with Laravel, I think the package will be quite useful for Fedify.
I look forward to receiving any further feedback on this.

dahlia (dismissed):
It might be because I'm not that familiar with AdonisJS, but this looks good to me!

packages/adonisjs/package.json
```diff
@@ -0,0 +1,126 @@
...
+      "default": "./dist/services/builder.js"
+    },
+    "./package.json": "./package.json"
+  },
+  "files": [
+    "dist/",
+    "package.json"
+  ],
+  "scripts": {
+    "clean": "node --eval \"require('node:fs').rmSync('dist', { recursive: true, force: true })\"",
```
2chanhaeng: Is there any reason to add this script?

pnpm-workspace.yaml
```diff
@@ -59,6 +61,8 @@ allowBuilds:
   workerd: true
 
 catalog:
+  "@adonisjs/assembler": ^8.0.0
+  "@adonisjs/core": ^7.0.0
```
2chanhaeng: Some of methods in the packages need higher version.

packages/adonisjs/README.md
```diff
@@ -0,0 +1,380 @@
...
+
+`fedifyMiddleware` is the package's main integration function and its default
+export.  It takes a `Federation` and a context-data factory, and has nothing to
+do with the IoC container, so an application that would rather build its own
+`Federation` can skip the provider entirely:
+
+~~~~ typescript
+// start/kernel.ts
+import fedifyMiddleware from '@fedify/adonisjs'
+import federation from '#federation/main'
```
2chanhaeng: Using subpath aliases without any explanation can cause confusion for users following along. If there's no need to define an alias, it seems better to avoid using one altogether.

packages/adonisjs/src/fedify_middleware.ts
```diff
@@ -0,0 +1,459 @@
...
+
+      /**
+       * Fedify has a route but cannot satisfy `Accept` (a browser at an actor
+       * URL).  AdonisJS gets a chance to serve its own representation first;
+       * 406 is only the answer if it has nothing either.  `Vary: Accept` is
+       * set before delegating and merged back in afterwards.
+       */
+      onNotAcceptable: async () => {
+        outcome.delegated = "notAcceptable";
+        ctx.response.header("Vary", "Accept");
```
2chanhaeng: `header()` replaces, so an upstream `Vary: Origin` (e.g. from CORS) is lost here and `mergeAcceptIntoVary` below can't recover it — this can cause CORS failures behind a shared cache. Use `mergeAcceptIntoVary(ctx)` here too, and add a test with an upstream middleware setting `Vary`.

packages/adonisjs/src/fedify_middleware.ts
```diff
@@ -0,0 +1,459 @@
...
+    },
+    { highWaterMark: 0 },
+  );
+}
+
+/**
+ * Keeps the first entry of a comma-separated authority (`X-Forwarded-Host`
+ * behind two proxies), which `new URL()` would reject.
+ */
+function normalizeAuthority(host: string): string {
```
2chanhaeng, line 131: Only the comma is stripped, but under `trustProxy` `authority()` returns `X-Forwarded-Host` verbatim, so a value like `example.com/x` or `example.com?a` reaches `buildRequestUrl` and shifts the path Fedify sees (`/x/users/alice`, or `/`) while AdonisJS still routes the real path. This lets a client bypass `ignoreRoutePrefixes`. Reject values containing `/ ? # \ @` or whitespace (return `""`) so the existing `PLACEHOLDER_HOST` fallback keeps the path intact.

2chanhaeng (changes requested):
The message 370c472ecf66255aaf2e35dfa1c719b677d9656e includes literally `\n` instead of a new line. Please fix it.

Also, if you use Fable, consider running an iterative code review loop before  requesting reviewing by a human. Fable is good at finding problems or areas for improvement.

pnpm-workspace.yaml
2chanhaeng: `request.authority()` (used in `toRequestUrl` from *packages/adonisjs/src/fedify_middleware.ts*) exists in `@adonisjs/http-server` from `9.x`. But `@adonisjs/core` `7.0.0–7.3.0` depends on `http-server` `^8.0.0/^8.1.1`, where the method is absent. So, an app on core `7.0–7.3` passes the peer check but every request through the middleware throws `TypeError: request.authority is not a function`. Therefore, the minimum version of `@adonisjs/core` should be `7.4`. Please either raise the floor to `^7.4.0`.

2chanhaeng (commented):
Sorry, but the comments on my latest review have not been resolved at all. Could you check if your remote branch was pushed correctly? Or is there any problem with the review?

Also, before merging this PR, please rebase from the main branch.

2chanhaeng (changes requested):

packages/adonisjs/package.json
scammo: I removed it

pnpm-workspace.yaml
scammo: I've raised the Versions

packages/adonisjs/README.md
scammo: I fixed this

packages/adonisjs/src/fedify_middleware.ts
scammo: I fixed this,  so an upstream Vary: Origin survives instead of being overwritten. This raised some additional breaking points which I also fixed, the copy loop merges `Vary` instead of replacing it, and a second regression test covers  the 200 path.

packages/adonisjs/src/fedify_middleware.ts
scammo: Thank you for this information I fxied it. normalizeAuthority: An authority containing / ? # \ @ or whitespace now returns an empty string, so the placeholder host fallback keeps the path intact, and a warning is logged when that happens.

scammo:
@2chanhaeng Sorry for the noise. I made a mistake regarding the commit in the last time. Now everything should be working fine, and hopefully also the renaming of the commit messages should be working. Thank you for your time and effort!

2chanhaeng (dismissed):

dahlia (changes requested):
@scammo I think we can merge this pull request now, but it looks like there are some file conflicts. Could you please resolve those? Thanks!

scammo:
@dahlia Thank you for your information. I think this should now be mergeable.

dahlia (approved):
@scammo Thanks again for your work! I'm merging this. It will ship with Fedify 2.4.0, the next minor release.

Issue #47, opened 2024-05-06
Width and height properties for Documents with image mediaTypes


jfietkau:
Mastodon uses `Document`s with image mediaTypes in the `attachments` list to model image attachments. These documents supply `width` and `height` properties with integer values to let clients know about the expected render size of the image. These properties are not part of the ActivityStreams vocabulary definition of `Document` nor a specified extension by Mastodon, so they're AIUI straight-up out-of-spec.

Nonetheless, it would be useful if Fedify could serialize `width` and `height` properties as part of the JSON-LD if they have been added to a `Document` object, to make it possible for clients to predict image dimensions before the attachments are downloaded.

PR #1020, opened 2026-08-30
Add `@fedify/pglite` with `PgliteKvStore`

### LINKED ISSUE #1018: Add `@fedify/pglite` with `PgliteKvStore`
## Background

`PostgresKvStore` requires a running PostgreSQL server. For local development, tests, and small single-process deployments, [PGlite] (PostgreSQL compiled to WebAssembly, `@electric-sql/pglite`) provides the same SQL without a server and runs on Node.js, Bun, and Deno.

I propose `PgliteKvStore` in `@fedify/pglite`. It takes a `PGlite` client and is otherwise identical to `PostgresKvStore`, sharing its table schema and default table name (the options only omit `unlogged`):

~~~~ typescript
const pg = new PGlite("./data/fedify");
const federation = createFederation({
  kv: new PgliteKvStore(pg),
});
~~~~

[PGlite]: https://pglite.dev/

## Verified

Every SQL statement in `kv.ts` runs unchanged on PGlite 0.5.8 (PostgreSQL 18.3) under Deno 2.9.5, Node.js 22.23.1, and Bun 1.2.22, including `text[]`/`jsonb`/`interval` parameters.

Differences to handle:

 -  A bare JavaScript string bound to `jsonb` fails; values must always be `JSON.stringify`’d and cast from `text`.
 -  Deno needs `--allow-read` (and `--allow-write` for on-disk data) to load the WebAssembly module.
 -  One PGlite instance per data directory; data is not shared across processes.

## Work

 -  Add `PgliteKvStore` in a new `@fedify/pglite` package.
 -  Add `testKvStore()`, a `KvStore` conformance suite, to `@fedify/testing` and run it against `MemoryKvStore` and `PgliteKvStore`.

## How about `PgliteMessageQueue`?

Not provided, since [separate instances do not share PostgreSQL's usual process and shared-memory state](https://github.com/electric-sql/pglite/issues/489), a queue could not distribute work across processes.

2chanhaeng:
Resolves #1018.

Adds `@fedify/pglite`, a new package providing `PgliteKvStore`, following the scope agreed in #1018: a key–value store in its own package, built on a caller-created `PGlite` instance, with no message queue because PGlite does not share data between processes (electric-sql/pglite#489, electric-sql/pglite#892).

`PgliteKvStore` reuses the `PostgresKvStore` schema and default table name, so a PGlite data directory can be migrated to PostgreSQL without conversion.  Values are bound as `text` and cast to `jsonb` because PGlite rejects bare JavaScript strings for `jsonb` parameters.  The store never creates or closes the PGlite instance; ownership stays with the caller, and the manual documents the single-instance, single-isolate constraint.

`@fedify/testing` gains `testKvStore()`, a `KvStore` conformance suite modeled on `testMessageQueue()`.  It runs against `MemoryKvStore` in `@fedify/testing` and against `PgliteKvStore` in `@fedify/pglite`.  Retrofitting the existing adapters to `testKvStore()` and adding PGlite to `fedify init` are left to follow-up issues.

A separate commit lowers the docs build heap cap from 12 GiB to 3 GiB.  The 12 GiB cap predates the Twoslash file system cache and exceeds common dev machines' RAM, which disables V8 GC pressure and triggers kernel OOM kills; measured builds peak around 3 GiB of heap even with a cold cache.

Tested with `mise run test-each pglite testing` on Deno 2.9.5, Node.js 22.23.1, and Bun 1.2.22, plus `mise run check` and `mise run docs:build`.

Assisted-by: Codex:gpt-5.6-sol to write code.
Assisted-by: Claude Code:claude-fable-5 to plan, review and etc.


2chanhaeng:
This is for [#47](https://github.com/fedify-dev/drfed/pull/47) of DrFed. So I assigned @dodok8 as one of the reviewers.

packages/testing/src/kv-tester.ts
```diff
@@ -0,0 +1,211 @@
...
+      await store.set(isolatedKey, index);
+    }
+    for (const [index, isolatedKey] of isolatedKeys.entries()) {
+      strictEqual(await store.get(isolatedKey), index);
+    }
+
+    if (options.testTtl ?? true) {
+      const expiringKey = key("ttl", "expiring");
+      await store.set(expiringKey, "expired", {
+        ttl: Temporal.Duration.from({ milliseconds: 50 }),
```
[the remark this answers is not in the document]
2chanhaeng: Already on [docs/install.md](https://github.com/fedify-dev/fedify/blob/main/docs/install.md#cli-toolchain).
> If you use Deno earlier than 2.7.0, add `--unstable-temporal` to the Deno installation command above.

dodok8 (approved):

packages/pglite/src/kv.ts
```diff
@@ -0,0 +1,294 @@
...
+  KvStore,
+  KvStoreListEntry,
+  KvStoreSetOptions,
+} from "@fedify/fedify";
+import { getLogger } from "@logtape/logtape";
+
+const logger = getLogger(["fedify", "pglite", "kv"]);
+
+function quoteIdentifier(identifier: string): string {
+  return `"${identifier.replaceAll('"', '""').replaceAll(".", '"."')}"`;
```
sij411: some edge cases which might be worth to prevent: 
`MySchema.MyTable` -> "MySchema"."MyTable", `literal.dot` -> "literal"."dot", `schema..dot` -> "schema".""."table"

ofc empty string returns ""

2chanhaeng: Thanks! Addressed in 41cebe9c3258d047e47e8e3e3aff76bc70d27dcc.

dahlia (approved):

PR #1022, opened 2026-09-03
Add `actor-preferred-username-required` lint rule

### LINKED ISSUE #895: Add an `@fedify/lint` rule for actor preferred usernames
## Why

Most fediverse software expects actor objects to expose a stable `preferredUsername`. Fedify applications can omit it, but that tends to make remote display, search, and profile rendering worse. A lint rule can catch this during development without changing Fedify runtime behavior.

## Scope

Add a new `@fedify/lint` rule that warns when an actor returned from `setActorDispatcher()` does not include `preferredUsername`. Follow the existing actor property rule patterns in *packages/lint/src/rules/*, *packages/lint/src/lib/*, and *packages/lint/src/tests/*.

The rule should recognize non-actor return values and avoid false positives in examples where the returned value is not a Fedify vocabulary actor.

## Suggested checks

Run `mise run check-each lint`. If rule documentation is generated or maintained manually, update *docs/manual/lint.md* as well.

Jae-Hyuk-Jang:
Closes #895

## Background

Most fediverse software expects actor objects to expose a stable `preferredUsername`. Fedify applications can omit it, but that tends to make remote display, search, and profile rendering worse. `@fedify/lint` already has rules that check other required actor properties (`id`, `inbox`, `publicKey`, etc.), but none of them cover `preferredUsername`.

## Changes

- Add the `actor-preferred-username-required` rule (*packages/lint/src/rules/actor-preferred-username-required.ts*), warning when an actor dispatcher's return value omits `preferredUsername`. Register it in Deno Lint, ESLint, and Oxlint (*mod.ts*, *index.ts*, *oxlint.ts*).
- `preferredUsername` has no corresponding `Context` getter method, unlike every other required-property rule, so `PropertyConfig.getter` is now optional (*lib/types.ts*) and `actorPropertyRequired()` (*lib/messages.ts*) falls back to a simpler message when it's absent.
- Generalize *lib/test-templates.ts*, which assumed every property has a getter in a couple of places (property-assignment code generation, and the mismatch tests' "wrong getter" picker), and add a dedicated test factory for `preferredUsername` since its setter (`setActorDispatcher`) is self-referential like `id`'s, unlike the generic `createRequiredDispatcherRuleTests` template.
- Add tests in *tests/actor-preferred-username-required.test.ts* and update *tests/integration.test.ts*'s baseline fixture and add a dedicated integration case.
- Document the rule in *docs/manual/lint.md*.
- Add a changelog fragment.

## Testing

- `mise run check-each lint`
- `node --experimental-transform-types --test 'src/tests/**/*.test.ts'` from *packages/lint* (610 tests pass, including oxlint's real-binary lane)

## AI disclosure

This was a mix of work I wrote myself and code Claude Code (`claude-sonnet-5`) wrote after we worked through the design together — in particular, it found and fixed two issues my first attempts missed: `test-templates.ts` generating invalid example code (`ctx.undefined(identifier)`) for a getter-less property, and a type error in the mismatch-test helpers caused by the same change. I reviewed and verified all of it.

### DIFF
--- a/changes.d/lint/actor-preferred-username-rule.md
+++ b/changes.d/lint/actor-preferred-username-rule.md
@@ -1,3 +1,8 @@
+---
+links:
+  '#1022': https://github.com/fedify-dev/fedify/pull/1022
+  '#895': https://github.com/fedify-dev/fedify/issues/895
+---
  -  Added the `actor-preferred-username-required` lint rule, which warns
     when an actor dispatcher's return value does not include a
-    `preferredUsername` property. [[#895]]
+    `preferredUsername` property. [[#895], [#1022] by Jae-Hyuk-Jang]
--- a/changes.d/lint/actor-preferred-username-rule.md
+++ b/changes.d/lint/actor-preferred-username-rule.md
@@ -1,0 +1,3 @@
+ -  Added the `actor-preferred-username-required` lint rule, which warns
+    when an actor dispatcher's return value does not include a
+    `preferredUsername` property. [[#895]]
--- a/docs/manual/lint.md
+++ b/docs/manual/lint.md
@@ -559,9 +559,9 @@ federation
 
 ### `actor-preferred-username-required`
 
-Ensures actors have a `preferredUsername` property.
-
 *This rule is introduced in Fedify 2.4.0.*
+
+Ensures actors have a `preferredUsername` property.
 
 **When this rule applies:**
 The actor dispatcher is configured with `setActorDispatcher()`, but the actor
--- a/docs/manual/lint.md
+++ b/docs/manual/lint.md
@@ -561,6 +561,8 @@ federation
 
 Ensures actors have a `preferredUsername` property.
 
+*This rule is introduced in Fedify 2.4.0.*
+
 **When this rule applies:**
 The actor dispatcher is configured with `setActorDispatcher()`, but the actor
 object doesn't include a `preferredUsername` property.
--- a/docs/manual/lint.md
+++ b/docs/manual/lint.md
@@ -557,6 +557,43 @@ federation
 
 [Object Integrity Proofs]: ./send.md#object-integrity-proofs
 
+### `actor-preferred-username-required`
+
+Ensures actors have a `preferredUsername` property.
+
+**When this rule applies:**
+The actor dispatcher is configured with `setActorDispatcher()`, but the actor
+object doesn't include a `preferredUsername` property.
+
+**Why it matters:**
+Most fediverse software expects actors to expose a stable
+`preferredUsername`.  Omitting it tends to make remote display, search, and
+profile rendering worse, even though Fedify itself doesn't require it.
+
+~~~~ typescript twoslash
+// @noErrors: 2345
+import { createFederation } from "@fedify/fedify";
+import { Person } from "@fedify/vocab";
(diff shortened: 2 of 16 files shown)


changes.d/lint/actor-preferred-username-rule.md
```diff
@@ -0,0 +1,3 @@
+ -  Added the `actor-preferred-username-required` lint rule, which warns
+    when an actor dispatcher's return value does not include a
+    `preferredUsername` property. [[#895]]
```
dahlia: Please add the number of this pull request, and your name for the credit.

Also, you need to run `sacho resolve-links`.

docs/manual/lint.md
```diff
@@ -557,6 +557,43 @@ federation
 
 [Object Integrity Proofs]: ./send.md#object-integrity-proofs
 
+### `actor-preferred-username-required`
```
dahlia, line 560: Since this rule will be added since Fedify 2.4.0, you need a sentence like:

> *This rule is introduced in Fedify 2.4.0.*

packages/lint/src/rules/actor-preferred-username-required.ts
```diff
@@ -0,0 +1,8 @@
+import { properties } from "../lib/const.ts";
+import {
+  createRequiredRuleDeno,
+  createRequiredRuleEslint,
+} from "../lib/required.ts";
+
+export const deno = createRequiredRuleDeno(properties.preferredUsername);
+export const eslint = createRequiredRuleEslint(properties.preferredUsername);
```
dahlia, line 8: `ActorDispatcher` explicitly permits `Tombstone`, and Fedify documents returning one for deleted actors. This rule currently reports a dispatcher that returns only a `Tombstone`, as well as a conditional dispatcher whose other branch returns an actor with `preferredUsername`. That creates a false positive in the recommended config and an error in the strict config, contrary to #895's requirement to skip non-actor return values. Please treat `Tombstone` branches like `null` and add regression tests for both direct and mixed returns.

packages/lint/src/lib/const.ts
```diff
@@ -114,6 +114,12 @@ export const properties = {
     requiresIdentifier: true,
     isKeyProperty: true,
   },
+  preferredUsername: {
+    name: "preferredUsername",
+    path: ["preferredUsername"],
```
dahlia, line 119: Fedify's actor constructors [also accept the plural `preferredUsernames` initializer](https://fedify.dev/manual/vocab#properties) for this vocabulary property. For example, `new Person({ preferredUsernames: [identifier] })` makes `actor.preferredUsername` return `identifier` and serializes to `preferredUsername`, but this path only accepts the singular initializer, so it reports a warning in the recommended config and an error under the strict config. Please accept `preferredUsernames` as satisfying the rule and cover it with a regression test.

dahlia (changes requested):

docs/manual/lint.md
```diff
@@ -557,6 +557,45 @@ federation
...
+
+Ensures actors have a `preferredUsername` property.
+
+*This rule is introduced in Fedify 2.4.0.*
```
dahlia, line 562: This sentence should go at the very beginning of this section.

dahlia (changes requested):

dahlia (approved):

PR #1027, opened 2026-09-08
Bound the lifetime of public-key and signature-spec caches

### LINKED ISSUE #1017: Bound the lifetime of public-key and HTTP Message Signatures caches
## Context

While looking into Redis usage on Hackers' Pub, I found that Fedify's public-key cache had grown much larger than expected.

A read-only snapshot from an instance running Fedify 2.3.6 contained:

- 194,424 public-key cache entries, using an estimated 165–170 MiB;
- 14,407 entries remembering the HTTP Message Signatures specification used by remote origins;
- no TTL on the sampled public-key entries.

This led me back to two write paths in Fedify. [`KvKeyCache.set()`](https://github.com/fedify-dev/fedify/blob/81416aeae68098e64869d6a0fdd461447a0f6960/packages/fedify/src/federation/keycache.ts#L63-L80) gives unavailable keys a short TTL, but stores successfully resolved keys without one. [`KvSpecDeterminer.rememberSpec()`](https://github.com/fedify-dev/fedify/blob/81416aeae68098e64869d6a0fdd461447a0f6960/packages/fedify/src/federation/middleware.ts#L4925-L4930) also stores its result without an expiry.

Both values are soft state. A public key can be fetched again, and the preferred signature specification for an origin can be learned again. Keeping either value forever means that a persistent `KvStore` grows with every remote key and origin the server has encountered, including peers it will never contact again.

Redis made the growth easy to see, but the same behavior applies to every persistent `KvStore`.

## Direction

I think these two caches should have bounded lifetimes. Their size should reflect a useful working set rather than the entire history of a Fedify installation.

I do not want to prescribe a particular API yet. The policy could live at the individual write sites, in federation configuration, in a reusable KV wrapper, or partly in a store adapter. The important part is that the resulting behavior should not depend unnecessarily on Redis.

The retention policy needs to balance freshness against remote requests. Expiring a key means fetching it again, which can fail while a peer is temporarily unavailable. On the other hand, Fedify already retries with a fresh key when verification using a cached key fails, so permanent retention is not required for key rotation.

We also need to account for existing entries. Applying a TTL only to future writes will leave every old entry that is never read or updated in place.

## Acceptance criteria

- [ ] Successfully resolved public keys have a documented, bounded retention policy.
- [ ] Remembered HTTP Message Signatures specifications have a documented, bounded retention policy.
- [ ] Expired entries are fetched or learned again without breaking signature verification or delivery.
- [ ] Existing shorter-lived behavior for unavailable keys and other Fedify state is preserved.
- [ ] The behavior is consistent across persistent `KvStore` implementations that support TTL.
- [ ] Existing entries written without a TTL are handled through migration behavior or clear operator guidance.
- [ ] Tests cover expiry and repopulation of both cache families.
- [ ] The KV documentation explains the retention and remote-request tradeoffs.

## Things to decide

- Whether bounded retention should be the default or opt-in
- Whether the two cache families need different lifetimes
- Where applications should configure the policy, if it is configurable
- Whether Fedify should migrate existing entries or document a one-time cleanup procedure

#107 introduced public-key caching to avoid a remote fetch for every verification. #916 and #917 dealt with a similar unbounded-growth problem in circuit-breaker state, including entries created by older versions.

heeoneie:
Fixes #1017.

`KvKeyCache.set()` stored successfully resolved keys without a TTL, and `KvSpecDeterminer.rememberSpec()` stored remembered specs without one, so a persistent `KvStore` grew with every remote key and origin the server had ever encountered.

What this implements
--------------------

This follows the direction settled in #1017:

 -  **No sweep or migration code in the library.** Both the `FederationImpl` coordination option (A) and the exported cleanup function (B) I had sketched are dropped. Both caches are soft state, so an operator can delete everything under the two prefixes and Fedify relearns it.
 -  **TTL at the write sites.** 30 days for cached public keys, 90 days for remembered HTTP Message Signatures specs.
 -  **Both lifetimes are configurable by the application.** `FederationOptions` gains `publicKeyTtl` and `httpMessageSignaturesSpecTtl`, named after the `kvPrefixes` entries they bound, the same way `taskDeduplicationTtl` is named after `kvPrefixes.taskDeduplication`. Both take a `Temporal.DurationLike` and fall back to the defaults above, so behavior is unchanged when they are omitted.
 -  **`KvSpecDeterminer`'s constructor is not broken.** The three existing positional arguments are untouched; the TTL arrives as an optional fourth `KvSpecDeterminerOptions` argument.
 -  **Existing entries are handled by documentation, not code.** They stay unexpired until something overwrites them, which is today's behavior.
 -  **Targets `main` (2.4)**, not a maintenance branch.

How the options reach the caches
--------------------------------

`FederationImpl` normalizes both options and passes them to every cache construction site: four `KvSpecDeterminer` sites and two `KvKeyCache` sites. The second key cache site is not in *middleware.ts* — `handleInbox()` in *handler.ts* constructs its own `KvKeyCache`, and that is the site that actually writes the cache during inbox signature verification. It receives the value through a new `InboxHandlerParameters.publicKeyTtl`. The internal `keyTtl` and `specTtl` options and the existing constructor arguments are unchanged.

One incidental change comes with that wiring. Both `KvKeyCache` sites used to pass the surrounding context object as the options bag, which implicitly supplied `tracerProvider` to `CryptographicKey.fromJsonLd()`. Passing an explicit options literal instead would have dropped that span linkage silently, so `KvKeyCacheOptions` now declares `tracerProvider` and both sites pass it.

Why existing entries are left alone
-----------------------------------

Applying a TTL only to future writes does leave every pre-2.4 entry in place, and that is deliberate here. Touching them would require exactly the sweep machinery this issue decided against: either coordination state on `FederationImpl` plus a marker read in the request path, or an exported function that becomes supported public API. Neither pays for itself for a one-time cleanup of values that are already soft state.

So the KV guide gains a *Clearing legacy cache entries* section instead. It describes both prefixes as defaults rather than fixed values, says what to substitute when `kvPrefixes` or the adapter-level namespacing is overridden, states plainly that leaving old entries alone is a valid choice, and gives concrete commands for Redis and PostgreSQL plus a generic `KvStore.list()` loop for everything else. A companion section, *Bounding how long cache entries live*, documents the two new options and the retention tradeoff: a shorter TTL increases remote requests, and refetching an expired key can fail while the peer is unavailable. The changelog entry says the same thing about pre-2.4 entries carrying no expiry.

Tests
-----

Three new end-to-end cases plus two extended unit cases, all passing locally:

 -  `createFederation()` defaults both TTLs to 30 and 90 days and applies an application-supplied override instead when one is given.
 -  Delivery through an overridden `httpMessageSignaturesSpecTtl`: the federation sends to a peer that rejects RFC 9421 and accepts draft-cavage, so the first delivery double-knocks and remembers the spec, the second skips the extra knock, and the delivery after expiry double-knocks and relearns it. The mock inbox runs `verifyRequest()` on every request it accepts.
 -  Verification through an overridden `publicKeyTtl`: signed activities are posted to an inbox, and the key is fetched, reused while the cache is warm, and refetched after expiry, with all three deliveries returning 202.
 -  `KvKeyCache` covers a cache miss followed by refetching and caching the key. The miss is asserted to be `undefined` rather than `null`, since `null` means the key is known to be unavailable and would make a caller treat the actor as keyless instead of refetching.
 -  `KvSpecDeterminer` covers a remembered spec expiring, falling back to the default, and being remembered again.

None of these tests sleep. Expiry is driven by `ManualClockKvStore`, a `KvStore` decorator in *src/testing/* that keeps the TTLs itself instead of passing them to the store it wraps, and expires an entry only when `advance()` moves its virtual clock. It is not in the package exports, so this adds no test-only API to the published surface, and it patches no global, so Deno, Node.js and Bun all behave identically. The two end-to-end tests wrap it in turn with a `KvStore` that records the TTL of every write, so the TTL assertions still hold after the entries themselves have expired.

The earlier revision waited for real time to pass instead, which was load-sensitive rather than merely slow. On the [Node.js CI job] for this branch, `KvKeyCache cached keys expire after keyTtl` failed with "Expected object to be an instance of `CryptographicKey` but was `undefined`": the 1 ms key TTL had already elapsed before the test could read the value back. The whole test lasted 6.98 ms, less than its own 10 ms sleep, so it died at that read-back and never reached the sleep at all. The TTLs are now stated in the units the feature actually uses -- 30 days for keys, 90 days for specs -- rather than in milliseconds chosen to keep a test fast.

Neutering `advance()` into a no-op fails all five tests that call it, so the assertions still depend on expiry actually happening rather than on the clock never moving.

[Node.js CI job]: https://github.com/fedify-dev/fedify/actions/runs/34366982210/job/102520307837?pr=1027

The full `packages/fedify/src/federation/` suite passes (235 tests, 390 steps, up from 232 tests before this round), so the added TTL on the success path does not disturb the existing key-cache, spec-determiner, or middleware tests. `mise run check` is clean, which covers `check:types`, `check:fmt`, `check:lint`, `check:md`, `check-versions`, `check:workspace-protocol`, `check:fixture-usage` and `sacho check`. `mise run docs:build` completes, which is what compiles the new Twoslash snippets in the manual.

*CHANGES.md* was edited by hand earlier on this branch; it is now regenerated with `sacho sync` instead, and the resulting diff is additions only.

What I could not verify
-----------------------

I have no production instance, so the 30/90-day defaults are reasoned rather than measured, as discussed on the issue. The Redis and PostgreSQL cleanup commands in the guide were written against the current `RedisKvStore` (`keyPrefix` default `"fedify::"`, parts joined with `"::"`) and `PostgresKvStore` (`tableName` default `fedify_kv_v2`, `key text[]`) implementations, but I have not run them against a populated instance of either.

Not included here
-----------------

CodeRabbit flagged that `RedisKvStore.set()` passes `ttl.total("second")` straight to `SETEX`, which rejects any TTL that is not a whole number of seconds. The finding is valid and I reproduced it against Redis 7, but it is a different package, unrelated to this issue, and its adapter tests are gated on `REDIS_URL`. It is filed separately as #1028.

AI usage disclosure
-------------------

Per *AI\_POLICY.md*: this change was AI-assisted. Claude Code drafted the implementation, the tests, and the first version of the documentation and this description; every commit carries an `Assisted-by: Claude Code:claude-opus-5` trailer. I reviewed and edited the result, and the repository's own check tasks (`mise run check`, the standard `deno test` invocation over the federation suite, and `mise run docs:build`) were run on my machine and their output checked before pushing.


packages/fedify/src/federation/keycache.ts
```diff
@@ -7,13 +7,26 @@ export interface KvKeyCacheOptions {
...
+   * The TTL for successfully cached keys.  `30` days by default.
+   *
+   * Entries written by Fedify versions older than 2.4.0 have no TTL and
+   * are left untouched by this option; see the *Clearing legacy cache
+   * entries* section of the key–value store guide if you want to expire
+   * them proactively.
+   * @default `Temporal.Duration.from({ days: 30 })`
+   * @since 2.4.0
+   */
+  keyTtl?: Temporal.Duration;
```
dahlia, line 23: The agreement in #1017 was to make both TTLs configurable. These options only configure internal classes: neither class is exposed through the package entry points, and the federation options do not pass application-supplied TTLs to their constructors. Applications therefore always get the defaults. Please wire both TTLs through the public federation configuration and all relevant cache construction sites, with documentation and tests showing that an application can override them. Keeping the existing constructor arguments is still appropriate.

packages/fedify/src/federation/middleware.test.ts
```diff
@@ -11277,6 +11277,45 @@ test("KvSpecDeterminer", async (t) => {
...
+
+    await determiner.rememberSpec(
+      "example.com",
+      "draft-cavage-http-signatures-12",
+    );
+    await new Promise((resolve) => setTimeout(resolve, 10));
+
+    // Falls back to the default spec once the remembered entry expires.
+    const spec = await determiner.determineSpec("example.com");
+    assertEquals(spec, "rfc9421");
```
dahlia: This checks fallback after expiry, but #1017 also requires expiry and repopulation coverage for both cache families. The new public-key test only reads the underlying `MemoryKvStore`, and neither test exercises fetching or learning the value again. Please cover a cache miss followed by refetching and caching the public key, and spec expiry followed by relearning and remembering the spec. The tests should also verify that signature verification and delivery continue to work through those paths.

docs/manual/kv.md
```diff
@@ -512,6 +512,92 @@ export default {
...
+remembered specs.
+
+Entries written by Fedify 2.3 or earlier have no TTL.  They are *not*
+migrated or expired automatically: they simply stay in your `KvStore` until
+something overwrites them, which is the same behavior Fedify has always had.
+Leaving them alone is a perfectly valid choice—Fedify keeps serving and
+refreshing them as before, and they get a TTL the next time they're written.
+
+If you'd rather not wait for that, you can clear the old entries yourself.
+Both caches live under fixed key prefixes:
```
dahlia: These are the default prefixes, not fixed ones: applications can override both through `kvPrefixes`. Please explain how to adapt all cleanup examples to `kvPrefixes.publicKey` and `kvPrefixes.httpMessageSignaturesSpec`, in addition to the adapter-level prefix or table name. This section also needs the retention tradeoff requested in #1017: shorter TTLs increase remote requests, and refetching an expired key can fail while the peer is unavailable. Describing cleanup as always safe with only a few extra fetches understates that operational cost.

CHANGES.md
```diff
@@ -108,11 +108,28 @@ To be released.
     `esnext.temporal` lib reference.
     [[#823], [#925]]
 
+ -  `KvKeyCache` and `KvSpecDeterminer` now write their cache entries with a
```
dahlia: Please keep this unreleased entry in the Sacho fragment and revert the direct changes to *CHANGES.md*, which is compiled at release time. The fragment should describe the user-visible cache lifetimes and the handling of existing entries, rather than internal class options and constructor changes. Please also follow the fragment conventions: a topic-based filename without the issue-number prefix, an opening past-tense verb such as Changed, and contributor attribution unless you prefer to remain anonymous.

changes.d/fedify/1017-kv-cache-ttl.md
```diff
@@ -0,0 +1,22 @@
+---
+links:
+  '#1017': https://github.com/fedify-dev/fedify/issues/1017
+  '#1027': https://github.com/fedify-dev/fedify/pull/1027
+---
+ -  `KvKeyCache` and `KvSpecDeterminer` now write their cache entries with a
+    TTL, so a `KvStore` that never sees an explicit clear no longer
+    accumulates entries for actors and origins that have stopped
+    federating.  [[#1017], [#1027]]
```
dahlia: Please describe the user-visible cache lifetimes and the handling of existing entries here, rather than internal class options and constructor changes. Please also use a topic-based filename without the issue-number prefix and start the entry with a past-tense verb such as “Changed.”

For contributor credit, please add your own name in Roman letters after the issue and PR references, in the form `[[#1017], [#1027] by John Doe]`, replacing `John Doe` with your name.

packages/fedify/src/federation/keycache.test.ts
```diff
@@ -128,3 +128,36 @@ test("KvKeyCache unavailable entries expire", async () => {
...
+    keyId,
+    new CryptographicKey({ id: keyId }),
+  );
+  // The value is written immediately...
+  assert(await kv.get(["pk", keyId.href]) != null);
+
+  // ...but disappears from the underlying KvStore once keyTtl elapses.
+  await new Promise((resolve) => setTimeout(resolve, 10));
+  assertEquals(await kv.get(["pk", keyId.href]), undefined);
+});
```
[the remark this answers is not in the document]
dahlia: Please address this as well; #1017 explicitly requires expiry and repopulation coverage.

dahlia (changes requested):
Thanks for taking this on! I've left a few comments on TTL configuration, test coverage, and the documentation and changelog. Please address those before we merge.

packages/fedify/src/federation/keycache.ts
heeoneie: Wired both TTLs through the public federation configuration in 425d3225.

`FederationOptions` gains `publicKeyTtl` and `httpMessageSignaturesSpecTtl`, named after the `kvPrefixes` entries they bound, the same way `taskDeduplicationTtl` is named after `kvPrefixes.taskDeduplication`.  Both take a `Temporal.DurationLike` and default to the existing 30 and 90 days, so behavior is unchanged when they are omitted.  `FederationImpl` normalizes them and passes them to every cache construction site: four `KvSpecDeterminer` sites and two `KvKeyCache` sites.

The second key cache site is worth calling out, because it is not in *middleware.ts*: `handleInbox()` in *handler.ts* constructs its own `KvKeyCache`, and that is the site that actually writes the cache during inbox signature verification.  It receives the value through a new `InboxHandlerParameters.publicKeyTtl`.

The internal `keyTtl` and `specTtl` options and the existing constructor arguments are unchanged, as you said was appropriate.

Documentation is in *docs/manual/federation.md* and *docs/manual/kv.md*, and the override coverage is the two `createFederation()` tests described in the test coverage thread.

One incidental change in the same commit, so it does not look unexplained: both `KvKeyCache` sites used to pass the surrounding context object as the options bag, which implicitly supplied `tracerProvider` to `CryptographicKey.fromJsonLd()`.  Passing an explicit options literal instead would have dropped that span linkage silently, so `KvKeyCacheOptions` now declares `tracerProvider` and both sites pass it.

packages/fedify/src/federation/middleware.test.ts
heeoneie: Covered in 821ca0c1.

`KvSpecDeterminer` now covers a remembered spec expiring, falling back to the default, and being remembered again, rather than stopping at the fallback.

For the paths themselves there are two new `createFederation()` tests, which double as the override demonstration requested on the options thread:

 -  Delivery: the federation is configured with `httpMessageSignaturesSpecTtl` and sends to a peer that rejects RFC 9421 and accepts draft-cavage.  The first delivery double-knocks and remembers the spec, the second skips the extra knock, and the delivery after expiry double-knocks and relearns it.  The mock inbox runs `verifyRequest()` on every request it accepts, so the signature is verified on each pass rather than assumed.
 -  Verification: the federation is configured with `publicKeyTtl` and receives signed activities on its inbox.  The key is fetched, reused while the cache is warm, and refetched after expiry, and all three deliveries return 202.

Both use a `KvStore` wrapper that records the TTL of every write, so the TTL assertions still hold after the entries themselves have expired.

packages/fedify/src/federation/keycache.test.ts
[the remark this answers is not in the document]
heeoneie: Done in 821ca0c1.  The test now continues past the store-level check: once the TTL elapses it asserts that `KvKeyCache.get()` reports a miss, then caches the key again and asserts that `get()` returns it.

One thing beyond the suggested diff.  The miss is asserted to be `undefined` rather than `null`, and `nullKeys` is asserted not to contain the key.  Those are different signals in this cache — `null` means the key is known to be unavailable — so a caller that saw `null` would treat the actor as keyless instead of refetching.  Asserting only that the result is not the key would pass on either one.

docs/manual/kv.md
heeoneie: Rewritten in d0b5d925.

The prefixes are now described as defaults rather than fixed values, with a note on each example saying what to substitute when `kvPrefixes` is overridden.  The same applies at the adapter level, which the section now names explicitly: `RedisKvStore.keyPrefix` and `PostgresKvStore.tableName`.

The retention tradeoff has its own section, *Bounding how long cache entries live*, which documents the two new options and states the cost directly: a shorter TTL increases remote requests, and refetching an expired key can fail while the peer is unavailable, so verification that would have succeeded from cache fails instead.  The “always safe, only a few extra fetches” framing is gone from the cleanup section too — it now says clearing carries that same cost, and suggests doing it while peers are reachable.

docs/manual/kv.md
```diff
@@ -512,6 +512,92 @@ export default {
...
+### Clearing entries in `RedisKvStore`
+
+[`RedisKvStore`] stores every key under a shared prefix (`"fedify::"` by
+default, configurable via `RedisKvStoreOptions.keyPrefix`), followed by the
+`KvKey` parts joined with `"::"`.  So with the default prefix, scan for and
+delete the two Fedify caches like this:
+
+~~~~ bash
+redis-cli --scan --pattern 'fedify::_fedify::publicKey::*' | xargs -r redis-cli del
+redis-cli --scan --pattern 'fedify::_fedify::httpMessageSignaturesSpec::*' | xargs -r redis-cli del
```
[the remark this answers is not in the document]
heeoneie: Applied in d0b5d925.  Both cleanup examples now materialize the full key list before deleting anything: the Redis example writes the scan output to a file and deletes from that, and the `KvStore.list()` example collects the keys into an array first, since iterating and deleting concurrently has the same skip hazard.

changes.d/fedify/1017-kv-cache-ttl.md
heeoneie: Rewritten in a9fcaf63.

The fragment is now *changes.d/fedify/kv-cache-ttl.md*, topic-based and without the issue-number prefix.  The entry starts with “Changed” and describes the user-visible cache lifetimes, the public options that configure them, the retention tradeoff, and how entries written by earlier versions are handled; the internal class options and constructor shapes are gone.  Credit is `[[#1017], [#1027] by Heewon Chae]`.

*CHANGES.md* had been edited by hand earlier on this branch, so I regenerated it with `sacho sync` rather than editing it again.  The resulting diff is additions only and leaves every other unreleased entry untouched.

One difference from the hand-edited version: the entry now sits at the top of the `@fedify/fedify` section rather than the bottom.  That is `sacho sync`'s own fragment ordering, not a placement I chose.

packages/fedify/src/federation/keycache.ts
```diff
@@ -76,7 +90,9 @@ export class KvKeyCache implements KeyCache {
     }
     this.nullKeys.delete(keyId.href);
     const serialized = await key.toJsonLd(this.options);
-    await this.kv.set([...this.prefix, keyId.href], serialized);
+    await this.kv.set([...this.prefix, keyId.href], serialized, {
+      ttl: this.keyTtl,
+    });
```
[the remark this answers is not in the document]
heeoneie: The finding is valid, and I reproduced it against Redis 7 before filing: `RedisKvStore.set()` passes `ttl.total("second")` straight to `SETEX`, which rejects the argument unless it is a whole number of seconds.  It is slightly broader than sub-second durations — `1500` milliseconds fails the same way, and a zero duration fails with a different message.

I am not fixing it in this pull request.  It is in a different package with different owners, it is unrelated to the review this branch is answering, and per *CONTRIBUTING.md* a non-trivial change should have an accepted issue first.  There is also a verification problem: *packages/redis/src/kv.test.ts* is gated on `REDIS_URL` and skips entirely without a live server, so adding a test here would mean shipping one I had not actually run.

Filed separately as #1028, with the reproduction and the full table of failing durations.

packages/fedify/src/federation/middleware.test.ts
```diff
@@ -11277,6 +11287,271 @@ test("KvSpecDeterminer", async (t) => {
...
+    kv: new MemoryKvStore(),
+    publicKeyTtl: { days: 1 },
+    httpMessageSignaturesSpecTtl: { hours: 12 },
+  });
+  assertEquals(overridden.publicKeyTtl.total("day"), 1);
+  assertEquals(overridden.httpMessageSignaturesSpecTtl.total("hour"), 12);
+});
+
+test("createFederation() applies httpMessageSignaturesSpecTtl to remembered specs", async () => {
+  fetchMock.spyGlobal();
```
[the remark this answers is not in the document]
heeoneie: Fixed in 779f79c6.  Both test bodies are wrapped in `try` / `finally` with `fetchMock.hardReset()` in the `finally` block, so the global fetch spy is restored even when an assertion fails partway through.

This also matches the pattern already used elsewhere in this file, which wraps `hardReset()` in a `finally` in about sixteen places, so it is the existing convention rather than a new one.

packages/fedify/src/federation/middleware.test.ts
```diff
@@ -11277,6 +11287,274 @@ test("KvSpecDeterminer", async (t) => {
...
+      };
+    });
+
+    const keyId = "https://example.com/person2#key3";
+    const kv = new TtlRecordingKvStore();
+    const federation = createFederation<void>({
+      kv,
+      documentLoaderFactory: () => mockDocumentLoader,
+      contextLoaderFactory: () => mockDocumentLoader,
+      publicKeyTtl: { milliseconds: 250 },
```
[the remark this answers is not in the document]
dahlia: Please address the timing races in both the unit and integration tests. The [Node.js CI job] now provides evidence for the earlier 1 ms TTL finding: `KvKeyCache cached keys expire after keyTtl` failed with `Expected object to be an instance of "CryptographicKey" but was "undefined".` The two 250 ms integration tests discussed here passed in that run, so this does not demonstrate a failure in those tests. Controlling the clock would let all three tests check cache hits and expiry without depending on runner speed.

[Node.js CI job]: https://github.com/fedify-dev/fedify/actions/runs/34366982210/job/102520307837?pr=1027

dahlia (changes requested):

2chanhaeng (dismissed):

packages/fedify/src/federation/middleware.test.ts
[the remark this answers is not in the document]
heeoneie: Addressed in f580e0d, and the CI job you linked turns out to show something narrower than a slow runner.

That run reports `duration_ms: 6.976622` for `KvKeyCache cached keys expire after keyTtl`, and the test's own sleep was 10 ms — so it never reached the sleep. The assertion that failed is the `assertInstanceOf(await cache.get(keyId), CryptographicKey)` immediately after the write, which asserts the entry is still *there*; the raw `kv.get()` on the same entry one line above it had just passed. The 1 ms TTL elapsed between those two lines. The test was not waiting too little for expiry, it was racing the TTL in the opposite direction, on the warm-cache half. A wider margin would have hidden that rather than fixed it, which is why I did not take the 2 s / 2.5 s bump proposed at the top of this thread.

So the clock is now the test's to move. `ManualClockKvStore` is a `KvStore` decorator in *src/testing/*: it keeps the TTLs itself instead of passing them to the store it wraps, and an entry only starts reading as missing once `advance()` moves its virtual clock. `MemoryKvStore` already decides expiry by comparing timestamps rather than by scheduling timers, so nothing has to wait at all. It is not reachable through the package exports, so this adds no test-only API to the published surface, and it patches no global, so Deno, Node.js and Bun all take the same path. All five tests that observe a TTL now use it and none of them sleep — the two `KvKeyCache` unit tests, the `KvSpecDeterminer` expiry step, and the two `createFederation()` integration tests, the 250 ms ones discussed on this thread. Their TTLs are also stated in the units the feature actually uses, 30 days for keys and 90 days for specs, rather than in milliseconds picked to keep a test fast.

To confirm the assertions still depend on expiry happening rather than on the clock simply never moving, I replaced `advance()` with a no-op and re-ran: all five fail. With it restored, the full `packages/fedify/src/federation/` suite is 235 passed / 390 steps.

The Tests section of the PR description is updated in the same push.

The two workflows on this head are still `action_required`; they need a maintainer to approve the run before `test-node` can report. Could you kick those off when you get a chance?

packages/fedify/src/testing/kv.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+   * Drops the entry if its TTL has elapsed.  Mirrors {@link MemoryKvStore},
+   * which treats an entry as live up to and including its expiration instant.
+   */
+  async #evictIfExpired(key: KvKey): Promise<void> {
+    const encodedKey = this.#encodeKey(key);
+    const expiration = this.#expirations.get(encodedKey);
+    if (expiration == null) return;
+    if (this.#now.until(expiration).sign >= 0) return;
```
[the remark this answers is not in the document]
heeoneie: The retention at `#now === expiration` is deliberate: it mirrors `MemoryKvStore`, which is the store this decorator stands in for.

`MemoryKvStore` drops an entry only when `until(expiration).sign < 0`, and it does so in all three places that read — `get()` (kv.ts:117), `cas()` (kv.ts:164), and `list()` (kv.ts:194). An entry is therefore live *up to and including* its expiration instant, and gone one nanosecond later. `#evictIfExpired()` mirrors `get()`/`cas()` and `#evictAllExpired()` mirrors `list()`, so `sign >= 0` puts the decorator on the same side of the boundary.

This was not accidental. The comparison and the docstring naming the mirror (“Mirrors `MemoryKvStore`, which treats an entry as live up to and including its expiration instant”) went in together in the same commit, f580e0d8 — the only commit this file has.

Moving the decorator to expire at the boundary would make it strictly stricter than every real `KvStore` it substitutes for, which inverts what it is for: a test could then pass against the decorator and fail against `MemoryKvStore`.

The observation itself is correct, though — `advance(ttl)` alone does not make the entry missing, and nothing asserted that either way. So I have taken the assertion half of the suggestion and left the comparisons alone. `packages/fedify/src/testing/kv.test.ts` is new and pins the boundary from both sides for `get()`, `list()`, and `cas()`:

```ts
kv.advance(ttl);
assertEquals(await kv.get(["k"]), "v");

kv.advance({ nanoseconds: 1 });
assertEquals(await kv.get(["k"]), undefined);
```

To check the assertions actually bite, I applied the change you asked for — `sign >= 0` → `sign > 0` in both methods — and re-ran: three of the new tests fail, while the pre-existing expiry tests in *keycache.test.ts* and *middleware.test.ts* stay green. That confirms your point that the existing tests advance past the TTL and would not notice a boundary change; it just means the gap was in the assertions rather than in the comparison. Full `@fedify/fedify` suite with the new file: 427 passed, 0 failed.

`MemoryKvStore` is untouched, as you noted it should be.

packages/fedify/src/federation/middleware.test.ts
```diff
@@ -11277,6 +11288,279 @@ test("KvSpecDeterminer", async (t) => {
...
+      });
+      return new Response(null, { status: key == null ? 401 : 202 });
+    });
+
+    const kv = new TtlRecordingKvStore();
+    const federation = createFederation<void>({
+      kv,
+      documentLoaderFactory: () => mockDocumentLoader,
+      contextLoaderFactory: () => mockDocumentLoader,
+      httpMessageSignaturesSpecTtl: { days: 90 },
```
dahlia, line 11428: Please use TTLs that differ from the defaults in both integration tests, for example 7 days for public keys and 14 days for remembered specs, and update the TTL assertions and clock advances accordingly. With the current 30-day and 90-day values, these tests would still pass if the configured TTLs stopped reaching the cache constructors, because the caches would fall back to those same defaults. Keep the virtual clock so expiry remains deterministic.

packages/fedify/src/testing/kv.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  async cas(
+    key: KvKey,
+    expectedValue: unknown,
+    newValue: unknown,
+    options?: KvStoreSetOptions,
+  ): Promise<boolean> {
+    if (this.#inner.cas == null) {
+      throw new TypeError("The wrapped KvStore does not support cas().");
```
dahlia, line 116: Please expose `cas` only when the wrapped store supports it. `KvStore.cas` is optional, and `planDeduplication()` checks `kv.cas != null` to choose the CAS path when the queue has no native deduplication. This decorator always exposes the method, so wrapping a store without CAS makes task enqueueing throw instead of using the configured open fallback. Please preserve the wrapped store’s capability and add a regression test covering a store without CAS.

dahlia (changes requested):
Thanks for fixing the timing issues. I've left two follow-up comments on testing TTL overrides and preserving the wrapped store's CAS support.  Please address these before we merge.

packages/fedify/src/federation/middleware.test.ts
heeoneie: You were right: with 30 and 90 days, cutting the configured TTL off from the caches went unnoticed.  Both integration tests now configure non-default TTLs, 7 days for `publicKeyTtl` and 14 days for `httpMessageSignaturesSpecTtl`, and the TTL assertions and virtual clock advances follow (8 and 15 days).  The clock is still `ManualClockKvStore`'s, so expiry stays deterministic.

To measure it, I dropped the TTL argument at every `KvKeyCache` and `KvSpecDeterminer` construction site and ran the two tests:

| Tests | TTL argument cut | Result |
| --- | --- | --- |
| Old values (30 / 90 days) | yes | 2 passed — the gap you described |
| New values (7 / 14 days) | yes | 2 failed |
| New values, TTL assertions removed | yes | 2 failed — the clock-advance assertions alone catch it |

The full `@fedify/fedify` suite passes on all three runtimes: Deno 429 passed (533 steps), 0 failed, 2 ignored; Node.js 949 of 950 passed, 0 failed, 1 skipped; Bun 426 passed, 0 failed, 1 skipped.  `mise run check` passes.

Fixed in c759c3596.

packages/fedify/src/testing/kv.ts
heeoneie: `ManualClockKvStore` now defines `cas` only when the wrapped store has one.  The constructor assigns it from the wrapped store's `cas`; otherwise the property stays `undefined`, so `planDeduplication()` sees the same capability it would see on the wrapped store.

Two regression tests wrap a store without `cas()`.  One checks that the decorator exposes no `cas`.  The other goes through `createFederation()`: it enqueues a task with a `deduplicationKey` under `taskDeduplicationFallback: "open"` and checks that the task is enqueued without the key.  Restoring the previous `cas()` method makes both fail, and the second fails with exactly the `TypeError` you predicted (`The wrapped KvStore does not support cas().`), thrown from the enqueue.

Since `cas` is now optional on the decorator, the three existing tests that called `cas()` directly use `cas?.()`.  That does not weaken them: if `cas` were missing, the call would yield `undefined` and their `true` / `false` assertions would fail.

The full `@fedify/fedify` suite passes on all three runtimes: Deno 429 passed (533 steps), 0 failed, 2 ignored; Node.js 949 of 950 passed, 0 failed, 1 skipped; Bun 426 passed, 0 failed, 1 skipped.  `mise run check` passes.

Fixed in 4f762f208.

dahlia (approved):

packages/fedify/src/federation/keycache.ts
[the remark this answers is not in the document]
heeoneie: Status update for anyone reading this after the merge: the fix for #1028 is open as #1034, targeting *2.0-maintenance* per @dahlia's direction on the issue. It rounds the `SETEX` TTL up to a whole second with a floor of one, and the reproduction table from #1028 was re-run on that branch against Redis 7.4.11.

Leaving this thread open until #1034 lands rather than resolving it now.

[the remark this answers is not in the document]
heeoneie: Handled in #1034, which landed on *2.0-maintenance* on 2026-09-18 (`78c8a5f`).

`RedisKvStore.set()` no longer passes `ttl.total("second")` to `SETEX` directly; it goes through `expirySeconds()`, which is `Math.max(1, Math.ceil(ttl.total("second")))`, so the sub-second, `1500` ms and zero-duration cases from the #1028 table all become whole seconds with a floor of one. It shipped in 2.0.28 and has since been merged upward through the maintenance branches into `main` (2.3.8, 2026-09-21).

Resolving this thread.

PR #1029, opened 2026-09-10
Add a Netlify Blobs-backed `KvStore`

### LINKED ISSUE #1010: Add a Netlify Blobs-backed `KvStore`
## Background

`@fedify/netlify` currently connects Fedify’s `MessageQueue` API to Netlify Async Workloads, but users must bring a separate `KvStore`. The documentation uses `PostgresKvStore` with Netlify Database because messages with an `orderingKey` require an `orderingKv` that supports `cas()`.

That setup is correct, but a Postgres database is a poor fit for small Fedify applications whose persistent state consists mostly of actor keys, follower records, delivery bookkeeping, and short-lived deduplication markers.

I encountered this while running a mostly static Fedify-backed blog on Netlify. During the current billing period, Netlify Database consumed 205.85 GB-hours and 2,058.5 of the project’s 2,343.8 credits. ActivityPub requests arrived often enough that the database rarely reached Netlify’s five-minute inactivity threshold. The data itself is tiny.

[Netlify Blobs](https://docs.netlify.com/build/data-and-storage/netlify-blobs/) is designed for frequently read, infrequently written data and can act as a simple key-value store. It persists site-wide data across deploys and is available from both Functions and Edge Functions. I propose adding a `NetlifyBlobsKvStore` implementation to `@fedify/netlify`.

## Proposed API

The adapter could accept a Netlify Blobs `Store`, or a structurally compatible subset, rather than creating the store itself:

```typescript
import { getStore } from "@netlify/blobs";
import {
  NetlifyBlobsKvStore,
  NetlifyMessageQueue,
} from "@fedify/netlify";

const store = getStore({
  name: "fedify",
  consistency: "strong",
});

const kv = new NetlifyBlobsKvStore(store);
const queue = new NetlifyMessageQueue({
  client: new AsyncWorkloadsClient(),
  orderingKv: kv,
});
```

Accepting a store object would keep `@netlify/blobs` optional for users who only need `NetlifyMessageQueue`. It would also make the adapter easier to test without a live Netlify project.

The implementation would encode structured `KvKey` values as blob keys, store values as JSON, use blob metadata for logical expiration, and implement prefix scans with `Store.list()`.

## Conditional writes

Current versions of `@netlify/blobs` support atomic conditional writes through `onlyIfNew` and ETag-based `onlyIfMatch`. These operations appear sufficient to implement per-key `cas()`:

- `cas(key, undefined, value)` can use `onlyIfNew`.
- `cas(key, expected, value)` can perform a strongly consistent `getWithMetadata()`, compare the stored value, and write with `onlyIfMatch`.
- `cas(key, expected, undefined)` cannot use a conditional delete because the public Blobs API does not expose one. It may need an ETag-guarded tombstone instead.

A CAS-capable implementation matters here. `NetlifyMessageQueue` uses CAS to reserve and advance ordering sequences, and Fedify also uses it for task deduplication when the queue does not provide native deduplication.

The tombstone case needs more design work. Netlify Blobs has no server-side TTL, so both expired values and conditional deletion markers require logical expiration and eventual reclamation. Cleanup must not delete a value that replaced an expired entry after the cleanup read. We should determine whether the underlying API can support conditional deletion or whether the adapter needs a different representation.

`setJSON()` did not reliably pass conditional-write headers until `@netlify/blobs` 10.7.12, so the adapter should either require that version or serialize JSON explicitly and use `Store.set()`.

## Expected work

- Add `NetlifyBlobsKvStore` to `@fedify/netlify`.
- Implement `get()`, `set()`, `delete()`, `list()`, TTL handling, and, if the deletion semantics can be made safe, `cas()`.
- Use strong consistency for operations that participate in CAS.
- Reject or document encoded keys that exceed Netlify Blobs’ 600-byte key limit.
- Cover concurrent creates and updates, logical expiration, prefix listing, tombstones, and pagination in unit tests.
- Add a Netlify Dev integration test for persistence and conditional writes.
- Document a database-free setup using `NetlifyBlobsKvStore` with `NetlifyMessageQueue`.
- Explain the remaining consistency and cleanup tradeoffs in the API documentation.

sij411:
Closes #1010.

Summary
-------

 -  Add `NetlifyBlobsKvStore` with expiration, prefix listing, and atomic compare-and-set operations.
 -  Replace PostgreSQL in the Netlify Dev fixture and explicitly test persistence and conditional writes.
 -  Document Netlify Blobs as the primary KV option and PostgreSQL as an alternative.
 -  Patch the filesystem-backed local Blobs server to return ETags, pending https://github.com/netlify/primitives/pull/772.

Testing
-------

 -  `mise run check`
 -  Netlify package Node suite: 63 passed, 0 failed, including the Netlify Dev integration test
 -  Frozen Deno and pnpm lockfile validation

AI assistance
-------------

Codex (`gpt-5.6-sol`) assisted with implementation review, CAS and tombstone changes, tests, dependency debugging, documentation, the changelog, and this pull request description. I reviewed the changes and manually ran the Netlify Dev integration test.


CHANGES.md
```diff
@@ -250,6 +250,10 @@ To be released.
 
 ### @fedify/netlify
 
+ -  Added `NetlifyBlobsKvStore`, a Netlify Blobs-backed key–value store with
+    expiration, prefix listing, and atomic compare-and-set operations.  Netlify
+    deployments can now persist Fedify state and preserve ordered queue
+    delivery without a separate database.  [[#1010], [#1029] by Jiwon Kwon\]
```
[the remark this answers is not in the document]
sij411: i think it's Sacho behaviour?

pnpm-workspace.yaml
```diff
@@ -80,6 +80,7 @@ catalog:
   "@multiformats/base-x": ^4.0.1
   "@nestjs/common": ^11.0.1
   "@netlify/async-workloads": ^0.0.106
+  "@netlify/blobs": ^11.0.3
```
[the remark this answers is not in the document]
sij411: @dahlia i fixed the `@netlify/blobs` version to 11.0.3 for now since it's the latest version and the blob Etag issue isn't resolved yet. would it be problematic in the future? Related: https://github.com/fedify-dev/fedify/pull/1029/commits/fe1fcd1a5feba3be4489e3524d3464c6623c0f6b

packages/netlify/src/kv.ts
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * {@inheritDoc KvStore.cas}
+   */
+  async cas(
+    key: KvKey,
+    expectedValue: unknown,
+    newValue: unknown,
+    options?: KvStoreSetOptions,
+  ): Promise<boolean> {
+    while (true) {
```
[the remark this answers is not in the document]
sij411: Im following existing precedent

packages/netlify/src/kv.ts
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Creates a new Netlify Blobs-backed key-value store.
+   * @param store The Netlify Blobs store to use.
+   */
+  constructor(store: Store) {
+    this.#store = store;
+  }
+
+  private serializeKey(key: KvKey): string {
+    const serialized = JSON.stringify(key);
```
dahlia: Netlify Blobs SDK appends the key directly to the request URL. JSON serialization leaves `#` and `?` unescaped, so key components such as `https://example.com/actor#key-1` and `https://example.com/actor#key-2` produce the same HTTP request path. The fragment never reaches the server. This affects `KvKeyCache`, which includes `keyId.href` in its keys, and can overwrite a different cached key.

Please use a reversible URL-safe encoding, preserve tuple-prefix matching in `list()`, and enforce the 600-byte limit after encoding. Add regression coverage through the SDK request path for fragment and query characters; the current map-backed mock bypasses URL handling.

packages/netlify/src/kv.ts
```diff
@@ -0,0 +1,177 @@
...
+ * @since 2.4.0
+ */
+export class NetlifyBlobsKvStore implements KvStore {
+  readonly #store: Store;
+
+  /**
+   * Creates a new Netlify Blobs-backed key-value store.
+   * @param store The Netlify Blobs store to use.
+   */
+  constructor(store: Store) {
```
dahlia: I reproduced the npm public API failure by compiling `new NetlifyBlobsKvStore(getStore({ name: "fedify", consistency: "strong" }))` against the built package. TypeScript reports `TS2345: Types have separate declarations of a private property 'client'`.

With `@netlify/blobs` only in `devDependencies`, tsdown copies its `Store` declaration into the published types. Its private members make that copy incompatible with the actual SDK class. Please keep the SDK type external with suitable dependency metadata, or accept a structural interface containing only the methods the adapter needs. I prefer the latter to preserve the optional dependency and allow test doubles. Please also check the documented example against the built declarations, since source-level Deno tests do not cover this failure.

packages/netlify/README.md
```diff
@@ -1,18 +1,19 @@
...
 [![@fedify@hackers.pub][@fedify@hackers.pub badge]][@fedify@hackers.pub]
 
 *This package is available since Fedify 2.4.0.*
 
-This package connects [Fedify]'s [`MessageQueue`] API to
-[Netlify Async Workloads].  `NetlifyMessageQueue` publishes durable events,
-while `createNetlifyQueueHandler()` turns a Netlify Function into their
-consumer.
+This package connects [Fedify]'s [`KvStore`] and [`MessageQueue`] APIs to
+Netlify.  `NetlifyBlobsKvStore` stores federation state in [Netlify Blobs],
```
dahlia, line 13: Please add `NetlifyBlobsKvStore` to *docs/manual/kv.md* as well. The contribution guide requires new `KvStore` implementations to be documented there, but this PR only adds the setup to the package README. The manual should include the database-free setup, strong-consistency configuration, encoded-key limit, and retained expired blobs and tombstones, with PostgreSQL kept as an alternative.

dahlia (changes requested):

dahlia:
A clarification on @coderabbitai's [above two outside-diff comments](https://github.com/fedify-dev/fedify/pull/1029#pullrequestreview-5175035252):

The dependency comment concerns the same public type issue already tracked in my earlier review. The documented example fails with TS2345 even with `@netlify/blobs` installed: tsdown copies the SDK's `Store` declaration into the published types, and its private `client` member makes that copy incompatible with the actual SDK class. Keeping the type external through an optional peer dependency is one possible fix. I would prefer a structural interface containing only the methods this adapter needs, which also keeps Blobs optional and makes test doubles easier to supply.

I would not treat the CAS retry comment as a blocker. Sustained contention can prolong the operation, but returning `false` after a retry limit would not bound `updateOrderingState()`, since that caller retries whenever CAS returns `false`. The current pattern also matches `DenoKvStore.cas()`, as discussed in the earlier resolved thread. Backoff could be considered separately; I do not think this PR needs the proposed retry limit.

packages/netlify/README.md
sij411: Added: https://github.com/fedify-dev/fedify/pull/1029/commits/d3de6e251cff55a088bcf8da2b832c9256ddcb7d

dahlia (approved):

Issue #1031, opened 2026-09-14
`PostgresKvStore` `initialized` option skips JSON serialization detection


heeoneie:
Description
-----------

`PostgresKvStore` has the same defect as #1014, in the same package and from the same cause: when it is constructed with `initialized: true`, the driver JSON serialization probe is skipped along with the schema DDL, so values are stored as JSONB strings instead of JSONB objects.

In *packages/postgres/src/kv.ts*, `#driverSerializesJson` defaults to `false` and is assigned only at the end of `initialize()`, after the `CREATE UNLOGGED TABLE` statement.  The constructor sets `#initialized = options.initialized ?? false`, so `initialize()` returns at its first line and reaches neither the DDL nor the probe.  `#json()` therefore always takes the `JSON.stringify()` branch, and postgres.js `sql.json()` serializes the result a second time.

There is no test covering the `initialized` option in *packages/postgres/src/kv.test.ts* today.

Reproduction
------------

Measured against *2.0-maintenance* at f0c40e4a (`@fedify/postgres` 2.0.27) on PostgreSQL 16.14 with postgres.js 3.4.8, and with `log_statement = 'all'` so the statements the store actually issues can be read off the server log:

~~~~ ts
import { PostgresKvStore } from "@fedify/postgres";
import postgres from "postgres";

const sql = postgres(Deno.env.get("DATABASE_URL")!);
const table = "kv_probe_repro";

await sql`
  CREATE UNLOGGED TABLE IF NOT EXISTS ${sql(table)} (
    key text[] PRIMARY KEY,
    value jsonb NOT NULL,
    created timestamp with time zone DEFAULT CURRENT_TIMESTAMP,
    ttl interval
  );
`;
await sql`SELECT 'MARKER START'`;

const store = new PostgresKvStore(sql, { tableName: table, initialized: true });
const value = { keyPair: { id: "https://example.com/actor#main-key" }, n: 1 };
await store.set(["cache", "a"], value);

const [row] = await sql`
  SELECT value, jsonb_typeof(value) AS json_type FROM ${sql(table)};
`;
console.log("stored json_type   :", row.json_type);
console.log("stored typeof      :", typeof row.value);

const readBack = await store.get(["cache", "a"]);
console.log("get() typeof       :", typeof readBack);
console.log(
  "get() === original :",
  JSON.stringify(readBack) === JSON.stringify(value),
);

await sql`SELECT 'MARKER END'`;

// A second store that never passes `initialized`, reading the row written
[...]

PR #1032, opened 2026-09-14
Separate the JSON serialization probe from the schema DDL

### LINKED ISSUE #1014: `PostgresMessageQueue` `initialized` option skips JSON serialization detection
Description
-----------

`PostgresMessageQueue` skips JSON serialization detection when constructed with `initialized: true`. With postgres.js, messages are then serialized twice and stored as JSONB strings instead of JSONB objects.

The `initialized` option is documented as indicating that the queue table has already been initialized. It should skip schema DDL, but it should not skip driver-specific runtime setup.

In *packages/postgres/src/mq.ts*, `#driverSerializesJson` defaults to `false` and is set by `driverSerializesJson()` only at the end of `#doInitialize()`. When `initialized: true`, `initialize()` returns before `#doInitialize()` runs. `#json()` therefore calls `JSON.stringify()` before passing the result to postgres.js `sql.json()`, whose serializer calls `JSON.stringify()` again.

Reproduction
------------

This occurs with `@fedify/postgres` 2.3.6 and postgres.js 3.4.8:

~~~~ ts
import { PostgresMessageQueue } from "@fedify/postgres";
import postgres from "postgres";

const sql = postgres(process.env.DATABASE_URL!);

await sql`
  CREATE TABLE IF NOT EXISTS fedify_message_v2 (
    id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
    message jsonb NOT NULL,
    delay interval DEFAULT '0 seconds',
    created timestamp with time zone DEFAULT CURRENT_TIMESTAMP,
    ordering_key text
  )
`;

const queue = new PostgresMessageQueue(sql, { initialized: true });
await queue.enqueue({ type: "inbox", payload: { id: "example" } });

const [row] = await sql`
  SELECT message, jsonb_typeof(message) AS json_type
  FROM fedify_message_v2
  ORDER BY created DESC
  LIMIT 1
`;

console.log(row.json_type); // "string"
console.log(typeof row.message); // "string"

await sql.end();
~~~~

Without `initialized: true`, the same message is stored as a JSONB object and postgres.js returns it as an object.

Impact
------

Queue consumers receive a string rather than the original message object. Fedify dispatches queued tasks by `message.type`, so the string has no recognized task type. A listener can delete the row without processing the activity or reporting an error, causing silent loss of queued federation work.

Expected behavior
-----------------

`initialized: true` should skip queue schema DDL while preserving the same enqueue and dequeue behavior as the default initialization path.

One possible fix is to run the driver serialization probe independently from schema initialization, then let `initialized` control only the `CREATE TABLE`, `ALTER TABLE`, and `CREATE INDEX` statements.

heeoneie:
Summary
-------

`PostgresMessageQueue`'s `initialized` option skipped the driver JSON
serialization probe along with the schema DDL, so a queue constructed with
`initialized: true` stored every message as a JSONB string instead of a JSONB
object.  This pulls the option's two jobs apart: it now gates the DDL only,
and the probe runs on first use either way.

Targets *2.0-maintenance* rather than *main*, per the branch policy for bug
fixes and @dahlia's direction on the issue.

Related issue
-------------

 -  fixes #1014

A second instance of the same defect, in `PostgresKvStore`, is filed
separately as #1031 and is deliberately **not** touched here.

Reproduction
------------

The script from the issue, abridged to the part that matters, run against a
real PostgreSQL 16.14 with postgres.js 3.4.8:

~~~~ ts
await sql`
  CREATE TABLE IF NOT EXISTS fedify_message_v2 (
    id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
    message jsonb NOT NULL,
    delay interval DEFAULT '0 seconds',
    created timestamp with time zone DEFAULT CURRENT_TIMESTAMP,
    ordering_key text
  )
`;

const queue = new PostgresMessageQueue(sql, { initialized: true });
await queue.enqueue({ type: "inbox", payload: { id: "example" } });

const [row] = await sql`
  SELECT message, jsonb_typeof(message) AS json_type
  FROM fedify_message_v2 ORDER BY created DESC LIMIT 1
`;
console.log(row.json_type, typeof row.message);
~~~~

Before this change it prints `string string`; after it, `object object`.

Cause
-----

`#doInitialize()` does two unrelated things.  It runs the `CREATE TABLE` /
`ALTER TABLE` / `CREATE INDEX` retry loop, and then, at the very end, it sets
`#driverSerializesJson` from the `driverSerializesJson()` probe.

The constructor assigned `options.initialized` straight into `#initialized`,
which is the same flag `initialize()` checks before returning early.  So
`initialized: true` did not skip the DDL — it skipped the whole method, probe
included.  `#driverSerializesJson` then stayed at its `false` default, `#json()`
took the `JSON.stringify()` branch, and postgres.js `sql.json()` serialized the
result a second time.

The downstream cost is silent.  Fedify dispatches queued tasks by
`message.type`, and a string has none, so a listener deletes the row without
processing the activity or raising an error.

Changes
-------

 -  *packages/postgres/src/mq.ts*: split the constructor flag in two.
    `options.initialized` now lands in a new `#skipDdl`, while `#initialized`
    always starts `false`.
 -  *packages/postgres/src/mq.ts*: move the DDL retry loop — including
    `INITIALIZE_MAX_ATTEMPTS`, `isInitializationRaceError()`, and the backoff —
    into `#initializeTable()`, and call it from `#doInitialize()` only when
    `#skipDdl` is unset.  The probe and `#initialized = true` now run
    unconditionally.  `initialize()` itself is untouched.
 -  *packages/postgres/src/mq.ts*: document that the option skips only the
    schema DDL.
 -  *packages/postgres/src/mq.test.ts*: two regression tests, one per half of
    the contract.
 -  *changes.d/postgres/initialized-json-probe.md* and *CHANGES.md*.

Benefits
--------

`initialized: true` stops corrupting queued messages, which is the documented
behaviour it was always supposed to have: skip the schema DDL, keep enqueue and
dequeue identical to the default path.  Deployments that manage their own
schema — the reason the option exists — no longer lose federation work to it.

Verification
------------

Run against PostgreSQL 16.14 in Docker, with `POSTGRES_URL` set so the
Postgres-gated tests actually execute rather than skip.

**The regression test fails without the patch.**  On the unmodified branch it
reports exactly the symptom the issue describes:

~~~~
AssertionError: initialized: true should still store the message as a JSONB object
'string' !== 'object'
~~~~

**Both halves are enforced.**  The second test guards the other direction: if
the `#skipDdl` gate is removed so the DDL always runs, it fails with `Missing
expected rejection`.  A fix that simply stopped honouring `initialized` would
not pass.

**The DDL really is still skipped.**  With `log_statement = 'all'`, comparing
the statements the queue issues between two marker queries:

| statement | `initialized: true` | `initialized: false` |
| --- | --- | --- |
| `CREATE TABLE` / `ALTER TABLE` / `CREATE INDEX` | none | one each |
| `SELECT $1::jsonb` (the probe) | 1 | 1 |
| resulting `jsonb_typeof` | `object` | `object` |

**Tests.**  `@fedify/postgres` across all three runtimes: Deno 25 passed / 0
failed (23 before these tests were added), Node.js 24 passed / 0 failed, Bun 25
passed / 0 failed.  `mise run check` passes.

**`mise run test` does not come back clean here, and it does not on the base
commit either.**  Both this branch and f0c40e4a fail in the same single place:

~~~~
packages/vocab-tools test:bun: error: Test "generateClasses() imports the
browser-safe jsonld entrypoint" timed out after 5141ms
~~~~

`@fedify/vocab-tools` passes 15/15 when run on its own; it only times out when
Deno, Node.js, and Bun are running the whole monorepo concurrently on this
machine.  Everything else reported as failed in that run is mise terminating
the sibling tasks afterwards — `Command failed with signal "SIGTERM"`, with no
`not ok` line anywhere.  I ran the base commit specifically to check this
rather than assume it, and nothing in this patch is in `@fedify/vocab-tools`'
dependency graph.

AI use
------

Claude Opus 5 assisted with the investigation, the patch, and running the
checks above; the commit carries an `Assisted-by` trailer.  Every number in
this description was measured on a real PostgreSQL instance rather than
inferred from reading the code.

Checklist
---------

    feature)? — not a new feature; the `initialized` option's TSDoc is updated
    to say what it skips.
    fix)?
 -  [x] Did you run `mise test` on your machine? — yes; it fails identically
    on this branch and on its base commit, see *Verification*.

Additional notes
----------------

 -  **This does not repair rows already written.**  Messages enqueued by a
    misconfigured process before the upgrade are already stored as JSONB
    strings and will still be dropped when they are dequeued.  Only new
    enqueues are affected.
 -  **One extra statement on the `initialized: true` path.**  The probe is a
    single `SELECT $1::jsonb`, run once per queue instance on first use.  That
    is the entire runtime cost of the fix.
 -  The DDL loop moved rather than changed; the diff reads more cleanly with
    whitespace hidden.


dahlia (approved):
Thanks for working on this!

dahlia:
Your fix is shipped with Fedify 2.0.27, 2.1.23, 2.2.12, and 2.3.7.

PR #1034, opened 2026-09-15
Round the Redis key–value store TTL up to whole seconds

### LINKED ISSUE #1028: `RedisKvStore.set()` rejects any TTL that is not a whole number of seconds
`RedisKvStore.set()` passes `options.ttl.total("second")` straight to Redis `SETEX`, which requires a whole number of seconds.  Any `Temporal.Duration` that is not an exact number of seconds is rejected by the server, so the write fails instead of storing the value with a rounded expiry.

This is reachable from application code: `KvStoreSetOptions.ttl` is public API and accepts any `Temporal.Duration`.  It also became easier to hit in 2.4, where `FederationOptions.publicKeyTtl` and `FederationOptions.httpMessageSignaturesSpecTtl` let an application choose the cache lifetimes directly (#1027).  Fedify's own defaults are all whole seconds, so nothing in the library trips it today.

Environment
-----------

 -  `@fedify/redis` 2.4.0 (`main`, commit 621958bd)
 -  Deno 2.9.5, macOS 26.6.2 (aarch64)
 -  Redis 7 (`redis:7-alpine` in Docker)

Steps to reproduce
------------------

~~~~ typescript
import { RedisKvStore } from "@fedify/redis/kv";
import { Redis } from "ioredis";

const store = new RedisKvStore(new Redis("redis://127.0.0.1:6379"));

await store.set(["k"], "v", { ttl: Temporal.Duration.from({ seconds: 1 }) });
await store.set(["k"], "v", { ttl: Temporal.Duration.from({ milliseconds: 500 }) });
~~~~

Expected behavior
-----------------

Both writes store the value.  A sub-second TTL is expected to be normalized to the smallest expiry Redis can express rather than rejected.

Actual behavior
---------------

The first write succeeds.  The second throws `ReplyError: ERR value is not an integer or out of range`.

The failure is not limited to sub-second durations.  Every duration whose total is not an integer number of seconds fails the same way, and a zero duration fails with a different message:

| TTL               | `ttl.total("second")` | Result                                        |
| ----------------- | --------------------- | --------------------------------------------- |
| 1 second          | `1`                   | stored                                        |
| 1500 milliseconds | `1.5`                 | `ERR value is not an integer or out of range` |
| 500 milliseconds  | `0.5`                 | `ERR value is not an integer or out of range` |
| 1 millisecond     | `0.001`               | `ERR value is not an integer or out of range` |
| 0                 | `0`                   | `ERR invalid expire time in 'setex' command`  |

Suggested fix
-------------

Normalize the argument to a positive integer before calling `SETEX`, so that a sub-second TTL becomes one second rather than an error.  `RedisKvStore` has no `cas()` method, so `set()` is the only affected call site.

The adapter tests are gated on `REDIS_URL` and skip entirely without a live server, so a regression test for this needs a Redis instance to be meaningful.

AI usage disclosure
-------------------

Per *AI\_POLICY.md*: this report was AI-assisted.  The problem was first raised by CodeRabbit on #1027; Claude Code drafted this report and ran the reproduction above against a local Redis 7 container.  I checked the reproduction output and edited the text before filing.

heeoneie:
Summary
-------

`RedisKvStore.set()` passed `options.ttl.total("second")` straight to Redis
`SETEX`, which takes only whole seconds.  Any `Temporal.Duration` that was not
an exact number of seconds was rejected by the server, so the write failed
instead of storing the value with an approximated expiry.  The TTL is now
rounded up to the next whole second, with a floor of one second.

The one-second granularity is `SETEX`'s, not Redis's — `SET` with `PX` expresses
millisecond expiries — so the rounding is a property of the command this
adapter uses rather than of the server.

Targets *2.0-maintenance* rather than *main*, per @dahlia's direction on the
issue: "Confirmed that this reproduces in 2.0.x. Since this is a bug fix,
please target `2.0-maintenance` rather than `main`."  ([comment])  The issue
was filed against *main*; the reproduction below was re-run on
*2.0-maintenance* rather than assumed.

[comment]: https://github.com/fedify-dev/fedify/issues/1028#issuecomment-5604179646

Related issue
-------------

 -  fixes #1028

Reproduction
------------

Against Redis 7.4.11 (`redis:7-alpine` in Docker) on *2.0-maintenance* at
9c8ec49c, writing the same value with each TTL and reading back `PTTL`:

| TTL | `ttl.total("second")` | before | after |
| --- | --- | --- | --- |
| 1 second | `1` | stored, `pttl` 1000 | stored, `pttl` 999 |
| 1500 milliseconds | `1.5` | `ERR value is not an integer or out of range` | stored, `pttl` 1999 |
| 500 milliseconds | `0.5` | `ERR value is not an integer or out of range` | stored, `pttl` 999 |
| 1 millisecond | `0.001` | `ERR value is not an integer or out of range` | stored, `pttl` 999 |
| 0 | `0` | `ERR invalid expire time in 'setex' command` | stored, `pttl` 1000 |

A whole number of seconds is unchanged, so nothing that worked before behaves
differently.

Why round up rather than to the nearest second
----------------------------------------------

Rounding to the nearest second would shorten an expiry by up to half a second.
Rounding up never shortens one, and three things point the same way:

**No other `KvStore` implementation shortens the lifetime a caller asked for.**

| adapter | TTL handling | lifetime visible to callers |
| --- | --- | --- |
| `SqliteKvStore` | `ttl.total({ unit: "milliseconds" })` | exact, to the millisecond |
| `DenoKvStore` | `expireIn: ttl.total("millisecond")` | exact, to the millisecond |
| `PostgresKvStore` | `ttl.toString()` into an `interval` | exact, full precision |
| `WorkersKvStore` | `Math.max(ttl.total("seconds"), 60)` physically, with the requested expiry kept in `metadata.expires` | exact — `get()` returns `undefined`, and `list()` skips the entry, once `metadata.expires` has passed |

`WorkersKvStore` is worth reading carefully, because its 60-second clamp looks
like a precedent for extending a lifetime and is not one.  It extends only the
*physical* expiry, then hides that behind a millisecond-precision check on
read, so a caller observes exactly the lifetime it asked for.  No adapter here
extends a caller-visible lifetime today, and this change makes `RedisKvStore`
the first to do so.  That is the honest framing; an earlier version of this
description had the comparison wrong.

What remains true is that Redis under `SETEX` is the only backend here that
cannot express the requested lifetime at all.  It has to miss in one direction
or the other, and rounding down would make it the only adapter that can drop a
value before its TTL.

`WorkersKvStore` does suggest a third option this PR does not take: Redis could
keep the requested expiry alongside the value and filter on read, which would
give exact sub-second lifetimes without leaving `SETEX`.  That changes the
stored value format for every key, which is more than #1028 asks for; noting it
so the choice is on the record rather than implied.

**The costs are asymmetric.**  Expiring late costs at most one extra second of
staleness on a cache.  Expiring early can change behaviour: a TTL used to
suppress duplicate work starts letting duplicates through.  Fedify does exactly
that — the inbox listener and the queue handler both mark an activity as
processed with `kv.set(cacheKey, true, { ttl: 1 day })`.

**It matches the issue's expected behaviour.**  "A sub-second TTL is expected to
be normalized to the smallest expiry Redis can express" is what rounding up
produces for the sub-second cases.

In practice nothing in the library on this branch is affected either way, since
every TTL Fedify itself passes is already a whole number of seconds.  The
reachable path is application code, which `KvStoreSetOptions.ttl` opens to any
`Temporal.Duration`, and which #1027 made easier to hit on newer branches by
letting an application choose `publicKeyTtl` and
`httpMessageSignaturesSpecTtl` directly.

Changes
-------

 -  *packages/redis/src/kv.ts*: add `expirySeconds()`, which returns
    `Math.max(1, Math.ceil(ttl.total("second")))`, and call it from `set()`.
    The `Math.max` matters only when `Math.ceil` yields zero — a zero or
    negative duration — and it documents why the rounding goes up.
 -  *packages/redis/src/kv.test.ts*: two regression tests, described below.
    The gated one bounds `PTTL` against the elapsed time rather than comparing
    `TTL` to an exact integer, so it does not depend on how fast the run is.
 -  *changes.d/redis/ttl-whole-seconds.md* and *CHANGES.md*.

`RedisKvStore` has no `cas()` method, so `set()` is the only call site.

Verification
------------

**The regression tests fail without the patch.**  Reverting `set()` to pass the
raw total fails both of them; the end-to-end one reports the error from the
issue verbatim:

~~~~
ReplyError: ERR value is not an integer or out of range
~~~~

**Each half of the expression is pinned by its own assertion.**

| mutation | test that fails |
| --- | --- |
| `Math.max(1, Math.ceil(…))` → `ttl.total("second")` | both; the gated one with the `ReplyError` above |
| `Math.ceil` → `Math.round` | `1.4s rounds up, not to the nearest second` — `1 !== 2` |
| drop the `Math.max(1, …)` floor | `a zero TTL stays at least 1` |
| clamp negatives to `0`, keep zero at `1` | `a negative TTL stays at least 1` |

The `Math.round` mutation is the one that matters for review: it still produces
a whole positive number for every case in the issue's table, and only the 1.4-
second case separates it from `Math.ceil`.  The last row is why the negative
cases are not redundant with the zero one — an implementation that clamped zero
to one second and let negatives through would have passed the table as it
originally stood, and reached `SETEX` with an expiry it rejects.

**One test needs no Redis server.**  `RedisKvStore.set() rounds a TTL up to
whole seconds` passes the store a stand-in client that records what `set()`
hands to `SETEX`, and checks the exact integer for 1s, 5min, 1.5s, 1.4s, 500ms,
1ms, 0, −1ms, −30s and −1h, asserting each result is a whole positive number.
`RedisKvStore.set() stores a sub-second TTL` is `REDIS_URL`-gated and covers the
end-to-end behaviour, including a −30s TTL and that a 30-second TTL is stored
unchanged.

**The gated test does not depend on how fast the run is.**  It used to compare
`TTL` against an exact integer.  `TTL` rounds the remaining lifetime to whole
seconds, so a delay of just over half a second reports `0` for a key written
with `SETEX 1` and `29` for one written with `SETEX 30` — a failure on a slow
run with a correct implementation.  Measured against Redis 7.4.11, 600 ms after
a `SETEX 1` write:

| | `EXISTS` | `TTL` | `PTTL` |
| --- | --- | --- | --- |
| `SETEX 1`, +600 ms | 1 | **0** | 387 |
| `SETEX 30`, +600 ms | 1 | **29** | 29392 |

It now reads `PTTL` and bounds it on both sides.  The upper bound — the
requested lifetime in milliseconds — rules out a longer expiry than intended.
The lower bound subtracts the time the write and the read took, which is an
upper bound on the key's age, so a correct implementation cannot cross it
however slow the run.  Reading the expiry straight after the write leaves only
the two Redis commands inside that window.  Checked with an injected delay: at
600 ms and 900 ms the new bounds hold on both keys while the old `TTL`
assertions fail on all four.

**Both new tests import `test` from `node:test` directly rather than from
`@fedify/fixture`,** following `packages/postgres/src/kv.test.ts`.  An earlier
version of this description justified that at length by a gap in the fixture,
which swallowed a `require("node:test")` failure in ES modules and so reported
Node.js passes for tests that never ran.  #1039 fixed it and is in this
branch's base as of the rebase below, so that justification is gone; the
postgres precedent is the whole of the reason now.  The tests run on all three
runtimes either way.

**Branch and changelog.**  Rebased onto the current *2.0-maintenance*
(1a2b510f), which now carries #1039.  2.0.27 was released while this PR was
open, so the entry sits under 2.0.28; `sacho sync` produces no change and
`sacho check` passes.

**Tests.**  `@fedify/redis` against Redis 7.4.11 in Docker: Deno 14 passed / 0
failed, Bun 14 passed / 0 failed, Node.js 14 passed / 0 failed.  The Node.js
count was 7 before the rebase, because the nine pre-existing fixture tests in
this package were among those #1039 brought back to life; nothing here was
skipped.  `mise run check` passes, `sacho check` included.

AI use
------

Claude Opus 5 assisted with the patch, the tests, and running the checks above;
the commit carries an `Assisted-by` trailer.  Every number in this description
was measured against a real Redis instance rather than inferred from reading
the code.

Checklist
---------

    feature)? — not a new feature; the rounding is documented on
    `expirySeconds()`.
    fix)?
 -  [x] Did you run `mise test` on your machine? — yes, for `@fedify/redis` on
    all three runtimes; see *Verification*.

Additional notes
----------------

 -  **A zero or negative TTL now stores the value for one second.**  Both used
    to fail with `ERR invalid expire time in 'setex' command`, so this is
    strictly better than the current behaviour, but it is not what the other
    adapters do: they read a non-positive TTL as already expired.  This is a
    policy choice rather than a consequence of the rounding, and it is now
    documented as one on `set()` and on the helper.  An alternative would be to
    delete the key instead of storing it for one second, which matches how the
    other adapters read a non-positive TTL.  I kept the one-second floor because
    it is the smallest expiry `SETEX` can express and because deleting on
    `set()` is a larger behavioural change than this fix needs; happy to switch
    if you prefer the other reading.
 -  **Reframed: the one-second granularity is `SETEX`'s, not Redis's.**  If you
    would prefer exact sub-second expiry, switching `set()` to `SET … PX` would
    remove the rounding — and with it the zero and negative edge cases —
    entirely.  I kept `SETEX` here to stay minimal on a maintenance branch and
    because #1028 asked for normalisation rather than a new expiry path, and
    you have said keeping it is fine; noting the option so the choice is on the
    record rather than implied.
 -  **A TTL can now be up to one second longer than requested.**  For the cache
    lifetimes a `KvStore` holds, that is an extra second of staleness at worst.
 -  **This does not change any existing write.**  Every TTL that worked before —
    every whole number of seconds, which is all Fedify itself uses — produces
    the same `SETEX` argument as before.


packages/redis/src/kv.test.ts
```diff
@@ -144,3 +149,106 @@ test("RedisKvStore.list() - empty prefix", { ignore }, async () => {
...
+      Temporal.Duration.from({ milliseconds: 500 }),
+      1,
+      "a sub-second TTL becomes the smallest expiry",
+    ],
+    [
+      Temporal.Duration.from({ milliseconds: 1 }),
+      1,
+      "a near-zero TTL stays at least 1",
+    ],
+    [Temporal.Duration.from({ seconds: 0 }), 1, "a zero TTL stays at least 1"],
```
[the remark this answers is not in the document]
dahlia: Please address this by adding a negative TTL case to the regression table if we keep the one-second floor for non-positive durations.

packages/redis/src/kv.ts
```diff
@@ -8,6 +8,23 @@ import type { Cluster, Redis, RedisKey } from "ioredis";
 import { Buffer } from "node:buffer";
 import { type Codec, JsonCodec } from "./codec.ts";
 
+/**
+ * Turns a TTL into the whole number of seconds Redis `SETEX` requires.
+ *
+ * Redis expiries have one-second granularity, so a duration that is not a
```
dahlia, line 14: Please qualify this as a `SETEX` limitation here, in the changelog, and in the PR description. Redis supports millisecond expiries through [`SET` with `PX`](https://redis.io/docs/latest/commands/set/); one second is the minimum for the command this adapter currently uses. Keeping `SETEX` for this fix is fine.

packages/redis/src/kv.ts
```diff
@@ -8,6 +8,23 @@ import type { Cluster, Redis, RedisKey } from "ioredis";
...
+ * than to the nearest second: every other {@link KvStore} implementation keeps
+ * a value for at least as long as it was asked to, and expiring early is the
+ * direction that can change behaviour rather than just cost a refetch — a TTL
+ * used to suppress duplicate work would start letting duplicates through.
+ *
+ * The result is clamped to 1, the smallest expiry `SETEX` accepts, so a
+ * sub-second duration stores the value instead of being rejected.
+ */
+function expirySeconds(ttl: Temporal.Duration): number {
+  return Math.max(1, Math.ceil(ttl.total("second")));
```
dahlia, line 25: Please document explicitly that zero and negative TTLs now store the value for one second. The issue suggests a positive-integer floor, but does not separately settle negative TTL semantics. The current behavior is a policy choice, and the rationale about preserving the requested lifetime does not explain these cases.

CHANGES.md
```diff
@@ -21,6 +21,19 @@ To be released.
 [#1014]: https://github.com/fedify-dev/fedify/issues/1014
 [#1032]: https://github.com/fedify-dev/fedify/issues/1032
 
+### @fedify/redis
```
dahlia, line 24: The [current lint job fails](https://github.com/fedify-dev/fedify/actions/runs/34939886671/job/104285928196) because `sacho check` reports that *CHANGES.md* is out of sync with the fragments. Please update the branch against `2.0-maintenance` and regenerate the changelog with `sacho sync`, then verify `sacho check` passes.

dahlia (commented):
The fix follows the approach suggested in #1028. Please address the inline comments on TTL semantics, documentation, and regression coverage, and resolve the changelog sync failure before merging.

packages/redis/src/kv.ts
heeoneie: Done — the helper now says the one-second granularity belongs to `SETEX` rather than to Redis, and points at `SET` with `PX` as the millisecond-capable alternative. The same qualification is in the changelog fragment and in the PR description.

Reframed: the one-second granularity is `SETEX`'s, not Redis's. If you'd prefer exact sub-second expiry, switching `set()` to `SET … PX` would remove the rounding — and with it the zero and negative edge cases — entirely. I kept `SETEX` here to stay minimal on a maintenance branch and because #1028 asked for normalisation rather than a new expiry path, and you've said keeping it is fine; I'm noting it so the choice is on the record rather than implied, and I'm happy to do the `PX` version instead if you'd rather have the precision.

packages/redis/src/kv.ts
heeoneie: Done. Both the helper and `set()` now state that a zero or negative duration stores the value for one second, and say plainly that this is a policy choice rather than a consequence of the rounding — the other `KvStore` implementations read a non-positive TTL as already expired, while this one keeps the value for the shortest lifetime `SETEX` can express, on the grounds that storing it briefly is closer to the caller's request than failing the write.

The note is on `set()` as well because the helper is not exported, so its TSDoc would not reach users. The changelog fragment says it too.

The alternative reading — deleting the key for a non-positive TTL, which is what "already expired" would mean here — is noted in the PR description rather than implemented. Say the word if you would rather have that.

packages/redis/src/kv.test.ts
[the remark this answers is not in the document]
heeoneie: Added. The regression table now carries −1ms, −30s and −1h alongside the existing zero and positive cases, and each row additionally asserts the result is a whole *positive* number rather than only a whole one. The `REDIS_URL`-gated end-to-end test writes with a −30s TTL and checks the key comes back with `TTL` 1.

They are not redundant with the zero case. An implementation that clamped zero to one second but let negatives through passes the table as it originally stood, and reaches `SETEX` with an expiry it rejects; with the new rows that mutation fails on `a negative TTL stays at least 1`, and the gated test fails with `ERR invalid expire time in 'setex' command`. The full mutation table is in the PR description.

CHANGES.md
heeoneie: Fixed. The branch is rebased onto the current *2.0-maintenance* (d863a12a) and the changelog regenerated with `sacho sync`; `sacho check` passes locally.

The cause was a stale base rather than the sync itself. The branch was cut at 9c8ec49c, and 2.0.27 has been released since — so the fragment's entry was still being materialised into a `2.0.27` unreleased section that no longer exists. After the rebase it lands under 2.0.28, with the `[#1034]` reference link that the previous run had left only in the fragment.

heeoneie:
Thanks — all four are in, in 9bff7898, on top of a rebase onto the current *2.0-maintenance*.

 -  **`SETEX`, not Redis.** The helper, the changelog fragment and the PR description now attribute the one-second granularity to `SETEX` and point at `SET` with `PX`. I have left a note on that thread offering the `PX` version if you would rather have exact sub-second expiry — it would remove the rounding and the non-positive edge cases entirely, but it is a different change from the normalisation #1028 asked for, so I did not take that on unilaterally.
 -  **Non-positive TTLs documented.** On `set()` as well as on the helper, since the helper is not exported, and stated as a policy choice rather than a side effect of the rounding.
 -  **Negative coverage.** −1ms, −30s and −1h in the regression table, a −30s case in the gated end-to-end test, plus a positive-integer assertion on every row. A mutation that clamps zero but not negatives now fails, which the table could not catch before.
 -  **Changelog.** The lint failure was a stale base: the branch was cut before 2.0.27 shipped, so the entry was being materialised into a version section that no longer exists. Rebased, `sacho sync` re-run, `sacho check` green.

`@fedify/redis` passes on all three runtimes against Redis 7.4.11 (Deno 14, Bun 14, Node.js 7 — the Node.js count is lower because fixture-registered tests do not execute there on this branch, which the PR description explains).

packages/redis/src/kv.test.ts
```diff
@@ -144,3 +149,138 @@ test("RedisKvStore.list() - empty prefix", { ignore }, async () => {
...
+    const { redis, keyPrefix, store, cleanup } = getRedis();
+    try {
+      // Before the fix this threw `ERR value is not an integer or out of
+      // range`.
+      await store.set(["foo", "sub"], "bar", {
+        ttl: Temporal.Duration.from({ milliseconds: 500 }),
+      });
+      assert.strictEqual(await store.get(["foo", "sub"]), "bar");
+      assert.strictEqual(
+        await redis.ttl(`${keyPrefix}foo::sub`),
```
dahlia: Please make the integration assertions tolerate elapsed time. Redis TTL rounds the remaining lifetime to seconds, so a delay of just over 500ms can make these checks return 0 instead of 1, or 29 instead of 30, even when the implementation is correct. I confirmed that a key set with `SETEX 1` still exists after 600ms while TTL returns 0. The unit test already checks the exact conversion; please use elapsed-time-aware `PTTL` bounds for the integration checks.

packages/redis/src/kv.ts
```diff
@@ -8,6 +8,34 @@ import type { Cluster, Redis, RedisKey } from "ioredis";
...
 
+/**
+ * Turns a TTL into the whole number of seconds Redis `SETEX` requires.
+ *
+ * The one-second granularity is `SETEX`'s, not Redis's: Redis can express a
+ * millisecond expiry through `SET` with `PX`, and `SETEX` is simply the
+ * command this adapter uses.  Within that command a duration which is not a
+ * whole number of seconds has to be approximated.
+ *
+ * It is rounded up rather than to the nearest second, because every other
```
dahlia, line 19: Please revise the Workers KV comparison in the PR description. `WorkersKvStore` clamps the physical storage TTL to 60 seconds, but also stores the requested expiry in metadata and returns `undefined` from `get()` once that expiry has passed. It therefore does not establish a precedent for extending the lifetime visible to callers. This is a correction to the rationale in the PR description; it does not require changing the Redis rounding policy.

dahlia (commented):
Thanks for the updates. Please address the two inline comments on test timing and the Workers KV comparison, and rebase onto the latest `2.0-maintenance`, which has been updated.

packages/redis/src/kv.test.ts
heeoneie: Fixed. Reproduced your measurement first — Redis 7.4.11, 600 ms after the write:

| | `EXISTS` | `TTL` | `PTTL` |
| --- | --- | --- | --- |
| `SETEX 1` | 1 | **0** | 387 |
| `SETEX 30` | 1 | **29** | 29392 |

so all three assertions were reachable, not just the one-second ones.

The gated test now reads `PTTL` and bounds it on both sides. The upper bound is the requested lifetime in milliseconds, which rules out a longer expiry than intended. The lower bound is that minus the time the write and the read took — that span is an upper bound on the key's age, so a correct implementation cannot cross it however slow the run. The expiry is read straight after the write, so only the two Redis commands sit inside the window.

Checked with an injected delay between the write and the read:

```
delay=   0ms SETEX 1  elapsed=  19 pttl=  1000 | new bounds: PASS | old TTL===1: PASS
delay= 600ms SETEX 1  elapsed= 602 pttl=   398 | new bounds: PASS | old TTL===1: FAIL (got 0)
delay= 900ms SETEX 1  elapsed= 903 pttl=    98 | new bounds: PASS | old TTL===1: FAIL (got 0)
delay= 600ms SETEX 30 elapsed= 603 pttl= 29397 | new bounds: PASS | old TTL===30: FAIL (got 29)
delay= 900ms SETEX 30 elapsed= 902 pttl= 29099 | new bounds: PASS | old TTL===30: FAIL (got 29)
```

Past 1000 ms the one-second key is genuinely gone and `PTTL` returns `-2`; that is inherent to a one-second TTL rather than something the bounds can absorb, and it is two Redis round trips away in the test.

The exact conversion stays where you said it belongs — the server-free unit test, untouched.

packages/redis/src/kv.ts
heeoneie: You're right, and I checked it in `packages/cfworkers/src/mod.ts` rather than taking it on trust. `set()` writes `metadata.expires = Date.now() + options.ttl.total("milliseconds")` next to the clamped `expirationTtl`, and `get()` returns `undefined` once that has passed — `list()` skips the entry on the same condition. So the extension is physical only and never visible to a caller. My claim that the adapter "extends rather than cutting short" was wrong.

The description now says so plainly, including the part I'd rather not have to write: no adapter here extends a caller-visible lifetime today, and this change makes `RedisKvStore` the first to do so. The table's Workers row records both halves — the 60-second physical clamp and the millisecond-precision check on read — so the 60 does not read as a precedent on its own.

The rationale that survives is the other two points, which are untouched: `SETEX` cannot express the requested lifetime at all, so it has to miss in one direction, and rounding down would make Redis the only adapter that can drop a value early; and the cost asymmetry, where a TTL used to suppress duplicate work starts letting duplicates through if it expires early. As you say, no change to the rounding policy.

One thing your correction surfaced, noted in the description rather than implemented: Workers' approach would work here too — keep the requested expiry alongside the value and filter on read — and would give exact sub-second lifetimes without leaving `SETEX`. It changes the stored value format for every key, so it is well beyond what #1028 asks for, but it is a real third option and I'd rather have it on the record than implied. Same footing as the `SET … PX` note above.

heeoneie:
Rebased onto `2.0-maintenance` at 1a2b510f; no conflicts, and `sacho sync` produces no change on the new base, so the 2.0.28 entry stands as it was. Both inline comments are addressed in their threads.

The rebase had one consequence worth flagging, since it lands on this PR rather than on the base. #1039 fixed the fixture's silent Node.js registration failure — which this PR's description had documented at length to justify importing `test` from `node:test` directly. That justification is now gone. Two things follow:

- The description's fixture section is cut down to the `packages/postgres/src/kv.test.ts` precedent, which is the whole of the reason now. I left the two new tests on `node:test` rather than moving them to the fixture: on a maintenance branch the smaller diff seemed right, and they run on all three runtimes either way. Happy to move them if you'd rather they went through the fixture now that it works.
- `@fedify/redis` on Node.js now reports **14 passed**, not 7 — the nine pre-existing fixture tests in this package were among those #1039 brought back to life. All three runtimes agree at 14 against Redis 7.4.11 in Docker, and `mise run check` passes.

dahlia (approved):
Thanks for working on this!

PR #1035, opened 2026-09-16
Lower suppressed vocabulary log level

### LINKED ISSUE #933: Lower suppressed vocabulary failures below error level
Summary
-------

Generated vocabulary accessors log parsing and fetch failures at error level even when callers pass `suppressError: true`. This turns expected rejection of malformed remote objects into application error-tracking events despite the accessor returning `null` as requested.

Observed behavior
-----------------

In *@fedify/vocab-tools/src/property.ts*, both the document-loader catch and JSON-LD parsing catch call `getLogger(["fedify", "vocab"]).error(...)` when `options.suppressError` is enabled, then return `null`.

Hackers' Pub calls `Announce.getObject({ suppressError: true })` for untrusted inbox input. A remote URL returned a document that was not an ActivityStreams `Object`, so the accessor correctly returned `null`, but its error log produced five Sentry events under `GRAPHQL-42` with `TypeError: Expected an object of any type of: https://www.w3.org/ns/activitystreams#Object`.

Expected behavior
-----------------

Failures intentionally absorbed by `suppressError: true` should be logged below error level, probably warning or debug. Callers that omit `suppressError` would continue to receive the thrown exception, so genuinely unhandled failures would remain visible to error tracking.

Production evidence
-------------------

The affected Sentry issue is https://hackerspub.sentry.io/issues/GRAPHQL-42.

Lumia1108:
Summary
-------

Vocabulary accessors previously logged fetch and parsing failures at the error level even when callers explicitly enabled `suppressError`.  The default document loader also emitted an error-level log for non-OK HTTP responses before the accessor caught the resulting exception, causing intentionally handled failures from untrusted remote objects to be reported as application errors.

This change passes the suppression state through the default and authenticated document loader paths.  Suppressed non-OK HTTP responses and accessor fetch and parsing failures are now logged at the warning level, while the document loader continues to throw for the accessor to handle and return `null`.  Calls without `suppressError`, or with it set to `false`, preserve the existing error-level logging and exception behavior.

Regression coverage verifies warning-level logs for suppressed HTTP fetch and parsing failures and confirms that unsuppressed calls still throw.

Closes #933

Testing
-------

 -  `mise run test:deno packages/vocab/src/vocab.test.ts --filter "Announce.getObject() logs suppressed failures as warnings"`—1 passed
 -  `mise run check-each vocab-runtime vocab-tools vocab fedify`
 -  `mise run test-each vocab-runtime vocab-tools vocab fedify`—427 passed, 1 skipped, 0 failed
 -  `sacho check`
 -  `mise run check:md`

AI assistance
-------------

Codex (GPT-5.6) was used to understand the issue, maintainer feedback, and contribution workflow; examine the document loader and generated accessor paths; discuss the appropriate log levels and suppression propagation; review the implementation and generated changes; and draft the regression test;

I personally wrote the source implementation and changelog changes.  Codex drafted the regression test, which I reviewed and independently ran.  I also regenerated and reviewed the Deno, Node.js, and Bun snapshots, verified the materialized `CHANGES.md`, and independently ran all checks listed above.  I confirmed that suppressed HTTP fetch and parsing failures are logged at warning level, that suppressed accessors return `null`, and that unsuppressed calls retain their existing error logging and exception behavior.


packages/vocab-tools/src/property.ts
```diff
@@ -97,7 +97,7 @@ async function* generateProperty(
           });
           span.end();
           if (options.suppressError) {
-            getLogger(["fedify", "vocab"]).error(
+            getLogger(["fedify", "vocab"]).debug(
               "Failed to fetch {url}: {error}",
               { error, url: lookupUrl }
             );
```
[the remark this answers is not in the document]
Lumia1108: Thanks for pointing this out. I confirmed that the default document loader logs a non-OK response at the error level before throwing the FetchError that the generated accessor suppresses. (`packages/vocab-runtime/src/docloader.ts`: line 188, [source](https://github.com/fedify-dev/fedify/blob/5c417eb8/packages/vocab-runtime/src/docloader.ts#L188-L202))

Addressing this would require propagating suppression through `DocumentLoaderOptions` and `getRemoteDocument()`, which expands the change beyond the two generated accessor catches identified in #933. Would you prefer that propagation to be included in this PR, or should this PR remain scoped to the accessor-level logs described in the issue?

packages/vocab-tools/src/property.ts
```diff
@@ -97,7 +97,7 @@ async function* generateProperty(
           });
           span.end();
           if (options.suppressError) {
-            getLogger(["fedify", "vocab"]).error(
+            getLogger(["fedify", "vocab"]).debug(
```
dahlia, line 100: Please use `.warn()` in both catches, as proposed in [the issue discussion]. These suppressed failures should remain visible for diagnostics without being treated as application errors. Using `debug` hides them when an application collects only warning-level logs and above. Please update the three runtime snapshots and the changelog to match.

[the issue discussion]: https://github.com/fedify-dev/fedify/issues/933#issuecomment-4942198180

packages/vocab-tools/src/property.ts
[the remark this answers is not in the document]
dahlia: Please pass the suppression state through to the document loader in this PR. With a mocked HTTP 404 response, `Announce.getObject({ suppressError: true })` returns `null`, but the default document loader still emits an error-level log before the accessor catches the exception. This leaves the error-reporting problem unresolved for HTTP failures.

Pass the suppression state through the loader path so these failures log at warning level while still throwing for the accessor to handle. Preserve the existing error logging and exception behavior when `suppressError` is absent or `false`. Please add regression coverage that captures logs for suppressed fetch and parsing failures and checks that unsuppressed calls still throw.

changes.d/vocab-tools/suppressed-error-log-level.md
```diff
@@ -0,0 +1,3 @@
+ -  Changed suppressed vocabulary fetch and parsing failures to log at the
+    debug level so that intentionally handled failures are not reported as
+    application errors.  [[#933] by Lumia1108]
```
dahlia, line 3: Please add the PR reference and use your romanized full name for the credit, as listed on [your CV]:

~~~~ markdown
[[#933], [#1035] by Jae Hui Hong]
~~~~

After updating the fragment, run `sacho resolve-links` to pin the references, ensuring #1035 links to the pull request URL. Please keep fragment changes limited to *changes.d/vocab-tools/suppressed-error-log-level.md*: the installed command has no per-file option, so check its diff and exclude any changes it makes to other fragments. Regenerate the materialized *CHANGES.md* from the fragment as well.

[your CV]: https://cuddlyblog.com/resume

dahlia (changes requested):
Sorry for the delay in reviewing this. Please address the comments and rebase your branch onto the latest upstream `main` to resolve the conflicts. Please use rebase rather than merging `main` into your branch (cross-merge).

Lumia1108:
Thanks for the review. I addressed the requested changes in `7adf1402`.

I rebased the branch onto the latest upstream `main`, changed both generated accessor catches to warning level, and passed `suppressError` through the default and authenticated document loader paths. Suppressed non-OK HTTP responses now log at warning level while the loader still throws for the accessor to handle. Unsuppressed calls retain their existing error-level logging and exception behavior.

I also added regression coverage for suppressed fetch and parsing failures and unsuppressed throws, regenerated all three runtime snapshots, and updated the changelog fragment and materialized `CHANGES.md` with the requested references and attribution.

I personally ran and verified:

 -  `mise run test:deno packages/vocab/src/vocab.test.ts --filter "Announce.getObject() logs suppressed failures as warnings"`—1 passed
 -  `mise run check-each vocab-runtime vocab-tools vocab fedify`
 -  `mise run test-each vocab-runtime vocab-tools vocab fedify`—427 passed, 1 skipped, 0 failed
 -  `sacho check`
 -  `mise run check:md`

packages/vocab-runtime/src/docloader.ts
```diff
@@ -182,18 +190,31 @@ export async function getRemoteDocument(
...
   if (!response.ok) {
-    logger.error(
-      "Failed to fetch document: {status} {url} {headers}",
-      {
-        status: response.status,
-        url: documentUrl,
-        headers: Object.fromEntries(response.headers.entries()),
-      },
-    );
+    if (options?.suppressError) {
```
dahlia, line 198: Please apply `suppressError` to the redirect-limit and redirect-loop branches in `getDocumentLoader()` as well. Both still call `logger.error()` unconditionally. I reproduced both cases: `Announce.getObject({ suppressError: true })` returns `null`, but an error-level log is emitted first. Use warning-level logging when suppression is enabled, preserve the existing throws and unsuppressed logging, and add regression coverage for both paths.

packages/vocab-tools/src/property.ts
```diff
@@ -90,15 +90,17 @@ async function* generateProperty(
         const lookupUrl = formatIri(url);
         let fetchResult: RemoteDocument;
         try {
-          fetchResult = await documentLoader(lookupUrl);
+          fetchResult = await documentLoader(lookupUrl, {
+            suppressError: options.suppressError,
```
dahlia, line 94: The suppression flag currently applies only to the initial document fetch. It is not passed to the `contextLoader` used by the JSON-LD parser. I reproduced this with a valid Note whose external `@context` returns `404 Not Found`: `Announce.getObject({ suppressError: true })` returns `null`, but the context loader emits an error-level log before the accessor logs the parsing failure as a warning. Please propagate suppression to context loads during parsing, preserve unsuppressed behavior, and add a regression test for this case.

packages/vocab-runtime/src/docloader.ts
```diff
@@ -46,6 +46,13 @@ export interface DocumentLoaderOptions {
    * @since 1.8.0
    */
   signal?: AbortSignal;
+
+  /**
+   * Whether to lower error-level logs for recoverable document loading
+   * failures to warning-level logs.  The loader still throws the error.
+   * @default `false`
+   */
+  suppressError?: boolean;
```
dahlia, line 55: Please also honor `suppressError` when URL validation rejects a private address, in both the default and authenticated document loaders. I tested `http://127.0.0.1/private` with each loader: the accessor returns `null` with `suppressError: true`, but `Disallowed private URL: {url}` is still logged at error level. No network request is made. Please retain the address checks and thrown exceptions, lower this log to warning only for suppressed calls, and add regression coverage for both loaders.

dahlia (changes requested):
Thanks for addressing the earlier feedback and rebasing. Three paths still emit error-level logs with `suppressError: true`: redirect failures, external JSON-LD context loading failures, and private-address rejection. Please address the inline comments and add regression tests for these paths while preserving the existing exceptions and address checks.

Issue #974, opened 2026-07-31
@fedify/lint: actor-id-required/actor-inbox-property-required/actor-shared-inbox-property-required false positive when dispatcher returns null


Jae-Hyuk-Jang:
Environment
-----------

 -  Package/tool: @fedify/lint (oxlint plugin)
 -  Runtime: Node.js
 -  Operating system: Linux (WSL2)

Steps to reproduce
------------------

Register an actor dispatcher that returns `null` for "not found" (as documented in `ActorDispatcher`'s JSDoc) before returning a `Person` with `id`, `inbox`, and `endpoints.sharedInbox` all set:

~~~~ ts
federation.setActorDispatcher("/users/{identifier}", async (ctx, identifier) => {
  const user = /* lookup */;
  if (user == null) return null;

  return new Person({
    id: ctx.getActorUri(identifier),
    inbox: ctx.getInboxUri(identifier),
    endpoints: new Endpoints({ sharedInbox: ctx.getInboxUri() }),
  });
});
~~~~

Run `oxlint` on the file.

Expected behavior
-----------------

No warning/error. The `Person`-returning branch already sets `id`, `inbox`, and `endpoints.sharedInbox` as recommended, and returning `null` for "actor not found" is explicitly documented as valid:

 -  `ActorDispatcher`'s return type includes `| null`  (packages/fedify/src/federation/callback.ts).
 -  `setActorDispatcher`'s JSDoc says the dispatcher "may return an actor, a `Tombstone`, or `null` if the actor is not found" (packages/fedify/src/federation/federation.ts).
 -  The handler treats a `null` return as a normal 404, not an error (packages/fedify/src/federation/handler.ts).

Actual behavior
---------------

`actor-id-required` (error), `actor-inbox-property-required`, and `actor-shared-inbox-property-required` all fire on the dispatcher, even
though the `Person`-returning path is fully compliant.

Cause
-----

In `packages/lint/src/lib/property-checker.ts`, `checkAllReturnPaths` requires *every* `return` statement in the dispatcher body to satisfy the property checker:

~~~~ ts
const checkAllReturnPaths = (propertyChecker: PropertyChecker) =>
(node) =>
  pipe(
    node,
    collectReturnPaths,
    cases(isEmpty, always(false), every(checkReturnStatement(propertyChecker))),
  );
~~~~

`checkReturnStatement` treats a non-object `argument` (such as `null`) as an automatic failure via `always(false)`:

~~~~ ts
[...]

PR #1040, opened 2026-09-19
Make outbox-listener-delivery-required path-aware

### LINKED ISSUE #900: Make the outbox delivery lint rule path-aware
## Why

The existing outbox listener rule in *packages/lint/src/rules/outbox-listener-delivery-required.ts* catches a useful class of delivery mistakes, but it currently decides success by finding a `ctx.sendActivity()` or `ctx.forwardActivity()`-style call somewhere in the listener source. That keeps the first implementation simple, but it can treat unreachable or unrelated calls as proof that every posted activity is delivered.

This matters because a local outbox handler that accepts a post without actually delivering it creates a confusing ActivityPub failure: the object exists locally, but followers never receive it.

## Current code

The rule already handles direct calls, aliases, destructuring, bracket notation, template literals, and named listener callbacks. The related tests live in *packages/lint/src/tests/outbox-listener-delivery-required.test.ts*.

The next step is not to rewrite the rule from scratch. It is to make the delivery check understand enough control flow to avoid obvious false negatives.

## Scope

Please update the rule so that it does not count delivery calls that are clearly outside the listener delivery path. Good starting cases are:

- `ctx.sendActivity()` inside a nested helper function that is declared but never called.
- `ctx.forwardActivity()` behind `if (false)` or after an unconditional `return`.
- ~~A delivery call that only exists in a callback passed to an unrelated API, such as `array.map(() => ctx.sendActivity(...))`, when the callback is not awaited or returned.~~ **Withdrawn.** This was written on the assumption that a callback whose result is dropped never runs. It does: `map()` invokes its callback for every element, so the delivery call executes and the promise is simply discarded. That is worth catching, but it is a different defect from a delivery call that never runs, and it needs a message that says so. It is now tracked in #1057.

The implementation can be conservative. When the rule cannot show that a delivery call fails to run, it should stay quiet: a missed warning is safer than a warning on a listener that delivers.

## Non-goals

Do not try to build a full TypeScript control-flow analyzer. Do not require type information. This rule should keep working in both the Deno lint plugin and the ESLint plugin surfaces.

## Suggested checks

Add focused tests in *packages/lint/src/tests/outbox-listener-delivery-required.test.ts*. Existing positive cases should keep passing, especially the alias and named-listener cases.

Jae-Hyuk-Jang:
Closes #900

## Background

`outbox-listener-delivery-required` (`@fedify/lint`) decided whether a listener delivers a posted activity by scanning the whole listener source as text for a `ctx.sendActivity()`/`forwardActivity()` call, without checking whether that code actually runs.

## Changes

- Add a reachability scan that follows the listener's own control flow (if/else, try/catch, switch, loops), skips statically-dead branches and code after an unconditional `return`/`throw`, and does not descend into a nested function's body unless that function is itself called from reachable code.
- Keep the existing text-based pattern matching for the delivery call itself (aliases, destructuring, bracket/template notation), now run only over this narrowed, reachable text.
- Add regression tests for the three false-negative cases from the issue (unused nested helper, dead branch, delivery call only inside an unrelated callback), plus a case after an unconditional `return`, and a positive case for a helper function that is actually called.

## Testing

- `mise run check-each lint`
- `mise run test-each lint` (Deno and Node.js, 515/515 pass)
- Reverted the fix locally and reran the suite: exactly the 4 new regression tests fail, confirming they demonstrate the bug.

## AI disclosure

This was implemented with Claude Code (`claude-sonnet-5`): I described the issue and reviewed the design and results at each step, and Claude Code designed the reachability-based rewrite, implemented it, found and fixed a bug during testing (Deno.lint exposes an AST node's children through prototype getters rather than own enumerable properties, so `Object.entries()` couldn't see them — `for...in` was needed instead), and verified the tests pass in both Deno and Node.js.


dahlia (changes requested):
Thanks for taking this on. Could you retarget this to `main` rather than `2.2-maintenance`?

I see why you chose `2.2-maintenance`: `outbox-listener-delivery-required` landed in 2.2.0, so that is the oldest maintenance branch carrying the rule. I don't think this is a bug fix, though.

The change makes the rule report more errors than before. Someone upgrading from 2.2.12 to 2.2.13 would get lint failures in code that passed before, without having changed anything on their side. That's a lot to spring on a patch release. The one lint change we've shipped to maintenance branches so far, #974, went the other way: it removed false positives, so the worst an upgrader saw was some warnings disappearing. ESLint's semver policy draws the same line, where a rule fix that produces more errors is a minor release.

Nothing in #900 marks it as a bug either. It calls the original implementation deliberately simple and frames this as the next step rather than reporting a defect, and the diff adds 187 lines of new reachability analysis, which fits that description.

So `main`, where the unreleased version is 2.4.0.

Two small things while you're rebasing. Please drop the *CHANGES.md* edit and keep only the fragment: fragments under *changes.d/* are materialized into *CHANGES.md* at release time, as configured in *sacho.toml*, so the hand-written section would end up duplicated.

In the fragment itself, could you describe the cases the rule now catches instead of calling this “Fixed”? That would tell users more clearly why they may start seeing new reports.

The rule work itself looks good, and the reverted-fix check in your testing notes was helpful. I'll do a full code review once it's on `main`.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -212,6 +212,189 @@ function getDeliveryAliasName(node: Node): string | null {
...
+  sourceCode: { getText(node: unknown): string },
+  node: Node,
+  helpersOut: Map<string, FunctionLikeNode>,
+): string {
+  let text = sourceCode.getText(node);
+  const nested: Array<{ name: string | null; fn: FunctionLikeNode }> = [];
+  collectNestedFunctions(node, nested);
+  for (const { name, fn } of nested) {
+    const fnText = sourceCode.getText(fn);
+    text = text.split(fnText).join("()=>{}");
```
dahlia, line 358: This is the common cause behind most of my other comments, so I'll start here.

`maskNestedFunctions()` replaces every nested function's source text with `()=>{}`, and only a function called by a local name is pulled back in.  An anonymous callback is therefore always treated as unreachable, even when its result is awaited immediately.  All three of these pass on `main` and are reported on this branch:

~~~~ typescript
await Promise.all(recipients.map((inbox) => ctx.sendActivity(sender, inbox, activity)));
return Promise.all(recipients.map((inbox) => ctx.sendActivity(sender, inbox, activity)));
await (async () => { await ctx.sendActivity(sender, inbox, activity); })();
~~~~

#900 scoped the `array.map()` case to “when the callback is not awaited or returned”.  The implementation cannot honor that qualifier, because it has no way to tell `recipients.map(cb)` with a discarded result from `await Promise.all(recipients.map(cb))`.  The changelog fragment's “a callback passed to an unrelated function” describes behavior the rule does not actually have, for the same reason.

The deeper problem is deciding reachability by masking source text and re-matching names with regexes.  That representation cannot express whether a callback's result is awaited or returned, whether an identifier is a call or a reference, or which scope a name resolves in.  My next three comments are all consequences of it, and patching them one at a time will keep producing this class of bug.

I'd suggest moving the reachability decision onto the AST: resolve a call expression's callee to a function node, and ask whether that node is invoked from a reachable position.  Textual pattern matching is worth keeping for finding the delivery method name itself, but not for control flow.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -212,6 +212,189 @@ function getDeliveryAliasName(node: Node): string | null {
...
+  collectReachableStatements(root, statements);
+
+  const helpers = new Map<string, FunctionLikeNode>();
+  const code = statements
+    .map((statement) => maskNestedFunctions(sourceCode, statement, helpers))
+    .join("\n");
+
+  const helperCode = Array.from(helpers)
+    .filter(([name]) =>
+      new RegExp(String.raw`\b${escapeRegExp(name)}\s*\(`).test(code)
```
dahlia, line 388: Whether a helper is really called is decided by a `\b<name>\s*\(` regex, but `code` has not been through `stripCommentsAndStrings()` yet, and the pattern cannot separate a call from a reference.  It goes wrong in both directions.

False positives, both clean on `main`:

~~~~ typescript
const deliver = (inbox) => ctx.sendActivity(sender, inbox, activity);
inboxes.forEach(deliver);  // passed by reference, so `deliver(` never appears
~~~~

~~~~ typescript
const $deliver = async () => { await ctx.sendActivity(sender, inbox, activity); };
await $deliver();  // `\b` cannot match before `$`
~~~~

The `$` case already has an answer in this file: `boundedName` in `buildContextExpressionPattern()` uses `(?<![\w$])…(?![\w$])` for exactly this reason, and matching it here would fix it.

False negatives:

~~~~ typescript
console.log(ctx.identifier /* call deliver() later */);  // comment read as a call
await someService.deliver();  // `.` is a word boundary, so the local `deliver` body is pulled in
~~~~

A comment on its own line is harmless, since it belongs to no statement's `getText()`.  Only a comment inside a statement triggers this.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -212,6 +212,189 @@ function getDeliveryAliasName(node: Node): string | null {
...
+    )
+    .map(([, fn]) =>
+      collectDeliveryScanCode(sourceCode, fn.body as Node, visited)
```
dahlia, line 391: The recursion descends only into `fn.body`, and `helpers` is collected per scope, so from inside one helper its siblings in the same scope are invisible.

~~~~ typescript
async function outer() { await inner(); }
async function inner() { await ctx.sendActivity(sender, inbox, activity); }
await outer();
~~~~

This passes on `main` and is reported here.  Nesting `inner` inside `outer` works, so the rule breaks at the moment someone extracts a second helper alongside the first, which is a hard failure to diagnose from the message alone.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -212,6 +212,189 @@ function getDeliveryAliasName(node: Node): string | null {
...
+    if (decl.init != null) collectNestedFunctions(decl.init, out, hint);
+    return;
+  }
+
+  // Deno.lint's AST nodes expose their children through prototype
+  // getters, not own enumerable properties, so `Object.entries()` sees
+  // none of them. `for...in` walks the prototype chain and works for
+  // both Deno.lint's and ESTree's node shapes.
+  const record = node as unknown as Record<string, unknown>;
+  for (const key in record) {
```
dahlia, line 337: The `VariableDeclarator` branch forwards `nameHint` to `decl.init`, but this generic walk recurses with `collectNestedFunctions(record[key], out)` and drops it.  A function reached through an object literal therefore gets `name: null`, is masked, and is never registered as a callable helper.

~~~~ typescript
const handlers = {
  deliver: () => ctx.sendActivity(sender, inbox, activity),
};
await handlers.deliver();
~~~~

Clean on `main`, reported here.  Method shorthand (`{ async deliver() {} }`) and methods of a class declared inside the listener behave the same way.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -212,6 +212,189 @@ function getDeliveryAliasName(node: Node): string | null {
...
+function collectReachableStatements(node: Node, out: Node[]): void {
+  switch (node.type) {
+    case "BlockStatement":
+      for (const statement of node.body) {
+        collectReachableStatements(statement as Node, out);
+        if (isUnconditionalExit(statement as Node)) return;
+      }
+      return;
+
+    case "IfStatement": {
```
dahlia, line 249: Lower severity, but worth recording while the approach is being reconsidered.

Control-flow head expressions are never collected.  `IfStatement` recurses into `consequent` and `alternate` but skips `test`; `SwitchStatement` walks only the case consequents, ignoring `discriminant` and each case's `test`; the loop branches walk only `body`, ignoring `init`, `test`, `update` and `right`.  A delivery call or a helper call in any of those positions is invisible.  Given `sendActivity()`'s return type a delivery call there is unrealistic, but a helper call in a loop or `if` head is not.

In the other direction, `if (false)` is treated as dead while the bodies of `while (false)` and `do…while (false)` still count.  That is a false negative rather than a false positive, so it matches the old behavior and sits outside what #900 asked for, but it makes the dead-branch handling inconsistent.

changes.d/lint/outbox-listener-path-aware.md
```diff
@@ -0,0 +1,15 @@
...
+links:
+  '#1040': https://github.com/fedify-dev/fedify/pull/1040
+  '#900': https://github.com/fedify-dev/fedify/issues/900
+---
+ -  Changed `outbox-listener-delivery-required` (`@fedify/lint`) to check
+    whether its `ctx.sendActivity()`/`forwardActivity()` call is reachable,
+    rather than merely present somewhere in the source.  It now reports a
+    listener whose only delivery call sits in an unused nested helper
+    function, behind a statically-dead branch (such as `if (false)` or code
+    after an unconditional `return`), or inside a callback passed to an
```
dahlia, line 11: Two things here, and the first one is my mistake.

Reverting the *CHANGES.md* edit was bad advice from me, sorry.  *sacho.toml* sets `materialize = true`, so the unreleased region of *CHANGES.md* has to agree with the fragments, and as it stands `sacho check` fails with “materialized changelog is out of sync with fragments”.  That command is part of the `check` task, which CI runs, so this would fail the build.  Please run `sacho sync` and commit the resulting *CHANGES.md* alongside the fragment.  Your original PR was right to carry both.

The backslash in `by Jae-Hyuk-Jang\]` was also my error.  Sacho generates it, and the same form is already in the released entries, including your #1022 one.  Nothing to do there.

Separately, “inside a callback passed to an unrelated function” does not match what the rule does: it also reports callbacks whose result is awaited.  Worth rewording once the implementation settles.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -212,6 +212,189 @@ function getDeliveryAliasName(node: Node): string | null {
...
+ * loops) but never descends into a nested function's body, and skips dead
+ * code (a statically-falsy `if` branch, or anything after an unconditional
+ * `return`/`throw`).
+ */
+function collectReachableStatements(node: Node, out: Node[]): void {
+  switch (node.type) {
+    case "BlockStatement":
+      for (const statement of node.body) {
+        collectReachableStatements(statement as Node, out);
+        if (isUnconditionalExit(statement as Node)) return;
```
dahlia, line 245: CodeRabbit flagged this too, and it is right.

`isUnconditionalExit()` is consulted only for a direct child of the block, so an exit nested one level down does not stop the enclosing statement list.  All of these leave the unreachable delivery call in the scan source, and the rule stays silent:

~~~~ typescript
if (true) return;
await ctx.sendActivity(sender, inbox, activity);
~~~~

~~~~ typescript
if (activity.id == null) { return; } else { return; }
await ctx.sendActivity(sender, inbox, activity);
~~~~

~~~~ typescript
switch (kind) {
  case "x":
    return;
    await ctx.sendActivity(sender, inbox, activity);  // same case, after the return
}
~~~~

`if (true) throw new Error()` behaves like the first one.

This sits inside what the PR already claims to handle.  There is a passing test for a delivery call after a direct `return`, so only the nested case leaks, which makes the gap easy to miss.  Returning a completion status from `collectReachableStatements()` and propagating it through `if`, `switch` and `try` is the shape CodeRabbit suggests, and it carries over to whichever approach you settle on.

Its other two comments, on the nested-function masking and on the helper-name regex, overlap with mine.  Both are valid, so please fold them into the same rework rather than treating them separately.

dahlia (changes requested):
Thanks for the rebase, and for the reverted-fix check in your testing notes.  That is the kind of verification I wish more pull requests came with.

I went through the rule closely, and I think the approach needs to be reconsidered before this can land.  The inline comments have the reproductions, but they share one cause, so I would rather state it here than leave it spread across seven threads.

Reachability is decided by taking the listener's source text, blanking out nested functions, and re-matching names with regexes.  That representation cannot carry the distinctions the rule has to make: whether a callback's result is awaited or returned, whether an identifier is a call or a reference, which scope a name resolves in, or whether a nested exit ends the enclosing statement list.  Each gap shows up as a concrete misreport.  `await Promise.all(recipients.map((inbox) => ctx.sendActivity(…)))` is flagged although it delivers; so are `inboxes.forEach(deliver)`, a helper held in an object literal, and a helper that calls a sibling helper.  In the other direction, `if (true) return;` before a delivery call is accepted silently.  I checked each case against `main`, and every false positive is clean there, so these are regressions this PR would introduce.

That is the part I want to be clear about: a lint rule that reports correct code is worse than one that misses cases.  Someone whose code is already right cannot act on the report, so they reach for a disable comment, and the rule stops being worth having.  That is why I would rather not patch these one at a time.

What I would suggest instead is to decide reachability on the AST.  Keep collecting reachable statements as you already do, but when you reach a call expression, resolve its callee to a function node rather than matching a name as text.  A local helper becomes a binding you can follow; an immediately invoked function expression becomes a callee that happens to be a function literal; a callback is reachable when the call it is passed to is reachable and its result is awaited or returned.  Textual matching is still the right tool for finding `sendActivity` and `forwardActivity` and their aliased forms, but structural questions want structural answers.  ESLint's scope analysis gives you variable resolution directly, and Deno's lint plugin API exposes enough of the AST for the same walk, so none of this needs type information, which keeps it inside #900's non-goals.

That amounts to a rewrite of what is here, and the history would not help anyone reading it later.  If starting from a clean branch is easier, please do: close this one, open a new pull request, and I will keep #900 assigned to you.  There is no expectation that you salvage the current commits.

Two things to carry over either way.

The rule's section in *docs/manual/lint.md* still says the rule fires when “the listener body never calls either delivery method”, with nothing about reachability, dead branches, or how nested helpers and callbacks are treated.  This is a user-visible behavior change, and the feature checklist in *CONTRIBUTING.md* asks for documentation to move with it.  The changelog fragment needs the same attention; I left a note on it about wording that does not match the implementation.

Also, please disregard my earlier advice to drop the *CHANGES.md* edit.  That was wrong of me and it broke `sacho check`.  The details are in the comment on the fragment, but the short version is that your original PR had it right.

The reachability idea itself is sound and worth having in the rule.  Your tests are well organized and the four regression cases you picked are the right ones.  It is the mechanism underneath them that I would like to see change.

Jae-Hyuk-Jang:
Closing this per your suggestion, and opening a fresh pull request against `main` with the AST-based reachability approach. Keeping #900 in mind — thanks for the detailed review.

PR #1047, opened 2026-09-22
Fix Relay tests discard inbox responses, hiding `500 Internal Server Error`s

### LINKED ISSUE #1046: Relay tests discard inbox responses, hiding at least one `500 Internal Server Error`
Several tests in *packages/relay* approve a `Follow`, throw away the `Response` that `relay.fetch()` returned, and assert only that the follower was written to the key–value store.  I confirmed that `handles public Follow activity` in *packages/relay/src/litepub.test.ts* receives a `500 Internal Server Error` while its store assertions still pass.  I printed the status from inside the test:

~~~~
PROBE public-follow status: 500
~~~~

Reading the code from there: the follower's inbox is `https://remote.example.com/users/alice/inbox`, and `remote.example.com` does not resolve.  After the change for [GHSA-f59r-8gcj-68f2], outbound delivery validates its destination before calling `fetch()`, so `validatePublicUrl()` rejects that address and the `UrlError` escapes the inbox listener before the `Accept` is sent.  The follower record is written earlier in the same path, which is why the store assertions hold.

What those assertions show is that the follower was persisted.  They do not show that the `Accept` went out, or that the inbox request returned `202 Accepted`.  A regression that broke every outgoing `Accept` would leave them green.

I only measured that one test.  Two counts bound the rest.  Thirteen calls to `relay.fetch(request)` in *litepub.test.ts* and *mastodon.test.ts* discard the returned `Response`.  Seventeen tests across the two files use alice's inbox on *remote.example.com*, though some of them only read from the store and never attempt delivery, so the affected subset is smaller than that. `replaces malformed follower data on Follow` and, in *mastodon.test.ts*, `stores follower in KV when Follow is approved` use the same fixture pattern, but I did not run them.

Three tests that do check delivery were repaired in [`c5c9d61`] by moving their destinations onto the RFC 3849 documentation prefix `2001:db8::/32`, which `validatePublicUrl()` accepts and which routes nowhere.  Each test file owns a disjoint /48 and answers every request inside it from a module-scope `fetch` interceptor.  Applying the same treatment to the module-level alice fixture should cover the remaining tests: move her inbox onto that prefix, then assert the response status wherever a test approves her `Follow`.  That was left out of the earlier change because it reaches much further into both files.

I found no sign of broken relay behavior in production here.  The hazard is in the tests: adding `strictEqual(response.status, 202)` to any of them fails today because of the fixture's hostname, not because of whatever change the contributor was making.

[GHSA-f59r-8gcj-68f2]: https://github.com/fedify-dev/fedify/security/advisories/GHSA-f59r-8gcj-68f2
[`c5c9d61`]: https://github.com/fedify-dev/fedify/commit/c5c9d61784d55f7ce383f4b6f5d39bc44bc6f519

sij411:
Fixes #1046 

- Move Alice's inbox fixtures to intercepted IPv6 documentation addresses.
- Assert `202 Accepted` for previously unchecked inbox responses.
- Verify approved follows send exactly one `Accept POST` from the relay, referencing the correct `Follow`

AI assistance
- I planned the fix and Codex:gpt-6-astra implemented it.


sij411:
@codex review

sij411:
@coderabbitai review

sij411:
https://github.com/fedify-dev/fedify/pull/1047/changes/af66039543d171e9ad98d2f9aea8e93b59705dce : not 100% sure this is the right solution so if y'all have better solution leave some comments on please

sij411:
@coderabbitai full review

packages/relay/src/mastodon.test.ts
```diff
@@ -1009,7 +1042,7 @@ describe("MastodonRelay", () => {
     const follower1 = new Person({
       id: new URL("https://remote.example.com/users/alice"),
       preferredUsername: "alice",
-      inbox: new URL("https://remote.example.com/users/alice/inbox"),
+      inbox: new URL(aliceInbox),
```
dahlia, line 1045: `follower2`'s inbox on line 1050 still points at `remote.example.com`, so this test mixes two fixture conventions.  Nothing fails today because it never delivers, but that is easy to misread later.  Either move bob onto the intercepted prefix as well, or note that this fixture is store-only.

packages/relay/src/litepub.test.ts
```diff
@@ -141,6 +142,25 @@ function recordInbox(inbox: string): DeliveredRequest[] {
...
+): void {
+  const accepts = deliveries.filter(({ body }) =>
+    body.type === "Accept" && body.object?.id === follow.id?.href
+  );
+  strictEqual(accepts.length, 1, "Expected exactly one Accept for this Follow");
+  strictEqual(accepts[0].method, "POST");
+  strictEqual(accepts[0].body.actor, "https://relay.example.com/users/relay");
+  strictEqual(accepts[0].body.object.type, "Follow");
+  strictEqual(
+    accepts[0].body.object.actor?.id ?? accepts[0].body.object.actor,
```
dahlia: This accepts either serialization shape, so it passes whichever one the relay happens to emit.  Pinning the shape the relay actually produces would catch a regression that changes it.  The same helper in *mastodon.test.ts* has the same issue.

packages/fedify/src/federation/send.ts
```diff
@@ -292,6 +292,11 @@ async function sendActivityInternal(
       try {
         await validatePublicUrl(url);
       } catch (error) {
+        // DNS resolution errors carry the resolver's failure as their cause;
+        // private-address and protocol policy rejections do not.
+        if (error instanceof UrlError && error.cause != null) {
```
dahlia: This is unrelated to #1046 and is not mentioned in the pull request description.  Could you split it into its own issue and pull request?

Two things to settle there:

 -  The bug is already released.  68ec7b4a shipped in 2.0.28 through 2.3.8, so under the branch policy the fix belongs on *2.0-maintenance* rather than *main*.
 -  `error.cause != null` relies on `validatePublicUrl()` attaching a cause only for DNS failures, which is an internal detail of the validator.  Please add an explicit property to `UrlError` and discriminate on that instead.  Adding API in a patch release is not ideal, but this is small enough to be acceptable.  It would also cover `DNS lookup did not return any IP address` in *packages/vocab-runtime/src/url.ts*, which carries no cause and so stays a `UrlError` under the current heuristic.

packages/fedify/src/federation/send.ts
```diff
@@ -333,6 +337,7 @@ async function sendActivityInternal(
   }
   let response: Response;
   try {
+    await validateUrl(inbox.href);
```
dahlia: Moving this inside the `try` also makes private address rejections record a delivery failure metric and an error log, which is a separate behavior change from the DNS one and is not in the changelog fragment.  Worth deciding explicitly when this moves to its own pull request.

packages/fedify/src/federation/send.test.ts
```diff
@@ -881,3 +881,65 @@ for (const signed of [false, true]) {
...
 }
+
+for (const signed of [false, true]) {
+  test({
+    name: `sendActivity() accounts for DNS failures (signed: ${signed})`,
+    // validatePublicUrl() intentionally skips DNS without Deno net permission.
+    ignore: "Deno" in globalThis &&
+      (await Deno.permissions.query({ name: "net" })).state !== "granted",
+    async fn(t) {
+      const destination = "https://fedify-delivery.invalid/inbox";
```
dahlia: `fedify-delivery.invalid` performs a real DNS query on Node.js and Bun.  On a network that answers `NXDOMAIN` with a public address, validation passes, the catch-all mock returns `202 Accepted`, and this fails for reasons unrelated to the code.  Stubbing the resolver would be better; failing that, a comment would save the next person the debugging.

dahlia (changes requested):
The test changes do what #1046 asked for, and the new assertions have teeth: putting Alice's inbox back on `remote.example.com` fails three tests in *litepub.test.ts*.

My one structural request is the last commit, “Preserve transport errors for DNS failures,” which changes delivery behavior in `@fedify/fedify` and is unrelated to this issue.  Details are inline.  Please drop it from this branch and open a separate issue for it; the relay tests pass without it, so nothing here depends on it.

The rest is minor.

packages/relay/src/mastodon.test.ts
sij411: Addressed in https://github.com/fedify-dev/fedify/pull/1047/commits/1ce0eb80089d25920a30bed8beb34cbe82acfb20, now bob is moved onto the intercepted prefix.

packages/relay/src/litepub.test.ts
sij411: Addressed in https://github.com/fedify-dev/fedify/pull/1047/commits/1ce0eb80089d25920a30bed8beb34cbe82acfb20, both helpers are fixed.

packages/fedify/src/federation/send.ts
sij411: Created an issue #1055

packages/fedify/src/federation/send.ts
sij411: Created an issue https://github.com/fedify-dev/fedify/issues/1055

packages/fedify/src/federation/send.test.ts
sij411: Also mentioned in #1055

dahlia (approved):
Great work, thanks!

PR #1050, opened 2026-09-22
Make outbox-listener-delivery-required path-aware

### LINKED ISSUE #900: Make the outbox delivery lint rule path-aware

Jae-Hyuk-Jang:
Closes #900

Supersedes #1040, which targeted `2.2-maintenance` on the assumption that this was a patch-safe bug fix. dahlia's review there concluded it isn't (the rule reports more errors than before, which doesn't belong in a patch release) and asked for the underlying mechanism to be redesigned, so this reopens against `main` with a clean history.

## Background

`outbox-listener-delivery-required` (`@fedify/lint`) decided whether a listener delivers a posted activity by scanning its source as flat text and re-matching names with regexes. That representation could not express whether a callback's result is awaited or returned, whether an identifier is a call or a reference, or which scope a name resolves in, which produced both false negatives (the original #900 cases) and, in the first attempt at this on #1040, new false positives.

## Changes

- Keep collecting reachable statements (if/else, try/catch/finally, switch, loops), fixing unconditional exits (e.g. `if (true) return;`) not propagating past a single nesting level.
- Resolve a call's callee to the function it actually invokes by following variable bindings (reusing the same resolution already  used for the listener argument itself) instead of regex-matching a helper's name. This correctly handles a helper called by reference, held in an object literal, or calling a sibling helper, and stops a member call like `someService.deliver()` from being confused with an unrelated local `deliver`.
- Fold in an anonymous callback's body when the call it's passed to is awaited or returned, and an IIFE's body unconditionally.
- Keep the existing text-based pattern matching only for recognizing the delivery method name itself (aliases, destructuring, bracket/template notation), now run over the resolved reachable text.
- Update the rule's documentation to describe reachability, and revise the changelog fragment to describe the cases the rule now catches instead of calling this a bug fix.

## Testing

- `mise run check-each lint`
- `mise run test-each lint` (Deno and Node.js, 658/658 pass)
- `mise run check:md`
- `mise run docs:build` (Twoslash blocks in the updated docs compile)
- Added regression tests for every false positive/negative dahlia and CodeRabbit found on #1040, plus the original four cases from #900. Reverted the implementation locally and reran the suite to confirm the new tests fail without it.

## AI disclosure

This was implemented with Claude Code (`claude-sonnet-5`): I directed the redesign based on dahlia's review (move reachability
and call resolution onto the AST, keep text matching only for the delivery method name itself) and reviewed the results at each step; Claude Code implemented it, found and fixed a bug during testing (an object-literal property's key was being counted as a reference to an unrelated same-named helper), and verified the tests pass in both Deno and Node.js.


packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +290,489 @@ const resolveListenerReference = (
...
+function collectReachableStatements(node: Node, out: Node[]): void {
+  switch (node.type) {
+    case "BlockStatement":
+      for (const statement of node.body) {
+        collectReachableStatements(statement as Node, out);
+        if (alwaysExits(statement as Node)) return;
+      }
+      return;
+
+    case "IfStatement": {
```
dahlia, line 373: The rewrite handles everything else I raised, so this is the one case still standing from my last review.

Head expressions are still never collected.  `IfStatement` recurses into `consequent` and `alternate` but not `test`, `SwitchStatement` walks the case consequents but not `discriminant` or each case's `test`, and the loop branches walk only `body`, skipping `init`, `test`, `update` and `right`.  Both of these pass on `main` and are reported here:

~~~~ typescript
if (await ctx.sendActivity({ identifier: ctx.identifier }, inbox, activity)) {
  console.log("sent");
}
~~~~

~~~~ typescript
async function deliver() {
  await ctx.sendActivity({ identifier: ctx.identifier }, inbox, activity);
  return 1;
}
switch (await deliver()) {
  case 1:
    break;
}
~~~~

A delivery call in a head position is unlikely given `sendActivity()`'s return type, and I said as much last time, so I am not treating this as a blocker.  A *helper* call in a head position is more plausible than the first example suggests, though, and the fix looks small: collect the head expressions alongside the bodies, since the rest of the machinery already handles whatever it finds there.

If you would rather leave it, say so and I will open a follow-up issue instead of holding this up.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +290,489 @@ const resolveListenerReference = (
...
+    case "SwitchStatement":
+      for (const switchCase of node.cases) {
+        for (const statement of switchCase.consequent) {
+          collectReachableStatements(statement as Node, out);
+          if (alwaysExits(statement as Node)) break;
+        }
+      }
+      return;
+
+    case "WhileStatement":
```
dahlia, line 403: Same category, other direction, and also unchanged from last time: a pre-test loop body counts even when the loop provably never runs.

~~~~ typescript
while (false) {
  await ctx.sendActivity({ identifier: ctx.identifier }, inbox, activity);
}
console.log(ctx.identifier);
~~~~

This is accepted, so the listener passes.  `main` behaves the same way, so it is not a regression, and #900 never asked for it.  I mention it only because `isStaticallyFalsy()` is right there and `if (false)` is already handled, which makes the gap look like an oversight rather than a decision.  `for (; false;)` is the same.  A `do` body runs once, so it should keep counting.

CodeRabbit raised this as well, with the same `do…while` caveat.  Happy to leave it for a follow-up either way.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +290,489 @@ const resolveListenerReference = (
...
+ * scope: a `function name() {}` declaration, or a `const name =
+ * function/arrow` binding. Does not descend into a found function's own
+ * body -- a helper nested inside another helper is only found once that
+ * outer helper is itself resolved as reachable, so it can be layered on
+ * top of (and correctly shadow) the outer scope's helpers of the same
+ * name. Object-literal properties are handled separately, through
+ * `resolveFunctionBinding`, since their name is only meaningful together
+ * with the object it lives on.
+ */
+function collectNamedHelpers(
```
dahlia, line 491: `collectNamedHelpers()` only walks the listener body, so a helper declared at module scope is invisible and its delivery call does not count:

~~~~ typescript
async function deliverIt(ctx, activity) {
  await ctx.sendActivity({ identifier: ctx.identifier }, inbox, activity);
}

federation
  .setOutboxListeners("/users/{identifier}/outbox")
  .on(Activity, async (ctx, activity) => {
    await deliverIt(ctx, activity);
  });
~~~~

This is reported, although the activity is delivered.  It behaves identically on `main`, so it is pre-existing rather than something this PR introduced, and I said last time that it was not your responsibility.  I am raising it again only because extracting a delivery helper to module scope is a normal thing to do, and this rewrite is the first version of the rule with the machinery to resolve it.  Whether that belongs here or in a follow-up is your call; a follow-up is fine by me.

changes.d/lint/outbox-listener-path-aware.md
```diff
@@ -0,0 +1,17 @@
...
+  '#900': https://github.com/fedify-dev/fedify/issues/900
+---
+ -  Changed `outbox-listener-delivery-required` (`@fedify/lint`) to decide
+    whether a `ctx.sendActivity()`/`ctx.forwardActivity()` call actually
+    runs,
+    instead of scanning the listener's source as a flat block of text. It
+    now reports a listener whose only delivery call sits behind a dead
+    branch, after an unconditional `return`/`throw`, or inside a local
+    helper function that is never actually called. A helper that is
+    called still counts, however it's referenced — by name, passed by
```
dahlia: Small style point, plus something more general behind it.

*CONTRIBUTING.md* asks for an em dash without surrounding spaces in narrative text, and the project prefers avoiding them altogether where a comma, colon or a reworded sentence will do.  There are three spaced ones in this PR: two here, on this line and on line 15, and one in *docs/manual/lint.md* on line 761.  Rewriting the sentence is probably better than re-punctuating it, since the clause it sets off would read fine after a colon.

The general point: that section of *CONTRIBUTING.md* is worth a read before the next round.  Spaced em dashes are a habit of AI-generated prose in particular, and a repository convention is exactly the kind of thing a model will not follow unless someone checks its output against it.  You have disclosed the assistance properly on both this PR and #1040, which I appreciate and which is more than many contributors do.  What *AI\_POLICY.md* asks for beyond disclosure is that a human has actually gone over the result, and details like this are the visible signal of whether that happened.  The code in this rewrite is genuinely good and clearly had thought put into it, so I would rather flag the pattern now than have it become a reason to distrust the work later.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +290,489 @@ const resolveListenerReference = (
...
+
+  const processScope = (
+    scopeRoot: Node,
+    outerHelpers: ReadonlyMap<string, FunctionLikeNode>,
+  ): void => {
+    if (visited.has(scopeRoot)) return;
+    visited.add(scopeRoot);
+
+    const statements: Node[] = [];
+    collectReachableStatements(scopeRoot, statements);
```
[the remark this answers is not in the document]
dahlia: Both findings in this comment are worth addressing.  I reached the same two independently, on the `IfStatement` case and on the loop cases, and confirmed each against `main`.

The `do…while` caveat here is right and my own comment was wrong before I corrected it: a `do` body executes once, so only the pre-test loops should be pruned.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +290,489 @@ const resolveListenerReference = (
...
+  collectReachableStatements(root, statements);
+
+  return statements
+    .map((statement) => {
+      let text = sourceCode.getText(statement);
+      const nested: FunctionLikeNode[] = [];
+      collectNestedFunctions(statement, nested);
+      for (const fn of nested) {
+        const fnText = sourceCode.getText(fn);
+        const replacement = used.has(fn)
```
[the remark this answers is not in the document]
dahlia: This one is real and I had missed it.  Reproduced: with two textually identical function literals in the same statement, whichever appears first decides the fate of both, because `text.split(fnText).join(…)` rewrites every occurrence and the second pass then finds nothing left to match.

~~~~ typescript
const handlers = {
  unused: () => ctx.sendActivity({ identifier: ctx.identifier }, inbox, activity),
  used: () => ctx.sendActivity({ identifier: ctx.identifier }, inbox, activity),
};
await handlers.used();
~~~~

That is reported, although `used` is called.  Swapping the two properties makes it pass, which is a good illustration of why the replacement should be keyed on each node's range rather than its text.  Worth addressing.

dahlia (changes requested):
This is the redesign I was hoping for.  Resolving a call's callee to the function it actually invokes, and folding in a callback only when its result is consumed, is the right shape, and `alwaysExits()` is a nice way to keep the dead-code reasoning conservative without overclaiming.

I re-ran every case from my review of #1040 against this branch, and checked each one against `main` to separate regressions from pre-existing behavior.  All seven false positives are gone, and all six false negatives now report, including the nested-exit propagation that CodeRabbit found.  The original #900 cases and the positive cases still behave.  `sacho check`, `mise run check-each lint` and the rule's own tests all pass here, the changelog fragment is accurate this time, and the documentation is updated.  Thank you for taking the rewrite on rather than patching around the old mechanism.

One thing I would like fixed before this lands.  CodeRabbit's comment about replacing nested functions by node range rather than source text is a real false positive, and I reproduced it:

~~~~ typescript
const handlers = {
  unused: () => ctx.sendActivity({ identifier: ctx.identifier }, inbox, activity),
  used: () => ctx.sendActivity({ identifier: ctx.identifier }, inbox, activity),
};
await handlers.used();
~~~~

This is reported even though `used` is called.  `text.split(fnText).join(…)` rewrites every occurrence, so when two function literals in one statement have identical source text, whichever comes first decides both and the second pass has nothing left to match.  Swapping the two properties makes it pass.  Keying the replacement on each node's range, applied from the end of the statement backwards, should settle it.  I have left a note on that thread as well.

The three remaining gaps are fine to leave.  Control-flow head expressions are still unscanned, pre-test loops with a statically false test still count, and a helper declared at module scope is still invisible.  The first two are narrow, the third is pre-existing and behaves the same on `main`, and none of them reports correct code in a way a real listener is likely to hit.  I will open a follow-up issue covering all three once this merges, so they do not get lost.

There is also a small note on the changelog fragment about the em dash convention.  It is minor on its own, but please give it a read.

packages/lint/src/rules/outbox-listener-delivery-required.ts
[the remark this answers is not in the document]
Jae-Hyuk-Jang: Fixed in 1a60aabe: `collectDeliveryScanCode` now splices by each function's own `range`, applied from the end of the statement backward, instead of matching by source text. Added two regression tests covering your repro in both property orders: the used one first, and the used one second, since the old bug depended on which byte-identical function got processed first.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +290,518 @@ const resolveListenerReference = (
...
+  const walkConsumed = (expr: unknown): void => {
+    if (expr == null || typeof expr !== "object" || Array.isArray(expr)) {
+      return;
+    }
+    if (!isNode(expr)) return;
+    if (isFunctionLikeNode(expr)) {
+      out.add(expr);
+      return;
+    }
+    if (expr.type === "CallExpression" || expr.type === "NewExpression") {
```
dahlia, line 608: The range-based splice works, and I checked it from both orders and with three identical literals.  While re-testing it I found one more false positive of the same class as the `Promise.all(recipients.map(…))` case from the last round, which I should have caught then.

`walkConsumed()` follows the arguments of a `CallExpression` or a `NewExpression` and stops there.  An `ArrayExpression`, a `SpreadElement` and an `ObjectExpression` are never walked, so a callback one level inside a collection is not marked as consumed:

~~~~ typescript
await Promise.all([
  ...inboxes.map((inbox) => ctx.sendActivity(sender, inbox, activity)),
  ...others.map((inbox) => ctx.sendActivity(sender, inbox, activity)),
]);
~~~~

~~~~ typescript
await Promise.all(
  [inboxes.map((inbox) => ctx.sendActivity(sender, inbox, activity))].flat(),
);
~~~~

Both are reported, while `await Promise.all(inboxes.map(cb))` passes, so the wrapping is the only difference.  This is not left over from the identical-text collision: two *different* callbacks in spreads fail the same way.  Both forms pass on `main`, so it is a regression this PR would introduce, and it predates your last commit.

The asymmetry is visible a few lines up.  `walkAwaitExpressions()` recurses through arbitrary node shapes to find an `await`, while `walkConsumed()` knows only two node types.  Teaching it to descend through array and object literals and through spread elements would line the two up.

Merging several recipient lists before delivering is ordinary enough that I would rather not ship this one.

packages/lint/src/rules/outbox-listener-delivery-required.ts
dahlia: Split out into #1052 so it does not hold this up.

packages/lint/src/rules/outbox-listener-delivery-required.ts
dahlia: Split out into #1053, tagged as a good first issue since the change is small.

packages/lint/src/rules/outbox-listener-delivery-required.ts
dahlia: Split out into #1054.  It is the largest of the three, so it may sit for a while.

packages/lint/src/rules/outbox-listener-delivery-required.ts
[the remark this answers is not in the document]
dahlia: Both of these are tracked separately now: the head expressions in #1052, the statically false pre-test loops in #1053.  The `do…while` caveat is recorded in #1053.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +290,518 @@ const resolveListenerReference = (
...
+    }
+  };
+
+  for (const statement of statements) {
+    if (statement.type === "ReturnStatement") {
+      // A return value is consumed by definition -- no `await` needed.
+      if (statement.argument != null) walkConsumed(statement.argument);
+      continue;
+    }
+    if (statement.type === "ExpressionStatement") {
```
dahlia, line 703: Related to my note on `walkConsumed()`, but a separate case, and this one leaves the rule inconsistent with itself.

An `ExpressionStatement` is only searched for an `await`, so a callback passed to a synchronous iteration method is never consumed:

~~~~ typescript
inboxes.forEach((inbox) => ctx.sendActivity(sender, inbox, activity));
~~~~

That is reported, while the same thing written through a named helper passes, because a by-name reference is resolved elsewhere:

~~~~ typescript
const deliver = (inbox) => ctx.sendActivity(sender, inbox, activity);
inboxes.forEach(deliver);
~~~~

Two spellings of one behavior, two different answers.  `main` accepts both.

`forEach` with an `async` callback is not good code, since nothing awaits the returned promises, but that is a different complaint from the one this rule makes, and the message would send someone looking for a missing delivery call that is right in front of them.  If the intent is to flag fire-and-forget delivery, that deserves its own rule and its own wording.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +290,518 @@ const resolveListenerReference = (
...
+
+    default:
+      return false;
+  }
+}
+
+function collectReachableStatements(node: Node, out: Node[]): void {
+  switch (node.type) {
+    case "BlockStatement":
+      for (const statement of node.body) {
```
dahlia, line 367: Function declarations hoist, so stopping the statement walk at an unconditional exit can drop a declaration that the reachable code above it already used:

~~~~ typescript
await deliver(inboxes[0]);
return;
function deliver(inbox) {
  return ctx.sendActivity(sender, inbox, activity);
}
~~~~

`deliver()` runs, yet this is reported.  Dropping the `return` makes it pass, which isolates it to this loop.  `main` accepts it.

Contrived as written, though declaring helpers below the code that uses them is a real style, and an early return above them is not exotic.  Collecting `FunctionDeclaration` nodes from the whole statement list before the reachability walk prunes anything would cover it, since hoisting is exactly what the language does here.

Lower priority than the other two from this round, and I am happy for it to become a follow-up issue instead.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +290,518 @@ const resolveListenerReference = (
...
+      collectNamedHelpers(statement, helpers);
+    }
+
+    const referencedNames = new Set<string>();
+    for (const statement of statements) {
+      collectReferencedNames(statement, referencedNames);
+    }
+    const reached = new Set<FunctionLikeNode>();
+    for (const [name, fn] of helpers) {
+      if (referencedNames.has(name)) reached.add(fn);
```
dahlia, line 800: The other side of the by-reference support, for the record rather than as a request.

Any mention of a helper's name marks it reached, so a helper that is only inspected counts as called:

~~~~ typescript
const deliver = (inbox) => ctx.sendActivity(sender, inbox, activity);
console.log(deliver);
~~~~

~~~~ typescript
const deliver = (inbox) => ctx.sendActivity(sender, inbox, activity);
const kept = deliver;
~~~~

Neither listener delivers, and the rule says nothing.

This is the cost of making `inboxes.forEach(deliver)` work, and telling the two apart properly would mean knowing which APIs invoke their argument, which is more than this rule should carry.  Missing a case is the safe direction here, so I would leave it as is.  Worth a comment in the code so the next person does not “fix” it and break passing a helper by reference.

dahlia (changes requested):
The range-based splice is right, and it holds up: I checked both property orders and a three-literal variant, and every case from the previous two rounds still behaves.  Thank you for the commit message on that one, the explanation of why the replacements run backward is exactly what someone will want in a year.

I ran another pass over the whole diff, with some machine help, and it turned up three more cases.  Two of them I would like fixed here; two can wait.

Both blockers are the same shape.  Deciding whether a callback is consumed only recognizes a narrow set of positions, so a delivery call that plainly runs gets reported:

~~~~ typescript
await Promise.all([
  ...inboxes.map((inbox) => ctx.sendActivity(sender, inbox, activity)),
  ...others.map((inbox) => ctx.sendActivity(sender, inbox, activity)),
]);

inboxes.forEach((inbox) => ctx.sendActivity(sender, inbox, activity));
~~~~

The first fails because `walkConsumed()` descends through call arguments but not through array literals, spread elements or object literals.  The second fails because an `ExpressionStatement` is only searched for an `await`.  Both pass on `main`, and the `forEach` one is inconsistent with the rule's own behavior, since writing the same thing as `inboxes.forEach(deliver)` with a named helper passes today.  Details are on the two threads.

Stepping back: this is the third round where consumption detection has been the thing that broke, each time in a different shape.  Direct `Promise.all(map)` was the first, collection wrapping is the second, synchronous iteration the third.  Rather than add two more cases, it may be worth one pass over `collectConsumedCallbacks()` that decides what “consumed” means and covers it in one go: a callback is consumed when the call receiving it is reachable and either its result is awaited or returned, or the receiving call invokes it synchronously.  Whatever shape you land on, please write it down in a comment, so the next reader does not have to reconstruct it from the cases.

The other two can wait, and I have left notes on both.  A hoisted `function` declaration below an unconditional `return` is dropped from the scan even though the code above already called it; that one is narrow enough that I am happy for it to become a follow-up.  Separately, any mention of a helper's name counts as invoking it, so a helper that is only logged or stored satisfies the rule.  That is a false negative, it is what makes passing a helper by reference work at all, and I would leave it alone.  A comment saying so would stop someone from “fixing” it later.

The three gaps from my previous review are now #1052, #1053 and #1054, so nothing there is blocking.

packages/lint/src/rules/outbox-listener-delivery-required.ts
Jae-Hyuk-Jang: Fixed in e4c56f62: `walkConsumed()` now also expands through `ArrayExpression` elements, `SpreadElement` arguments, `ObjectExpression` property values, and a chained call's receiver (for the `.flat()` case), instead of only a call's own arguments. Added tests covering both of your examples.

packages/lint/src/rules/outbox-listener-delivery-required.ts
Jae-Hyuk-Jang: Fixed in e4c56f62: `forEach()` callbacks are now recognized independently of the await/return check, since `forEach()` always returns `undefined` and there is nothing to consume. `map()`/`filter()`/etc. still require an await or return, since their return value is usually meant to be consumed, so the existing unawaited-`map()` case still reports. Added tests for both directions.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +305,600 @@ const resolveListenerReference = (
...
+    }
+    return;
+  }
+  if (!isNode(node)) return;
+  const n = node;
+  if (isFunctionLikeNode(n)) return;
+
+  if (n.type === "CallExpression") {
+    const callee = n.callee as Expression;
+    const resolved = callee.type === "Identifier" && helpers.has(callee.name)
```
dahlia, line 562: `resolveFunctionBinding()` is reached only for an `Identifier` callee or the shapes it already knows, so a helper that arrives through a property lookup is not resolved.  All three of these deliver, all three are reported, and all three pass on `main`:

~~~~ typescript
const handlers = { deliver: () => ctx.sendActivity(sender, inbox, activity) };

const alias = handlers.deliver;
await alias();

await handlers["deliver"]();

const { deliver } = handlers;
await deliver();
~~~~

The direct form `await handlers.deliver()` passes, which is what makes the rest look arbitrary to someone hitting it.  The destructuring case seems worth special attention, since pulling a method off an object is ordinary in a listener that groups its helpers.

This and my note on `collectNamedHelpers()` are the same underlying gap seen from two sides: binding resolution covers the shapes that were tested and not the neighboring ones.  Rather than adding three more branches, it may be worth deciding what set of shapes counts as “this identifier holds that function” and covering it once, the way you consolidated callback consumption last round.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -273,11 +305,600 @@ const resolveListenerReference = (
...
+    return;
+  }
+  if (
+    n.type === "FunctionExpression" || n.type === "ArrowFunctionExpression"
+  ) {
+    return;
+  }
+  if (n.type === "VariableDeclarator") {
+    const decl = n as VariableDeclarator;
+    if (decl.id.type === "Identifier" && decl.init != null) {
```
dahlia, line 514: Only a declaration's own initializer is recorded, so a helper assigned afterwards is never linked to its name:

~~~~ typescript
let deliver;
deliver = (inbox) => ctx.sendActivity(sender, inbox, activity);
await deliver();
~~~~

Reported, though it delivers, and accepted on `main`.

Recording a plain `AssignmentExpression` whose left side is an identifier and whose right side is a function literal would cover it.  Worth thinking about alongside the property-lookup cases on my other comment, since both are the same question about which shapes bind a name to a function.

packages/lint/src/rules/outbox-listener-delivery-required.ts
dahlia: Still open after the latest push, and now the last unaddressed thread on this PR.

The fix looks small next to what you have already done here: collect `FunctionDeclaration` nodes from the whole statement list before the reachability walk prunes anything, since hoisting is what the language does with them regardless of where they sit.

I would rather close this one out here than carry it, but if you would sooner stop, say so and I will split it into a follow-up issue and approve.

dahlia (changes requested):
Both blockers from the last round are fixed, and you fixed them the way I hoped.  `SYNCHRONOUS_ITERATION_METHODS` draws the line where it belongs and the comment explains why `forEach` sits on one side and `map` on the other, which is the part that will keep someone from widening it carelessly later.  The note on the deliberate false negative is there too.  I ran the previous rounds' cases again along with new ones for collection wrapping, chained receivers and `forEach`, and nothing regressed; catching in 82f2b5da that your own fixture was not actually awaiting anything is the kind of thing most people would have let slide.

I ran another pass, and it found a cluster of false positives in how a name gets bound to a function.  Every one of these delivers, every one is reported, and every one passes on `main`:

~~~~ typescript
const handlers = { deliver: () => ctx.sendActivity(sender, inbox, activity) };

const alias = handlers.deliver;
await alias();

await handlers["deliver"]();

const { deliver } = handlers;
await deliver();
~~~~

~~~~ typescript
let deliver;
deliver = (inbox) => ctx.sendActivity(sender, inbox, activity);
await deliver();
~~~~

`await handlers.deliver()` passes, which is what makes the rest look arbitrary from the outside.  Details are on the two threads.

This is the second round where the failure has the same shape.  Last time a resolution layer covered the callback positions that had tests and not the ones beside them; this time it is binding resolution.  I do not think you are being careless, I think the rule is trying to do something genuinely hard, which is to resolve arbitrary JavaScript binding shapes without type information, and each round finds the next shape nobody thought to write down.

So rather than ask for four more branches, let me put an option on the table.  The file already says in two places that missing a case is the safe direction, and I agree.  You could make that the rule's actual contract: report only when the analysis can account for every `ctx.sendActivity()` and `ctx.forwardActivity()` call it can see and show that each one is unreachable.  When a callee will not resolve, stay quiet.  That turns this whole class from a false positive into a missed warning, which is the direction a lint rule should fail in, and it closes the class permanently instead of one shape at a time.  The dead-code cases from #900 still report, because there the rule can see the delivery call and prove the branch is dead.  If you take that route, please write the contract down in *docs/manual/lint.md* so users know what the rule does and does not claim.

The hoisted `function` declaration case from the last round is still open and is the only other thing outstanding.

Five rounds is a lot to ask of anyone, and you have answered every one of them properly, so I want to be clear about where the exit is.  If you would rather stop here, say so: I will file the binding-resolution cases and the hoisting case as issues, and approve what you have, since it is already a large improvement over what is on `main`.  If you would rather finish it, I would take the contract change above over more branches.  Either answer is fine by me.

packages/lint/src/rules/outbox-listener-delivery-required.ts
Jae-Hyuk-Jang: Fixed in 4626ac44, by taking the contract route you suggested instead of adding branches. A function held under a name (a declaration, a variable and whatever its initializer contains, an assignment, an object property, a class method) now counts as used as soon as that name is mentioned in code that runs, however it is mentioned. The alias, the `handlers["deliver"]()` call and the destructured `deliver` all resolve without the rule following the value. Each of your three shapes has a test, along with the neighbors I could think of: arrays, a wrapper call such as `once(fn)`, a class method, a property assigned after the object is created, and `const promises = inboxes.map(...); await Promise.all(promises)`, which was reported too. All the new cases pass on `main` and fail on the previous commit.

packages/lint/src/rules/outbox-listener-delivery-required.ts
Jae-Hyuk-Jang: Fixed in 4626ac44 as part of the same change. An assignment now records what it assigns under the name it writes to, for both `deliver = fn` and `handlers.deliver = fn`. The target of an assignment counts as a write rather than a mention, so a function that is assigned and never used is still reported. There is a test for each direction.

packages/lint/src/rules/outbox-listener-delivery-required.ts
Jae-Hyuk-Jang: Fixed in 4626ac44. The block walk now keeps function declarations that sit below an unconditional exit instead of pruning them with the dead code, so `await deliver(); return; function deliver() {...}` passes, while a hoisted helper that nothing mentions is still reported. Both directions have a test.

Jae-Hyuk-Jang:
Thanks for laying out the exit so clearly. I took the contract route, and I would rather finish than stop.

Since your last review:

- 4626ac44 makes a function held under a name count as used as soon as the name is mentioned in code that runs. That covers the binding cases from your review and their neighbors, and the hoisting case. The contract is written down in *docs/manual/lint.md*, including what the rule does not claim.
- e1f77d8c also scans control-flow head expressions, which covers #1052. I ran into it while sweeping about 35 further binding and callback shapes (factories, classes, `bind`, `Map`, spreads, `as`/`satisfies`): `for (const job of jobs) await job()` was reported because `jobs` only appears in the loop head. It is a separate commit, so it is easy to drop if you would rather keep #1052 as its own change.

`mise run check-each lint` and the package tests pass on Deno and Node, and the docs build passes.

One question I cannot settle on my own. Anonymous callbacks keep last round's rule: they count when awaited, returned, or passed to `forEach()`, which is what the third pattern in #900 asks for. So a callback passed to some other call whose result is dropped, such as `queue.push(() => ctx.sendActivity(...))` or `setTimeout(() => ..., 0)`, is still reported. Is that the intended behavior, or should the rule stay quiet there as well when it cannot tell where the callback goes? The manual states the current behavior, so users can see where the line is.

packages/lint/src/rules/outbox-listener-delivery-required.ts
```diff
@@ -43,6 +43,21 @@ const isChainedFromOutboxListeners = (
...
+ * for every element regardless. `forEach` always returns `undefined`; a
+ * callback passed to it has run by the time the call completes whether or
+ * not anything looks at what the call returns. This is deliberately
+ * narrower than the full iteration protocol: `map`, `filter` and the rest
+ * return something that usually *is* meant to be consumed (an array of
+ * promises passed to `Promise.all`, for instance), so calling one of those
+ * without awaiting or returning the result stays a real thing to flag, not
+ * a rule limitation to work around.
+ */
+const SYNCHRONOUS_ITERATION_METHODS = new Set(["forEach"]);
```
dahlia: Answering the question from your comment here, since this set is where the answer lands: the rule should stay quiet, and the gating should come out entirely.

I checked what the rule does today with delivery that is never awaited, and the picture decided it for me.  All of these pass:

~~~~ typescript
ctx.sendActivity(sender, inbox, activity);
void ctx.sendActivity(sender, inbox, activity);
ctx.sendActivity(sender, inbox, activity).catch(console.error);
const pending = ctx.sendActivity(sender, inbox, activity);
[inbox].forEach((i) => ctx.sendActivity(sender, i, activity));
deliver();
~~~~

Only the dropped `recipients.map(...)` is reported.  So the rule does not actually check awaiting; one shape falls through a gap in consumption detection, and that shape happens to be the one #900 named.  `queue.push(cb)` and `setTimeout(cb, 0)` are the same accident seen from the other side.

Which means `SYNCHRONOUS_ITERATION_METHODS` is not the line between two coherent behaviors, it is a patch over an inconsistency.  A callback passed anywhere is a callback the rule cannot show does not run, so under the contract you wrote it counts, whether the receiving call is `forEach`, `map`, `queue.push` or something it has never heard of.  That should let the awaited-or-returned machinery and this set both go away, which is a simplification rather than another branch.

I should own a mistake here.  Last round I argued that `forEach` had to be quiet because the delivery call plainly runs, and you implemented exactly that.  The reasoning was half right: the call does run, but a promise nobody awaits can still be cut off before the activity leaves, which is real on Cloudflare Workers where pending work is dropped once the response is returned.  So the concern behind #900's third pattern is genuine, and this rule is the wrong place for it, because it cannot even see the plainest version of the problem, the bare unawaited call on the first line above.  I will open a separate issue for a rule that checks delivery is awaited, where the message can say that instead of telling someone to call a method they can see they already called.

docs/manual/lint.md
```diff
@@ -752,13 +752,37 @@ Warns when an outbox listener body does not deliver the posted activity with
...
+missed warning is the safe direction, while a warning on code that delivers is
+not.  In practice:
+
+ -  A function held under a name counts as used as soon as that name is
+    mentioned anywhere in code that runs, however it is mentioned: called,
+    passed to another function, aliased, destructured from an object, or
+    reached through an array or a wrapper call.  The rule does not follow the
+    value any further, so a function that is only logged or stored, and never
+    called, is not reported.
+ -  An inline callback counts when its result is awaited or returned, such as
```
dahlia: This bullet goes away with the gating.  Something closer to:

> An inline callback counts wherever it is passed, since the rule cannot show that the receiving call never runs it.

The sentence below it about only reading the listener body should stay.

While you are here, the “Why it matters” paragraph may be worth a line saying what the rule does not check, so nobody reads the reachability contract as a guarantee that a reported-clean listener actually federates.  Delivery that is never awaited is the case I have in mind, now filed as #1057, so a pointer to that would do.

dahlia (changes requested):
The contract route was the right call, and you carried it further than I expected.  Flipping the default so a named function counts as used the moment its name appears in code that runs, and proving the negative instead of enumerating binding shapes, is exactly the inversion I was fishing for.

I spent most of my time trying to break it, because that flip is the kind of change that quietly guts a rule.  It did not.  Every case from #900 still reports: a helper declared and never used, a delivery call behind `if (false)` or after a `return`, the dead branch of `if (true)`, a listener that delivers nothing at all.  The scoping is tight in the places it would have been easy to get wrong.  A name mentioned only in a dead branch, only after an unconditional exit, only inside a comment, only inside a string, or only as an assignment target does not count as a mention.  Every binding case from my last review passes now, so do the head expressions, and nothing from the earlier rounds regressed.  The manual saying plainly what the rule does not claim is the part I am most glad to see.

Keep e1f77d8c here rather than splitting it out.  You found it from a real case, `for (const job of jobs) await job()`, and the commit stands on its own; I will close #1052 when this merges.

That leaves your question, and it turned out to be the most useful thing in this round.  I checked what the rule does with delivery that is never awaited, and the answer is that it does not check at all: a bare `ctx.sendActivity(...)` with no `await`, a `void`ed one, one with a `.catch()` hung off it, one stored in a variable and forgotten, a helper called without `await`, all pass.  The only unawaited shape that gets reported is the dropped `recipients.map(...)`, and it gets reported because of how consumption detection is built, not because the rule decided anything about awaiting.  `queue.push(cb)` and `setTimeout(cb, 0)` are the same accident from the other side.

So the answer is to stay quiet, and to take the gating out rather than add a case to it.  A callback passed anywhere is a callback the rule cannot show does not run, and the contract already says what to do about that.  `SYNCHRONOUS_ITERATION_METHODS` and the awaited-or-returned check should both come out, which should make this smaller rather than larger.  There is a note on the manual bullet that goes with it.

I owe you a correction.  Last round I told you `forEach` had to be quiet because the delivery call plainly runs, and you did that.  The call does run, but an unawaited promise can still be cut off before the activity leaves, which is not hypothetical on Cloudflare Workers, where pending work is dropped once the response is returned.  So the worry behind #900's third pattern is real.  It does not belong in this rule, which cannot see the plainest version of it, and the message here would be actively confusing for someone looking at a delivery call they can see.  #1057 now covers a rule that checks delivery is awaited, with the shapes above as its starting cases.  I will also amend that bullet on #900, since it reads as though a dropped `map` never runs its callback.  It does.

Once the gating is out and the manual matches, this is ready as far as I am concerned.  #1053 and #1054 stay as they are.

packages/lint/src/rules/outbox-listener-delivery-required.ts
Jae-Hyuk-Jang: Done in 3e862dcd. `SYNCHRONOUS_ITERATION_METHODS` and the awaited-or-returned check are gone, along with `collectResolvedCallTargets()`, which stopped changing any result once callbacks counted on their own, so the rule file is about 190 lines shorter. A function held under a name (a declaration, a variable or assignment, an object or array literal, a class member) still counts once the name is mentioned, and every other function literal now counts wherever it appears: a callback handed to `map`, `forEach`, `queue.push`, `setTimeout` or a call the rule has never heard of, an immediately invoked function, a returned closure.

Two existing tests flipped from Bad to Good: the dropped `recipients.map(...)` from #900 and my own kept-but-unused `map` result. New tests cover `queue.push`, `setTimeout`, a `then` that is not awaited, a callback inside an object passed to `Object.assign`, and a bare unawaited `ctx.sendActivity()`, so that the rule not checking `await` is pinned down. A callback that contains no delivery call is still reported.

docs/manual/lint.md
Jae-Hyuk-Jang: Done in 3e862dcd. The bullet now reads "An inline callback counts wherever it is passed, since the rule cannot show that the receiving call never runs it", and the sentence about reading only the listener body stays. I also dropped "or in an inline callback whose result is dropped" from the first paragraph, which said the opposite, and added a paragraph to "Why it matters" saying the rule checks that a delivery call exists and can run, not that the delivery completes, with a pointer to #1057. The changelog fragment matches.

Jae-Hyuk-Jang:
That should be everything from your last review. `mise run check-each lint` and the package tests pass on Deno and Node (706 tests), and the docs build passes.

Thank you for the careful rounds, and for the correction on `forEach` and the Cloudflare Workers context. I agree that concern belongs in a separate rule rather than in this one.

dahlia (approved):
The gating is out and the manual matches.  Everything I could think to throw at this behaves: the dropped `map()`, `queue.push()`, `setTimeout()` and `forEach` all stay quiet, while every case from #900 still reports, including a callback that appears only inside `if (false)`, only after a `return`, or only inside a helper nobody uses.

What I like most is that the rule kept getting smaller as it got more correct.  This last commit dropped the call-target resolver outright, and what the rule does now fits in one sentence.

Six rounds is a lot, and you answered each one by fixing the cause rather than the symptom.  Thank you for that, and for the question about unawaited delivery, which was the most useful thing said in this review; it became #1057.

#1052 is covered by e1f77d8c, so I will close it when this merges.  #1053 and #1054 stay open.

